PR #5, opened 2021-04-23
Re add post install scripts

mikecovesw:
My .gitignore was causing any new files under install/bin to be ignored. I updated .gitignore and re-installed the postInstall and postInstall.bat scripts.


MikeNeilson (approved):
That'll do it.

PR #8, opened 2021-07-09
Snotel bug fixes

mikecovesw:
These changes affect new code implemented for USDA SNOTEL data processing. It is a new module that acts similarly to a DECODES routing spec but uses custom code to convert data into the SNOTEL CSV ingest format.
These changes are low-risk.


schema/noaa/createLrgsDb.sql
```diff
@@ -0,0 +1,158 @@
...
+--------------------------------------------------------------
+-- data_source
+-- Each msg in storage holds a data_source_id.
+-- Data Source uniquely identified by type and name.
+-- DDS & DAMS_NT: name is host/ip-address of server.
+-- Name not used for DOMSAT data source.
+-- Data source record cannot be deleted as long as 
+-- any messages in storage hold it's ID.
+--------------------------------------------------------------
+CREATE TABLE data_source
```
MikeNeilson, line 78: based purely on the comment above  I would expect the constraints on the table to be as follows:
```
(
	data_source_id INTEGER NOT NULL primary key,

	-- For multi-LRGS databases, specify which LRGS this is a data source for.
	lrgs_host VARCHAR(64) NOT NULL,
	data_source_name VARCHAR(64) NOT NULL,
	data_source_type VARCHAR(24) NOT NULL,
	UNIQUE(data_source_name,data_source_type) -- perhaps lrgs_host,data_source_name,data_source_type since everything is not null
);
```

Generally when using a surrogate key, you don't combine it it another key, otherwise you end up having to put to column in to a table that references the columns.

src/main/java/lrgs/multistat/AlarmList.java
```diff
@@ -0,0 +1,152 @@
...
+		/* Add the alarm. */
+		alarmList.add(alarm);
+		if (getNumAlarms() == 1)
+// Already in GUI thread.
+//			SwingUtilities.invokeLater(
+//				new Runnable()
+//			{
+//				public void run()
+//				{
+					setValue(alarm);
```
MikeNeilson, line 55: This is rather nested for a straight function call. combined with the already in GUI thread comment it's probably reasonable to just deleted the commented code.

src/main/java/lrgs/multistat/SoundThread.java
```diff
@@ -0,0 +1,147 @@
...
+	public synchronized void
+		playMulti(File soundFile, long pauseMsec, int numTimes)
+	{
+		shutup();
+		this.soundFile = soundFile;
+		this.pauseMsec = pauseMsec;
+		this.numTimes = numTimes;
+	}
+
+	public synchronized void shutup()
```
MikeNeilson, line 48: I like this function  name.

MikeNeilson (approved):
Cool.

schema/noaa/createLrgsDb.sql
mikecovesw: This table was implemented only for NOAA. At Wallops they have two LRGS servers CDADATA and CDABACKUP. They both share a database used for authentication and statistics gathering. The DB happens to be hosted on Postgres on CDADATA. That's why there's no Oracle version for this script -- nobody else is using it.
The 'lrgs_host' column will either be cdadata or cdabackup. Perhaps if they implement a 3rd LRGS at the station there will be a 3rd possibility. 
So within the day files (not SQL), each message contains a data_source_id referring to one of these records. This capability for a multi-LRGS database was grafted on later. That's the reason for the complex surrogate key.

src/main/java/lrgs/multistat/AlarmList.java
mikecovesw: Done.

schema/noaa/createLrgsDb.sql
MikeNeilson: > `That's the reason for the complex surrogate key`

The problem is, a surrogate key should never be complex; it defeats the purpose of a surrogate key which is be devoid of all semantic meaning other that "this single digit/uuid/etc points to this particular row."

(I'm certainly not going to prevent the change from getting in, because as you said for NOAA, but I would be remiss not to point out the longer term maintenance issue of what you're doing.)

Arguments about programming correctness aside, in the above scenario. would it be correct to say that a data source is completely identified by the combination of lrgs_host and data_source_name? thus making (lrgs_host,data_source_name) the natural primary key? 

I would strongly suggest making the table the following: 
```sql
(
	data_source_id INTEGER NOT NULL primary key,

	-- For multi-LRGS databases, specify which LRGS this is a data source for.
	lrgs_host VARCHAR(64) NOT NULL,
	data_source_name VARCHAR(64) NOT NULL,
	data_source_type VARCHAR(24) NOT NULL,
	UNIQUE (lrgs_host,data_source_name,data_source_type) -- this is the "REAL" e.g. natural primary key, but the data_source_id is easier to use externally
);```

PR #19, opened 2021-11-09
[build] removed duplicate compile-test build element.

MikeNeilson:
Sorry, sloppy copy paste job from my local build.xml when I set that up.

Also I took the time to get the initial automated build infrastructure working. Current windows doesn't work. not sure if that's a lack of https on covesw.com or some other thing, but I've decided not to pollute the history with a bunch of commits to figure that out. I'll go figure that out in a separate place.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        platform: [ubuntu-latest,windows-latest,macos-latest]
+        platform: [ubuntu-latest,macos-latest]
     runs-on: ${{matrix.platform}}
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


MikeNeilson:
@mikecovesw please don't forget about this one.

Also quick note, I've verify the build automation include isn't going to prevent merges at this time. I'd like to see how it behaves over time though; I'm still not certain as to why it was failing other than the covesw host not liking so many downloads in quick succession.

mikecovesw (approved):
Eventually we will want to download jars from where they are maintained, rather than from covesw.com, right?

MikeNeilson:
Yes. that was the purpose of the original ant_build_deps PR I had, but I realized I was doing too much change all at one there. So one of my next will be, say ONLY, izpack having it's download from a repo vs trying to do literally everything all in one go and confusing people.

PR #40, opened 2022-02-01
HDB Rating changes

agilmore2:
Changes for HDB tickets #822 and #1135:
HDB rating tables can apply to more than one site/datatype
HDB ACAPS rating now can use variable exponent rating

### DIFF
--- a/src/main/java/decodes/hdb/HDBRatingTable.java
+++ b/src/main/java/decodes/hdb/HDBRatingTable.java
@@ -122,9 +122,11 @@ public class HDBRatingTable {
 		else try 
 		{
 			q="select effective_start_date_time,effective_end_date_time from ref_site_rating " +
-			"where rating_id = ?";
+			"where rating_id = ? and rating_type_common_name = ? and indep_site_datatype_id = ? ";
 			cstmt = conn.prepareCall(q);
 			cstmt.setInt(1, (int)ratingId.getValue());
+			cstmt.setString(2, ratingType);
+			cstmt.setInt(3, (int)sdi.getValue());
 
 			Boolean result = cstmt.execute();
 			if (!result) {
--- a/src/main/java/decodes/hdb/HDBRatingTable.java
+++ b/src/main/java/decodes/hdb/HDBRatingTable.java
@@ -73,16 +73,18 @@ public class HDBRatingTable {
 	public void findRating(Date valueDate)
 	throws DbCompException
 	{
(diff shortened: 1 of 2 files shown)


src/main/java/decodes/hdb/HDBRatingTable.java
```diff
@@ -73,16 +73,18 @@ public HDBRatingTable ()
...
 	{
-		String q = "select ratings.find_site_rating('" + ratingType + 
-		"'," + sdi + ", + to_date(?,'DD.MM.YYYY HH24:MI:SS')) from dual";
+		String q = "select ratings.find_site_rating(?,?," +
+		"to_date(?,'DD.MM.YYYY HH24:MI:SS')) from dual";
 		CallableStatement cstmt =  null;
 
 		try
 		{
 			cstmt = conn.prepareCall(q);
```
MikeNeilson, line 82: if you move this to the try-with-resources syntax you can ditch the finally block bellow.
e.g. try( cstmt = conn.prepalCall(q); ) {

src/main/java/decodes/hdb/HDBRatingTable.java
```diff
@@ -122,9 +124,11 @@ public void findRating(Date valueDate)
 		else try 
 		{
 			q="select effective_start_date_time,effective_end_date_time from ref_site_rating " +
-			"where rating_id = ?";
+			"where rating_id = ? and rating_type_common_name = ? and indep_site_datatype_id = ? ";
 			cstmt = conn.prepareCall(q);
```
MikeNeilson, line 128: same comment as above.

src/main/java/decodes/hdb/HDBRatingTable.java
```diff
@@ -122,9 +124,11 @@ public void findRating(Date valueDate)
 		else try
```
MikeNeilson, line 124: I don't think you wrote this bit but this line strongly implies a syntax structure that doesn't exist in java. Just wanted to document it for later review.

MikeNeilson (approved):
Both of the try blocks bail immediately from the computation. But in two different sections, if the function was altered to have one try it may be more readable. Maybe not. Would depend on whether you can cleanly sort out the exact cause on the SQL return, but food for thought.

agilmore2:
Thanks for the feedback!
The two catch blocks do have different error messages that might not be distinguishable from the exception caught. 
I wrote the original of this code a long time ago, so yes, this is my awkwardness.

I appreciate the merge!

PR #41, opened 2022-02-09
OPENDCS-13 add in a time interval threshold to only process contiguou…

adamkorynta:
…s data so that gaps don't end up getting filled in by downstream processes

### DIFF
--- a/src/main/java/opendcs/opentsdb/OpenTsdb.java
+++ b/src/main/java/opendcs/opentsdb/OpenTsdb.java
@@ -12,6 +12,7 @@ import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.Properties;
+import java.util.concurrent.TimeUnit;
 
 import opendcs.dai.IntervalDAI;
 import opendcs.dai.ScheduleEntryDAI;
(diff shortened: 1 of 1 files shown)


adamkorynta:
This change still needs testing

MikeNeilson:
... that is, way simpler that I was originally thinking it would be to do that. However it would need to be separately put in https://github.com/opendcs/opendcs/blob/c5cd9f7933c23ea1d3450f510e9f148d54140990/src/main/java/decodes/cwms/CwmsTimeSeriesDb.java#L1287.

However hold off. once the connection pool branch gets merge in the getNewData functions will get moved around.

MikeNeilson (changes requested):
There will need to be changes after a different branch is moved it so just keeping this one from accidentally get merged in.

mikecovesw (commented):
The approach is good. Stop when a large gap is seen. This will force CCP to only work on relatively contiguous data.
Two things:
1. The getNewData method has been moved to the TimeSeriesDAO (and subclasses for HDB, OpenTSDB, and CWMS). It is no longer in the tsdb class directly. (see email I just sent explaining this.)
2. This commit is for the OpenTSDB subclass, not CWMS. I think you want to make your mods in decodes.cwms.CwmsTimeSeriesDAO.

adamkorynta:
Moved code change to PR: https://github.com/opendcs/opendcs/pull/86

PR #44, opened 2022-04-05
Add check of validty for statement and connection before use.

MikeNeilson:
During testing of CONNECT_TIMEOUT settings The routing scheduler was
still showing "closed connection" and "closed statement" errors.
Since statemetns and result sets are invalid after a connection is
closed they must be rebuilt.

### DIFF
--- a/src/main/java/opendcs/dao/DaoBase.java
+++ b/src/main/java/opendcs/dao/DaoBase.java
@@ -117,9 +117,42 @@ public class DaoBase
 	protected Connection getConnection()
 	{
 		// local getConnection() method that saves the connection locally
-		if (myCon == null)
+		if (myCon == null || connectionClosed() )
+		{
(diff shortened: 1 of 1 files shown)


MikeNeilson:
I've left this test running for 2 days now with a CONNECT_TIMEOUT of 2 minutes and am still able to process data without database/statement connection errors.

mikecovesw (commented):
This looks fine, but I'm surprised that it's necessary. The DAOs are supposed to be short-lived. So why is a connection that it just got off of the pool getting closed?

mikecovesw (approved):

MikeNeilson:
So far as I could tell it wasn't a connection just off the pool. It appeared to be triggering the "I've already got a connection" logic. It's likely better solved more upstream from that particular bit of code but this seemed the fastest implement and easiest to follow.

PR #45, opened 2022-04-11
validation check for queryStmt2.

MikeNeilson:
This is part of the previous PR. Just realized I missed this particular update.

### DIFF
--- a/src/main/java/opendcs/dao/DaoBase.java
+++ b/src/main/java/opendcs/dao/DaoBase.java
@@ -208,7 +208,7 @@ public class DaoBase
 		}
 		try
 		{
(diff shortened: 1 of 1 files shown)


src/main/java/opendcs/dao/DaoBase.java
```diff
@@ -208,7 +208,7 @@ public ResultSet doQuery2(String q) throws DbIoException
 		}
 		try
 		{
-			if (queryStmt2 == null)
+			if (queryStmt2 == null|| statementInvalid(queryStmt1))
```
agilmore2, line 211: Sounds like you mean to check queryStmt2 here?

MikeNeilson: Wow. good catch.

PR #46, opened 2022-04-25
Fix ACAPS Area differentiation result with variable M coeff

agilmore2:


src/main/java/decodes/tsdb/algo/HdbACAPSRating.java
```diff
@@ -164,7 +164,7 @@ protected void doAWTimeSlice()
 			//See Documents from Rick Clayton on these equations.
```
MikeNeilson, line 164: It dawns on me that we probably able several comments like this (HDB, and CWMS) where someone not on the inside will go "who is that, and why should we have to look are their notes."

We might want to start gathering those up and including them either in source or in the wiki going forward.

MikeNeilson (approved):

PR #63, opened 2022-06-24
Make sure an objc jar is included for hdb.

MikeNeilson:
With all the dependency management changes it looks like we didn't
update the HDB objc jar in the izpack installer. It was still using ojdbc6 which is fairly old.
Oracle generally maintains backwards compatibility so this shouldn't be a huge change.

I did bump the version. Once approved an merged I'll make a new release.


Gaelmare (approved):

mikecovesw (commented):
Two questions:
Are there any legal issues with including Oracle's jar?

What is the current state of the JOOQ stuff? It used to be that having an ojdbc in the 'dep' directory (so it gets added to system classpath) broke JOOQ, and that's why it was included only if the user select HDB in the installer.

MikeNeilson:
Oracle changed the licensing for the ojdbc jar, actually another reason for using ojdbc8 as that's when the licensing changed.

As for the classloader issue... still working out the kinks there. It's supposed to not be an issue now but my issue tests had issues; but that could've been because of a bunch of other changes we made internally. Still investigating.

For the moment best to leave the design as is.

agilmore2 (approved):
Agreed that the ojdbc8 licensing allows for us to distribute with OpenDCS. I don't know about the JOOQ issue.

agilmore2:
I don't actually see an ojdbc8.jar in the resulting build? I thought I tested this, but I don't see one in the artifact!

We may need another build pronto.

MikeNeilson:
yep, just installed on a windows system hadn't had OpenDCS before. The jar is present.

agilmore2:
Crazy.

'unzip -l opendcs-ot-7-0-1.jar |grep -i jdbc' didn't show it, so I made a
poor assumption. I have an installation where they installed, but no file.

Thanks for triple checking me!

MikeNeilson:
yeah. I was thinking that too. but everything's embedded in those "PACK" files in the installer.

If it's still missing after running the install, my first though it usually "did you actually check the box?" That's happened to our CWMS users a lot.

PR #74, opened 2022-07-29
Modifications to build to support dependency management in ivy.

MikeNeilson:
This branch is now at a point where the build as a whole works and the procesess all appear to function (at least tested against CWMS, and will test later today/tonight/over weekend for the OpenDCS timeseries db) and I'll try to run the CWMS regression tests.

If anyone has time please check it out and see what happens with your setup. And if there's something glaring I just up and missed. but I think I got it all.

the steps to get the opendcs-ot-<version> izpack installer jar are as follows:
- ant 1.10+ installed
- java 1.8+ (for ant)
- clone project

run:
```bash
cd <project directory>
ant opendcs
```
The installer will be in `stage/` and will run as normal.

If you want an ide (like VSCode or eclipse to pick up the .classpath file, there is a new target `eclipse-ide-files` to generate the .classpath file and well as some other eclipse project files. I don't really use eclipse so I only verify I could open the project and run some targets:

```
ant eclipse-ide-files
```

I decided to move from the maven-ant-task jar to Apache Ivy. The maven ant plugin wasn't behaving correctly to exclude dependencies which was causing the dependency resolution to fail and I couldn't figure out way and there were no reports available. and it's kind of a hack in to ant. Apache Ivy is designed specifically for ant and uses the same general concepts. It was even easier to add in that Cobra.jar from github as part of the build. I didn't need a special seperate download.

To get a report of the dependencies and what dependencies they pull in run:

`ant common.report`

The resulting html files are in build/

Still todo:
- [x] build PDFs of documents for release. They do build and render at https://opendcs-env.readthedocs.io/en/7.1/ but we still want to include them in the release.
  - FYI, the PDF is a little, wonky. Scaling issues on the table. Modifiy build to include both HTML and PDF documentation.
-  [x] integrate with the ivy system a little more for handling the version number.
  -  this is for the opendcs.jar getting used as a normal "dependency" in sub- and 
downstream projects  and to keep the version number in one place  
  - Just baseline configuration. Actually publishing to maven central will require setting up code signing. Most likely first path is just publishing to github packages for downstream project development.
- [x] probably move the cwms depedencies into their own classpath so it's easier to lump them into the "CWMS Components" in the izpack. Especially since the the few components we need pull in so much.
- [x] checkstyle file as discussed here https://github.com/opendcs/opendcs/discussions/50
  - not run automatically. Most of our files don't match the current standard we came up with above. However there are checkstyle actions for github that will only look at code that was added.


MikeNeilson:
ALso ignore the "400+" file changes. It's not actually that much. The large number is from extracting the images to individual files for the documentation.

MikeNeilson:
FYI, I just tagged everyone to get your attention. This build works, but I went to setup an "OpenTSDB" postgres database to better test things and the bootstrapping was... well it's Monday and it's still not working (I started the previous Monday). 

If anyone has time some feedback would be good before I start doing more drastic changes to things like how the database files are structured and maintained. I want to make sure the build itself is buttoned up before doing too much more extra work.

FYI, I think my next step will actually be setting up the checkstyle file so the repo can help us keep the code format our definition of correct.

ivy.xml
```diff
@@ -0,0 +1,98 @@
...
+        <dependency org="org.codehaus.izpack" name="izpack-standalone-compiler" rev="4.3.5" conf="izpack->default"/>
+
+        <!-- groovy deps for biuld script -->
+        <dependency org="org.codehaus.groovy" name="groovy-ant" rev="3.0.10" conf="groovy->default"/>
+        <dependency org="org.codehaus.groovy" name="groovy" rev="3.0.10" conf="groovy->default"/>
+        <dependency org="org.codehaus.groovy" name="groovy-xml" rev="3.0.10" conf="groovy->default"/>
+
+        <!-- keep unused transient dependencies out -->
+
+        <exclude org="mil.army.usace.hec" module="loading-gui-lib"/>
```
adamkorynta, line 87: These `mil.army.usace.hec` dependency excludes can be removed when an hec-monolith 3.0.0 is released (non-rc).

adamkorynta (approved):

MikeNeilson:
@opendcs/opendcs-core-devs 

Okay, I need to merge current master in and rename some things but I'd like to move forward with getting this merged in. It will remove our dependency on covesw.com which is the primary culprit in breaking the automated test builds.

The only thing I really changed was updating was updating CWMS side dependencies and I've tested that against CWMS. if there are no objections I'll set the versioning for this back to 7.0 and we can go from there.

I've also got a multi-project build staged (separated opendcs.jar, docs, and the installer into subprojects) but That should be a step after this.

For those using Eclipse or  VScode the eclipse-ide-files task will properly generate the .project, .classpath, and other files as needed.

For eclipse, ivy provides an IvyDE plugin that likely has better integration, however I haven't used it.

adamkorynta (approved):
These changes look good to me and I was able to use this branch to more easily implement code changes.

mikecovesw (approved):
I had to upgrade ANT on my macbook, and then everything worked fine. Good job!

PR #75, opened 2022-08-01
Bc hydro alarm mods and improvements to Load Message dialog

mikecovesw:


src/main/java/decodes/dbeditor/DecodesScriptEditPanel.java
```diff
@@ -691,22 +690,17 @@ private void selectDecodedDataCell(int row, int col)
...
-				//TODO Highlight all cells in this row.
-				//TODO Highlight corresponding raw data
 			}
 			else
 			{
 				DecodedSample decSamp = findDecodedSampleContaining(
 					decodedDataTableModel.getTimedVariableAt(row, col));
 				if (decSamp == null)
 				{
-//System.out.println("No decoded sample at row " + row + ", col " + col);
```
MikeNeilson, line 704: this would probably be better as an if != null instead of just the empty if block.

MikeNeilson (approved):

src/main/java/decodes/dbeditor/DecodesScriptEditPanel.java
mikecovesw: Agreed. That's the way it "evolved". The only thing I did was remove the System.out call.

Issue #64, opened 2022-07-19
dbimport does not ignore historical versions


adamscarberry:
**Describe the bug**
When using dbimport, the docs and command line help indicates that historical versions will be excluded by default.  This doesn't appear to be the case.

**To Reproduce**
1. run dbimport against a platform export file with historical version included
2. check database or opendcs GUI
3. historical versions will appears

**Expected behavior**
Historical versions should not be imported by default.

**Screenshots**
![image](https://user-images.githubusercontent.com/46494851/179785151-453db24f-b8ec-4f40-b772-1023c3d6f925.png)

**Desktop/ServerSide (please complete the following information):**
 - OS: Linux Docker container
 - DECODES DB: OpenTSBD/Postgresql
 - openjdk version "11.0.15" 2022-04-19
 - OpenJDK Runtime Environment (build 11.0.15+10-post-Debian-1deb11u1)
 - OpenJDK 64-Bit Server VM (build 11.0.15+10-post-Debian-1deb11u1, mixed mode, sharing)

**Additional context**
`allowHistoricalArg` referenced below doesn't appear to be used in code:
 https://github.com/opendcs/opendcs/blob/b8d31a04b72395fbf8c5aa08e0d90a9be392849a/src/main/java/decodes/dbimport/DbImport.java#L153

zipped platform export below:
[p.lrn.xml.zip](https://github.com/opendcs/opendcs/files/9142434/p.lrn.xml.zip)

mikecovesw:
Fix completed 8/1/2022

adamscarberry:
awesome.  thanks @mikecovesw !

MikeNeilson:
will be in next release.

PR #76, opened 2022-08-01
Bug Fix: -H arg for dbimport was being ignored.

mikecovesw:
dbimport will now ignore historical versions on imported XML files unless the -H command line argument is given.


mikecovesw:
This is issue #64

MikeNeilson (approved):

PR #78, opened 2022-08-02
Implement function requested by zamusacelre.

### LINKED ISSUE #66: New Function to be added
**Add function into next version**
Given a daily time series,
When a new value comes in,
Then we want to write the latest value in the month at the end of the month.

**About**
Mike N has provided LRE with a jar that does this.  Would like to make sure future versions of OPENDCS have it so that as we upgrade we are sure the functionality remains.

MikeNeilson:
Closes #66.

### DIFF
--- a/install/imports/comp-standard/algorithms.xml
+++ b/install/imports/comp-standard/algorithms.xml
@@ -456,4 +456,18 @@ If
       <ParmType>o</ParmType>
     </AlgoParm>
   </Algorithm>
+  <Algorithm name="Last Value From Period">
+    <Comment>
+      Within a given window select the most recent valid value to be at the end
+    </Comment>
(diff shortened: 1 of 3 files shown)


agilmore2:
This is assuming triggering data comes in newest last. If a value comes in for the first of the month after the end of the month, that value will overwrite. If that's the expected behavior, this works.
Otherwise, I would have expected this algorithm to fill out the input timeseries for the input month and use the value for the latest date in the month as the output.

MikeNeilson:
It's an aggregating algorithm. By the time it gets to the doAWTimeSlice the system should've already filled out the window on our behalf with any available data regardless of which value within the window triggered the computation.

agilmore2:
Interesting. So every timeslice it sets the value, and the last available data wins. Nice.

mikecovesw (approved):

PR #82, opened 2022-08-05
Update to use IzPack 4.3.5 so console install works correctly.

### LINKED ISSUE #81: HDB headless install impossible
Since we are producing one installer without HDB components by default, HDB installs on headless servers are not working without manually copying at least the ojdbcX.jar file.

java -jar <opendcsinstaller.jar> -console or -properties <file> do not give the option of installing the HDB components. It doesn't look like our version of izPack allows this.

We either need to produce an HDB specific installer, an HDB specific server tarball, or find a way to select optional components in a headless install.

MikeNeilson:
Fixes #81.
4.3.2 and 4.3.4 were both tried. 4.3.3 was never uploaded to maven by codehaus.
Build downloads izpack on its own saving developers a step in the build.


agilmore2:
How does this solve the problem?

MikeNeilson:
Built with the newest 4.3.x version of IzPack, the console installer actually lets you pick the packages. Presumably the properties file would work else but I'm not super familiar with that so an extra is probably is in order.

agilmore2 (approved):

Issue #80, opened 2022-08-02
better handling of historic data


MikeNeilson:
**Is your feature request related to a problem? Please describe.**

There is a set of condition that when met can cause OpenDCS to query too much data from the database for a query.

Namely, if one is processing historical data that ends say, a year back. Saves it, and there is current data; depending on the exact volumes of the data it will end up getting the trailing end of the historic data and the newer data; this will be followed by OpenDCS trying to fill in that gap; which the CWMS database will struggle to do.

**Describe the solution you'd like**
Not having to worry about when historical get stored.

**Describe alternatives you've considered**

**Additional context**

MikeNeilson:
@adamkorynta don't forget about this one. Ruth has been chomping at the bit for a while and has been very patient with me.

PR #86, opened 2022-08-10
#80 add a limit to the amount of time between gaps of data for CWMS ts processing

adamkorynta:
resolves https://github.com/opendcs/opendcs/issues/80

The maximum time gap is set to 7 days with an overridable system property "opendcs.cwmstsdb.max_timegap_historical.days"

### DIFF
--- a/src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
+++ b/src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
@@ -155,7 +155,6 @@ public class CwmsTimeSeriesDAO
 {
 	protected static DbObjectCache<TimeSeriesIdentifier> cache = 
 		new DbObjectCache<TimeSeriesIdentifier>(60 * 60 * 1000L, false);
-	private static final long MAX_TIME_GAP_FOR_HISTORICAL = Integer.getInteger("opendcs.cwmstsdb.max_timegap_historical.days", 7) - 1L;
 	protected SiteDAI siteDAO = null;
 	protected DataTypeDAI dataTypeDAO = null;
 	private String dbOfficeId = null;
@@ -1424,7 +1423,7 @@ public class CwmsTimeSeriesDAO
 			return false;
 		}
 		long daysBetween = TimeUnit.MILLISECONDS.toDays(currentTimestamp.getTime() - lastTimestamp.getTime());
-		return daysBetween > MAX_TIME_GAP_FOR_HISTORICAL;
+		return daysBetween > (DecodesSettings.instance().cp_cwmstsdb_getNewData_max_timegap_days - 1);
 	}
 
 
--- a/src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
+++ b/src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
@@ -115,6 +115,7 @@ import java.util.Date;
 import java.util.Iterator;
 import java.util.List;
 import java.util.TimeZone;
+import java.util.concurrent.TimeUnit;
 
 import decodes.db.Constants;
 import decodes.db.DataType;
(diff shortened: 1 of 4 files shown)


src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
```diff
@@ -154,6 +155,7 @@ public class CwmsTimeSeriesDAO
 {
 	protected static DbObjectCache<TimeSeriesIdentifier> cache = 
 		new DbObjectCache<TimeSeriesIdentifier>(60 * 60 * 1000L, false);
+	private static final long MAX_TIME_GAP_FOR_HISTORICAL = Integer.getInteger("opendcs.cwmstsdb.max_timegap_historical.days", 7) - 1L;
```
MikeNeilson, line 158: Is this an jvm -D variable or within the (decodes|user).properties file?

generally we stick things like this within those as there's not really a good way for users to specific jvm parameter.... whether that should change though is a different question.

MikeNeilson (commented):
That otherwise looks good to me. But I don't want to merge it in directly yet.
I'll take a look and see if I can extend it to apply to the other systems as well.

src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
adamkorynta: That means adding it to the DecodesSettings class?

MikeNeilson: Yeah.

adamkorynta: It doesn't look like DecodesSettings supports properties with names that can't be java field variables. I set it to: `cp_cwmstsdb_getNewData_max_timegap_days`

MikeNeilson: ... well that makes sense. but is a whole other update.

src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
```diff
@@ -1407,9 +1414,16 @@ public DataCollection getNewData(DbKey applicationId)
...
+	}
+
+	final boolean exceedsMaxTimeGap(Date lastTimestamp, Date currentTimestamp)
+	{
+		if(lastTimestamp == null || currentTimestamp == null)
 		{
+			return false;
 		}
+		long daysBetween = TimeUnit.MILLISECONDS.toDays(currentTimestamp.getTime() - lastTimestamp.getTime());
+		return daysBetween > (DecodesSettings.instance().cp_cwmstsdb_getNewData_max_timegap_days - 1);
```
MikeNeilson, line 1426: wouldn't this just be `>=` or am I mixing up direction?

adamkorynta: Yup, you're right. Much easier to read.

src/test/java/decodes/cwms/TestCwmsTimeSeriesDAO.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.Calendar;
+import java.util.TimeZone;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+final class TestCwmsTimeSeriesDAO
```
MikeNeilson, line 11: .. wait, don't we usually put "test" at the end of the test class name?

MikeNeilson (approved):
This looks good to me.

Will give others some time to look before merging.

src/test/java/decodes/cwms/TestCwmsTimeSeriesDAO.java
adamkorynta: Yes, I believe that is the standard imposed by xUnit. I renamed

PR #91, opened 2022-08-23
Feature/dao improvements

MikeNeilson:
In a report I made during an analysis of the rest_api code that got uploaded I had discussed improving the SQL DAO usage to have zero non bind variable queries as well as simplifying resource (Connection, PreparedStatement, ResultSet) management.

This is an initial start on that work. I've create a handful of functional interfaces (the default Consumer did not handle checked exceptions), and modified EnumDAO to use exclusively use the new functions.

Names of the functions were retained to keep the same sense flow. They were expanded to take a caller provided consumer function and a varargs parameter list to bind each variable.

As can be seen in the example, it is not difficult to handle situations where different db versions require more or different parameters while retaining using bind variables.

Additionally it moves the boiler plate of Connection get/free, Statement open/close, and ResultSet open/close to a single location.

Generally this code assumes the jdbc driver or database will handle any statement caching. Based on my research this is true of Oracle and Postgres, which are our primary targets.

I still need to write some test to also demonstrate usages and verify operations. But thought I should get the conversation started.


src/main/java/decodes/db/DbEnum.java
```diff
@@ -369,5 +370,25 @@ public void clear()
...
+		sb.append(enumName).append("{")
+          .append("id=").append(getId().getValue()).append(",")
+		  .append("description=").append(description).append(",")
+		  .append("defaultValue=")
+		  	.append(defaultValue != null ? defaultValue : "not set").append(",")
+		  .append("values=[")
+		  .append(String.join(",",
+		  		              enumValues.stream()
+							            .map(EnumValue::getFullName)
+										.collect(Collectors.toList())
```
agilmore2, line 387: is this indention github's mess?

MikeNeilson:
Since I made a few minor tweaks (to the build, not the DAO code) and added the test I'll leave this open until sometime tomorrow before I merge it in.

Issue #90, opened 2022-08-20
Run Computations TS Group Evaluation is parameter order dependent


agilmore2:
**Describe the bug**
When running a group computation from the computation editor, the Select Timeseries GUI that shows up appears to key off of the first parameter listed in the computation editor, even if it's not a group parameter.

![image](https://user-images.githubusercontent.com/6852298/185726042-2eafaa09-4de5-4007-9577-1f00d586ea33.png)

If you delete the cur_cu parameter and run the computation, the Timeseries GUI looks a little better:
![image](https://user-images.githubusercontent.com/6852298/185726070-139f0caa-deec-48c7-9ab7-c205b651d8ec.png)

PR #95, opened 2022-09-07
Fix issue 21 reload algo before test run in GUI

### LINKED ISSUE #21: PythonAlgorithms python not reloaded automatically in compedit -> run
It's been observed that if you make a change to the Python of a PythonAlgorithm that you have to close the computation (both the run windows and the definition entry screen) and open them up again for that change to be picked up.

Steps.
1. create python algorithm 
2. create computation using algorithm
3. open run dialog and generate results
4. determine a change is required, leave the computation open in the gui, go change the python
5. go back to the computation and click run again
6. get same results

Expected outcome for 6, get different results using the new python script.

mikecovesw:
## Problem Description

Fixes #21
This was not limited to python algorithms. E.g. If some simple algorithm defined a default property value and it wasn't overloaded in the computation, the same behavior would appear.

## Solution

In the computation-run-GUI, reload the algorithm before preparing the computation for execution.

## how you tested the change

One of the regression tests used a python algorithm. I performed the steps described in the issue to recreate the problem and then to verify that it is fixed after changing the code.

## Where the following done:

### DIFF
--- a/src/main/java/decodes/tsdb/comprungui/CompRunGuiFrame.java
+++ b/src/main/java/decodes/tsdb/comprungui/CompRunGuiFrame.java
@@ -60,6 +60,7 @@ import javax.swing.border.EtchedBorder;
 import javax.swing.border.TitledBorder;
 import javax.swing.table.AbstractTableModel;
 
+import opendcs.dai.AlgorithmDAI;
 import opendcs.dai.TimeSeriesDAI;
 import opendcs.dai.TsGroupDAI;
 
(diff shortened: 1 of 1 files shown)


MikeNeilson (approved):

agilmore2:
Will this break the workaround for #90?

MikeNeilson:
I don't think it would have much affect on #90. This issue isn't with the gui, but making when Run gets called what's it's actually running the the latest information from the database.

The question would be would comp.setAlgorithm do enough to either harm or or correct the situation. I suspect neither.

Issue #100, opened 2022-09-11
More user control of logging system


MikeNeilson:
**Is your feature request related to a problem? Please describe.**
I'm getting ready to do some "cloud" (mostly just running in docker containers) development and need the logs to output on stdout. With the way docker works there wouldn't be a stable place to put file based logs.

I've made a simple tweak to allow specifiying a default log class to override the initial StderrLogger however, every application immediately goes and sets a file logger based on either a default filename or name passed in. 
There is a null/empty check in the code of the file name but due to the way the command line handling is down, input required, it would never bypass. 

**Describe the solution you'd like**
Something more consistent.

**Describe alternatives you've considered**

- Shift everything to java.util.logging, create listeners as appropriate for things like the comp editor stack trace.
- Shift everything to java.util.logging, create wrappers for all the existing log code (plus the listeners above.)
- Add a new command line switch to be specific "no log file" so that the user specified logger sticks.
- Have logger check and see if it *should* be overriden and ignore the setLogger call.

**Additional context**
The logger type *MUST* be set before the first value logged so it's passed in with a java -D property flag. So it wouldn't make sense to add a new command line switch to the command line parser.

I think I'm going to go forward with the 4th option, if the user set it, it makes sense to lock it in place. I'm also not opposed to the work to to 1 or 2. The logging frameworks have come a ways and there's a handful of useful helper libraries (like google flogger) that add a simple extra layer to top to keep code a little cleaner.

PR #101, opened 2022-09-12
Feature/stdout logger

MikeNeilson:
## Problem Description

Fixes/Implements #100.

## Solution

Went with Option 4. 

## how you tested the change

Ran dbedit to make sure logger wasn't reset. Still need to test compproc and other applications. But they all seem to hook into the logging system the same way.

I wasn't planing to worry about the GUI applications, the use case for this is all server side so fact that the things like the Trace dialog capture log output shouldn't be affected.

## Where the following done:

   - [ ] Unit tests that run during ant test *I Plan to do this so that test logging is captured*
- [ ] Was relevant documentation updated? //TODO

### DIFF
--- a/src/main/java/ilex/util/Logger.java
+++ b/src/main/java/ilex/util/Logger.java
@@ -8,6 +8,8 @@ import java.util.TimeZone;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.io.*;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 
 /**
 * Abstract logging mechanism.  Most code can be written to write log
@@ -32,6 +34,7 @@ import java.io.*;
 */
 public abstract class Logger
 {
+	private static String DEFAULT_LOGGER="ilex.util.StderrLogger";
 	protected static TimeZone tz;
 	/** Process name to be included in log messages (optional). */
 	protected String procName;
@@ -79,6 +82,9 @@ public abstract class Logger
 	*/
 	public static final int E_DEFAULT_MIN_LOG_PRIORITY = E_INFORMATION;
 
+	/** If true, calls to setLogger will silently fail */
+	private static boolean preventChange = false;
+
 	/** If true, include priority in output messages. */
 	protected boolean usePriority;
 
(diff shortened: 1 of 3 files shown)


MikeNeilson:
There aren't specifically tests of the mechanism, but I'll be using it when I automate the regression tests so I can capture input.

src/main/java/ilex/util/Logger.java
```diff
@@ -128,12 +138,49 @@ public static void setLogger( Logger logger )
...
+	}
+
+	/**
+	 * Get instance of logger. Create with given process name
+	 * if it doesn't exist yet. 
+	 * Process name will not be reset on subsequent calls.
+	 * @param procName
+	 * @return
+	 */
+	public static Logger instance(String procName)
```
adamkorynta, line 152: not `synchronized` because this only happens on application startup?

src/main/java/ilex/util/Logger.java
```diff
@@ -128,12 +138,49 @@ public static void setLogger( Logger logger )
...
+			preventChange = System.getProperty("opendcs.logger",null) != null;
+			String loggerClassName = System.getProperty("opendcs.logger", "ilex.util.StderrLogger");			
+			System.err.println("Loading Logger Class " + loggerClassName);
+			Class<?> loggerClass = Class.forName(loggerClassName);
+			Constructor<?> constructor = loggerClass.getDeclaredConstructor(String.class);
+			Logger logger = (Logger)constructor.newInstance(procName);
+			logger.info(String.format("Logger %s started",loggerClassName));
+			return logger;
+		} catch (ClassNotFoundException e) {
+			throw new RuntimeException("Unable to initialize logging system. Logger not found.",e);
```
adamkorynta, line 173: Worth adding in the system property here: 
`"Unable to initialize logging system. Logger not found for class specified with opendcs.logger system property"`

src/main/java/ilex/util/StdOutLogger.java
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+	* Logs a message.
+	* @param priority the priority
+	* @param text the formatted text
+	*/
+    @Override
+    public void doLog(int priority, String text)
+    {
+        System.out.println(standardMessage(priority, text));
```
adamkorynta, line 29: Why not JUL here? Seems like a good place to start as well as setting up for enterprise logging configurations with cloud deployments.

MikeNeilson: I didn't want to have extra config files involved at this point.

MikeNeilson:
Closing in favor of a bulk migration to JUL.

PR #118, opened 2022-09-22
Bugfix/routing

### LINKED ISSUE #117: CWMS Connection Pool mechanism leaking connections during RouteScheduler
**Describe the bug**
Eventually the system exits because no connections are available.

**To Reproduce**
1. setup system (making sure tracing is turned on)
2. run routsched
3. watch logs
4. see after freeConnection see number increase

**Expected behavior**
connections close system is stable.

 - OS: Ubuntu 22.04/Java 11
 - Version: master branch latest

Continuing to investigate.

MikeNeilson:
## Problem Description

Fixes #117 

## Solution

Migrating sql queries to the new handlers to wrap the connection open/close.

## how you tested the change

Running apps gui manually to create, list, and delete. 
Running command line apps and waiting for results.

## Where the following done:

### DIFF
--- a/src/main/java/decodes/cwms/CwmsConsumer.java
+++ b/src/main/java/decodes/cwms/CwmsConsumer.java
@@ -1,12 +1,12 @@
 /*
 *  $Id: CwmsConsumer.java,v 1.13 2020/01/31 19:30:23 mmaloney Exp $
-*  
+*
 *  This is open-source software written by ILEX Engineering, Inc., under
 *  contract to the federal government. You are free to copy and use this
 *  source code for your own purposes, except that no part of this source
 *  code may be claimed to be proprietary.
 *
-*  Except for specific contractual terms between ILEX and the federal 
+*  Except for specific contractual terms between ILEX and the federal
 *  government, this source code is provided completely without warranty.
 *  For more information contact: info@ilexeng.com
 *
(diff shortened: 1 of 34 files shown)


MikeNeilson:
Update:

Still getting "exceeded maximum connect time, you are being logged off" (database is set to 5 minute max connect) in places, but it's not tanking the entirty of the RoutingScheduler anymore. Is taking out compproc.

The may be an actually setting on the pool itself and not the pool code that was created. Still working out how to deal with it though. The Connections from the pool are supposed to have a validation check run on them on borrow but it's possible it's just a timing issue.

MikeNeilson:
For the record I am aware of the irony of having just merged a file that states we don't want PRs that touch almost everything... but so far this has been a game of whack-a-mole. Does feel like it's getting better though.

MikeNeilson:
FYI, in a few places, especially ComputationDAO I started calling the other DAOs instance of the lambda functions.

Better design might be to just separate the lambdas from the DAO itself or have versions that take a connection that the DaoVersion would call but for now that was the easiest way to ensure we weren't pulling a bunch of extra connections from the pool.

NOTE: I know this doesn't compile right now. done for the day and wanted to make sure I had a backup.

MikeNeilson:
For the curious, I think I've got it stable. That said while I'm storing data comps don't seem to be running anymore and the depends list seems a bit small.... but I'm getting way less connection timeout or other sql errors.

MikeNeilson:
Okay, that got out-of-hand. I stand by the intent but I discovered after I got so far I was just breaking things and have taken a new approach.

PR #120, opened 2022-10-14
Bugfix/routing2

### LINKED ISSUE #117: CWMS Connection Pool mechanism leaking connections during RouteScheduler

MikeNeilson:
## Problem Description

Fixes #117. 

## Solution

Reduce duplication in logic and allow better tracking to ensure connections are regularly returned to the pool for recycling.

## how you tested the change

Running programs with Oracle connect_time set to 5 minutes. routsched, compproc, compdepends, gui apps.

## Where the following done:


MikeNeilson:
This is draft number 2, still working out some kinks but the initial changes will make it easier to track down the problem I'm having.

src/main/java/decodes/cwms/CwmsConnectionPool.java
```diff
@@ -0,0 +1,399 @@
...
+        return ret;
+    }
+
+    /**
+	 * Converts a CWMS String Office ID to the numeric office Code.
+	 * @param con the Connection
+	 * @param officeId the String office ID
+	 * @return the office code as a DbKey or Constants.undefinedId if no match.
+	 */
+	private DbKey officeId2code(Connection con, String officeId) throws SQLException
```
adamkorynta, line 312: Small efficiency here: we could cache the mapping of officeId -> code since I doubt the codes would ever change throughout the lifetime of the application.

src/main/java/decodes/cwms/CwmsConnectionPool.java
```diff
@@ -0,0 +1,399 @@
...
+    //				+ "schema=CCP and table=PLATFORMCONFIG");
+            testStmt.execute();
+        }
+        catch (SQLException ex)
+        {
+            errMsg = "Error setting VPD context for '" + dbOfficeId + "': " + ex;
+            log.severe(errMsg);
+            System.err.println(errMsg);
+            ex.printStackTrace(System.err);
+            throw new DbIoException(errMsg);
```
adamkorynta, line 382: Whole lot of logging the same error message here.

src/main/java/decodes/cwms/CwmsConnectionPool.java
```diff
@@ -0,0 +1,399 @@
...
+        }
+        return td;
+	}
+
+
+    public Connection getConnection() throws SQLException
+    {
+        connectionsRequested++;
+        Connection conn = pool.getConnection(info.getLoginInfo());
+        WrappedConnection wc = new WrappedConnection(conn,(c)->CwmsDbConnectionPool.close(c),Optional.of(this));
```
adamkorynta, line 158: Shouldn't this call `returnConnection(c)` instead of passing to the `CwmsDbConnectionPool`?

src/main/java/decodes/cwms/CwmsConnectionPool.java
```diff
@@ -0,0 +1,399 @@
...
+		String ret = null;
+		try
+		{
+			ArrayList<StringPair>officePrivileges = determinePrivilegedOfficeIds(conn);
+			// MJM 2018-12-05 now determine the highest privilege level that this user has in
+			// the specified office ID:
+			log.finest("Office Privileges for user '" + user + "'");
+			for(StringPair op : officePrivileges)
+			{
+				if (op == null)
```
adamkorynta, line 191: logging at warning seems pretty high for these statements, especially without any context in the message.

src/main/java/decodes/cwms/CwmsSqlDatabaseIO.java
```diff
@@ -179,114 +193,113 @@ public void connectToDatabase(String sqlDbLocation)
...
+					}
+					catch(SQLException ex)
+					{
+						String msg = "Error in '" + q + "': " + ex
+							+ " -- will proceed anyway.";
+						Logger.instance().failure(msg + " " + ex);
+					}
+
+					cgl.setLoginSuccess(true);
+					Logger.instance().info(module +
```
adamkorynta, line 255: This logged message contains redundancies with previous "Connect to DECODES CWMS Database" message

src/main/java/decodes/cwms/CwmsSqlDatabaseIO.java
```diff
@@ -179,114 +193,113 @@ public void connectToDatabase(String sqlDbLocation)
...
-			String msg = "Error in '" + q + "': " + ex
-				+ " -- will proceed anyway.";
-			Logger.instance().failure(msg + " " + ex);
-		}
-		finally
-		{
-			try { st.close(); } catch(Exception ex) {}
-		}
+					String q = null;
+					try(Statement st = conn.createStatement();)
```
adamkorynta, line 226: recommend moving every statement into its own method to help debug connection management, as well as keeping the scope that is using the connection as minimal as possible.

src/main/java/decodes/cwms/CwmsSqlDatabaseIO.java
```diff
@@ -323,97 +336,34 @@ public IntervalDAI makeIntervalDAO()
...
-		}
-		catch (Exception ex)
-		{
-			String msg = module + ".getConnection() Cannot set VPD context username/officeId username='" 
-				+ conInfo.getLoginInfo().getUser()
-				+ "', officeId='" + dbOfficeId + "', officeCode=" + conInfo.getDbOfficeCode() + ", priv='"
-				+ conInfo.getDbOfficePrivilege() + "':" + ex;
-			Logger.instance().failure(msg);
-			try { CwmsDbConnectionPool.close(ret); } catch(Exception ex2) {}
+			Logger.instance().fatal("Unabled to get connection " + ex.getLocalizedMessage());
```
adamkorynta, line 350: Losing the `SQLException` stack trace info here

src/main/java/decodes/cwms/CwmsSqlDatabaseIO.java
```diff
@@ -323,97 +336,34 @@ public IntervalDAI makeIntervalDAO()
...
+		{
+			pool.returnConnection(con);
+		}
 		catch(SQLException ex)
 		{
-			Logger.instance().warning(module 
-				+ ".freeConnection() Error in CwmsDbConnectionPool.close: " + ex);
-			ex.printStackTrace(Logger.instance().getLogOutput() != null ? 
-				Logger.instance().getLogOutput() : System.err);
+			Logger.instance().warning("Unable to close returned connection: " + ex.getLocalizedMessage());
```
adamkorynta, line 365: Losing the SQLException stack trace here

src/main/java/decodes/cwms/CwmsSqlDatabaseIO.java
```diff
@@ -179,114 +193,113 @@ public void connectToDatabase(String sqlDbLocation)
...
 		try
 		{
-			st = conInfo.getConnection().createStatement();
-			
-			// Hard-code date & timestamp format for reads. Always use GMT.
-			q = "ALTER SESSION SET TIME_ZONE = 'GMT'";
-			Logger.instance().info(q);
-			st.execute(q);
+			pool = CwmsConnectionPool.getPoolFor(conInfo);
+			if( pool != null)
```
adamkorynta, line 206: This should never be null

src/main/java/decodes/cwms/CwmsSqlDatabaseIO.java
```diff
@@ -179,114 +193,113 @@ public void connectToDatabase(String sqlDbLocation)
...
-			Logger.instance().info(q);
-			st.execute(q);
+					// CWMS OPENDCS-16 for DB version >= 68, use old OracleSequenceKeyGenerator,
+					// which assumes a separate sequence for each table. Do not use CWMS_SEQ for anything.
+					int decodesDbVersion = getDecodesDatabaseVersion();
+					Logger.instance().info(module + " decodesDbVersion=" + decodesDbVersion);
+					keyGenerator = decodesDbVersion >= DecodesDatabaseVersion.DECODES_DB_68 ?
+							new OracleSequenceKeyGenerator() :
+							new CwmsSequenceKeyGenerator(decodesDbVersion);
+						TimeSeriesDb.readVersionInfo(this, conInfo.getConnection());
```
adamkorynta, line 222: `conInfo.getConnection()` is confusing here. I think this is the same `Connection` instance as the `conn` variable

src/main/java/opendcs/util/sql/WrappedConnection.java
```diff
@@ -0,0 +1,394 @@
...
+    private ZonedDateTime start = ZonedDateTime.now();
+    private ZonedDateTime end = null;
+
+    
+
+    public WrappedConnection(Connection realConnection){
+        this(realConnection, (c) -> {},Optional.ofNullable(null));
+    }
+
+    public WrappedConnection(Connection realConnection, final ThrowingConsumer<Connection,SQLException> onClose, Optional<CwmsConnectionPool> pool )
```
adamkorynta, line 63: `Optional` pool is strange to me. How would the connection not be null, but the pool it came from not be null. Also this is an anti-pattern for the usage of Optional, it shouldn't be used as a parameter.

src/main/java/opendcs/util/sql/WrappedConnection.java
```diff
@@ -0,0 +1,394 @@
...
+
+    @Override
+    public Statement createStatement() throws SQLException {
+        logPlainCreate();
+        return realConnection.createStatement();
+    }
+
+    private void logPlainCreate() {
+        log.debug2("Dev Msg: Plain create statement from:");
+        StackTraceElement stk[] = Thread.getAllStackTraces().get(Thread.currentThread());
```
adamkorynta, line 160: There should definitely be a `isLoggable()` check before snagging the stack traces. Also I would think that `Thread.currentThread().getStackTrace()` is cheaper than getting every thread's stack.

src/main/java/opendcs/util/sql/WrappedConnection.java
```diff
@@ -0,0 +1,394 @@
...
+    public void setTypeMap(Map<String, Class<?>> map) throws SQLException {
+        realConnection.setTypeMap(map);
+    }
+
+    public CompositeData asCompositeData() throws OpenDataException
+    {
+        String[] itemNames = new String[1];
+        Object[] values = new Object[1];
+        itemNames[0] = "Lifetime";
+        values[0] = Duration.between(start, end == null ? ZonedDateTime.now() : end)
```
adamkorynta, line 389: seems like it would be good to log here if end is null and the duration between start and now has gone past some threshold.

src/main/java/decodes/cwms/CwmsConnectionPool.java
MikeNeilson: true, but it's also only called the once.

adamkorynta (commented):
The change could use a good reformat to make them easier to read

src/main/java/decodes/cwms/CwmsConnectionPool.java
adamkorynta: Guess that's true, kinda hard to tell from the context of that single class

src/main/java/decodes/cwms/CwmsConnectionPool.java
```diff
@@ -0,0 +1,399 @@
...
+                throw new BadConnectException("Unable to set office context.",ex);
+            }
+        }
+        return ret;
+    }
+
+
+    private CwmsConnectionPool(CwmsConnectionInfo info, Connection conn) throws BadConnectException,SQLException,DbIoException
+    {
+        this.info = info;
```
adamkorynta, line 76: This constructor could use some love. I'd recommend removing all logic from here and moving to the factory method instead. Using one parameter to populate the other is adding a lot to the cognitive load here.

src/main/java/opendcs/util/sql/WrappedConnection.java
MikeNeilson: oh wow. I was doing currentThread in other places but totally missed that one.

src/main/java/opendcs/util/sql/WrappedConnection.java
adamkorynta: Yeah that idea makes sense, but weird dependency-wise having the `opendcs` package depend on a `cwms` package. `private Optional<ConnectionPool> getConnectionPool() { Optional.ofNullable(this.pool);}` would work for this case.

MikeNeilson: Good point, actually I think it can just be dropped in favor of a simple boolean. The original intent had it actually using the object but now it's just don't do extra operations unless told.

src/main/java/decodes/cwms/CwmsConnectionPool.java
```diff
@@ -0,0 +1,422 @@
...
+        try
+		{
+            String name = String.format("CwmsConnectionPool(%s/%s)",info.getLoginInfo().getUrl(),info.getLoginInfo().getUser());
+			ManagementFactory.getPlatformMBeanServer()
+							 .registerMBean(this, new ObjectName("org.opendcs:type=ConnectionPool,name=\""+name+"\""));
+		}
+		catch(JMException ex)
+		{
+			log.warning("Unable to register tracking bean " + ex.getLocalizedMessage());
```
adamkorynta, line 128: Should include the exception itself as to not lose the stack trace

MikeNeilson:
I do still have it running but thus far about 30 minutes with a CONNECT_TIME setting of 5 minute vs the 48 hours we have by requirement.

Plus in jconsole I see that the get/close is balanced.

Also running compproc and compdepends; both show  connection that stays open but those do tend to open and hold but each handles database failures fairly gracefully.

src/main/java/decodes/cwms/CwmsConnectionPool.java
MikeNeilson: No, it's operating on the actual connection.... granted that would've worked to and would clean up some logic I just implemented.

src/main/java/decodes/cwms/CwmsSqlDatabaseIO.java
MikeNeilson: it is, I was keeping as much original logic as possible before doing any larger tweaks. I think I can probably rip out the Connection variable from CwmsConnectionInfo now though.

src/main/java/opendcs/util/sql/WrappedConnection.java
MikeNeilson: cleaned up.

MikeNeilson:
okay, compdepends failed, slight error in logic wrapping the connection (some how null is getting to the wrapped connection object) and for routsched a class seems to have gone missing. still, ran for several hours. so progress.

src/main/java/decodes/cwms/CwmsSqlDatabaseIO.java
MikeNeilson: agreed, but slightly beyond scope of this PR.

src/main/java/opendcs/util/sql/WrappedConnection.java
MikeNeilson: Yeah, but the bean info doesn't currently even show up. I'll fix up the JMX values in a later PR.

MikeNeilson:
The loop in CompDepends has a balanced connections in/out. I suspect it's just failing at refreshCaches because I have a connect_time at 5 minutes. Then it appears to fail at the check for lock which in theory happens with a new connection so still a bit of investigation.

src/main/java/decodes/cwms/CwmsConnectionPool.java
MikeNeilson: thank you, that simplified things a lot.

MikeNeilson:
> okay, compdepends failed, slight error in logic wrapping the connection (some how null is getting to the wrapped connection object) and for routsched a class seems to have gone missing. still, ran for several hours. so progress.

The logic to retrieve a connection and deal with a CONNECT_TIME error or Pool Full was wrong. Did the sane thing and made it a loop instead of trying to be clever with a goto (though it was fun to try.)

MikeNeilson:
Okay, with the last to commits it's finally processing data. I will now let it run for the day and see how it survives.

MikeNeilson:
It ran overnight! Success!

@adamkorynta A final sanity check is in order.

MikeNeilson:
Though if I here nothing from people for a while I'm just going to merge it in.

ivysettings.xml
```diff
@@ -23,6 +23,7 @@
                 <!-- https://github.com/lobobrowser/Cobra/releases/download/1.0.2/Cobra.jar -->
                 <artifact pattern="https://github.com/[organisation]/[module]/releases/download/[revision]/[artifact].[ext]"/>
             </url>
+            <ibiblio name="local" m2compatible="true" root="file:///home/mike/.m2/repository"/>
```
adamkorynta, line 26: Local environment setting

src/main/java/decodes/cwms/CwmsSqlDatabaseIO.java
```diff
@@ -323,97 +337,34 @@ public IntervalDAI makeIntervalDAO()
 	public Connection getConnection()
 	{
 		// Called from DAOs and DbIo to get a new connection from the pool.
-		if (conInfo == null || conInfo.getLoginInfo() == null)
+		if (conInfo == null || conInfo.getLoginInfo() == null || pool == null)
 		{
 			Logger.instance().failure(module + ".getConnection -- loginInfo is null! DB not initialized?");
 			return null;
```
adamkorynta, line 343: Eventually I'd like to see this behavior changed to throwing a runtime exception. Trying to setup isolated tests becomes difficult when I'm getting null pointer exceptions on null connections.

src/main/java/decodes/cwms/CwmsSqlDatabaseIO.java
```diff
@@ -323,97 +337,34 @@ public IntervalDAI makeIntervalDAO()
...
-			String msg = module + ".getConnection() Cannot set VPD context username/officeId username='" 
-				+ conInfo.getLoginInfo().getUser()
-				+ "', officeId='" + dbOfficeId + "', officeCode=" + conInfo.getDbOfficeCode() + ", priv='"
-				+ conInfo.getDbOfficePrivilege() + "':" + ex;
-			Logger.instance().failure(msg);
-			try { CwmsDbConnectionPool.close(ret); } catch(Exception ex2) {}
-			return null;
-			// TODO Should this method throw BadConnectException?
-//			throw new BadConnectException(msg);
+			Logger.instance().fatal("Unabled to get connection " + ex.getLocalizedMessage());
```
adamkorynta, line 351: typo `Unable`, also do we really need a log line here when we are throwing the RuntimeException?

src/main/java/decodes/cwms/CwmsSqlDatabaseIO.java
MikeNeilson: That's a good point, there's not really any recovery upstream from this anyways.

adamkorynta:
Did we officially ditch JDK 7 support with the OpenDCS 7 release?

ivy.xml
```diff
@@ -83,6 +83,7 @@
         <dependency org="org.junit.jupiter" name="junit-jupiter-engine" rev="5.8.1" conf="test->default"/>
         <dependency org="commons-io" name="commons-io" rev="2.11.0" conf="test->default"/>
         <dependency org="org.apache.derby" name="derby" rev="10.14.2.0" conf="test->default"/>
+        <dependency org="org.mockito" name="mockito-core" rev="4.8.0" conf="test->default"/>
```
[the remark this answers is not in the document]
MikeNeilson: From an original attempt at setting this up. Should we leave it in those as a "yes, please go ahead and use this library" though?

MikeNeilson:
Yes, we ditched JDK 7 support. why?

ivy.xml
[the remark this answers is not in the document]
adamkorynta: I wouldn't, no need to add in an extra dependency and declare a version if it isn't being used.

adamkorynta:
I saw the lambdas, but didn't remember if that version cutoff was official.

ivy.xml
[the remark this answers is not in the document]
MikeNeilson: fair.

MikeNeilson:
Ah, okay. Yeah. I believe that was why. Way too useful to ignore.... and no one should be using JDK/JRE 7 anyways.

adamkorynta (approved):

PR #125, opened 2022-10-25
Implement dbimport -p argument to import only platform-related elements

mikecovesw:
Needed by BC Hydro

## Problem Description

BC Hydro is in a transitional phase where they need to continually update their Oracle opendcs database from an old DCS Toolkit database. For various reasons, they only want to update platform-related elements, which include Platforms, Sites, Configurations, Equipment Models, and Network lists. A "-p" boolean argument has been added to dbimport telling it to ignore anything but one of these elements.

## how you tested the change

Create empty database for testing. Import a snap of a working OpenDCS and TSDB database.
Use DbImport to import platform-related elements from a different database.
View changes in dbedit. Verify only platform related elements were modified/imported.

## Where the following done:


MikeNeilson (approved):
Only suggestion I would make is just delete the large blocks that are now commented out. Probably good to leave the comment about why they were remove for future devs though.

PR #128, opened 2022-11-01
#121 add in a test to confirm issue with tab vs space delimiter on decode

### LINKED ISSUE #121: DECODES format statement bug:  setMissing(ND)
I'm trying to use the OpenDCS web pull features to process USGS data using USGS Water Services.  For seasonal sites, the USGS uses the symbol "Ssn" to denote missing data.  I'm trying to decode that data as missing, not an error.  In the latest manual, I found the following notes:

image

I set up my format statement using setMissing(Ssn), but it does not read the value as missing.  If I manually change my RDB file to show "///" instead of "Ssn", a known symbol for missing values, the value is parsed correctly.  See screenshot below, specifically the "data" format statement for processing the values.  I tested putting the setMissing code in each of the format statements as well with the same result.

image

adamkorynta:
## Problem Description
Fixes #121. 

The RDB format is using a tab delimiter which and the decode format was using a space delimiter

## Solution

Adding in a whitespace delimiter.

## how you tested the change

Running unit tests to prove the error and then will add a mechanism for a whitespace delimiter and prove success.

## Where the following done:


adamkorynta:
From code inspection the delimiter ' \t' should have worked as well as '\\t ' based on the delimiter "indexOf(byte)" check here: https://github.com/opendcs/opendcs/blob/master/src/main/java/decodes/decoder/DataOperations.java#L604. It wasn't being escaped because the whitespace escaped characters were only being replaced if they were the only characters in the delimiter.

PR #130, opened 2022-11-05
Cleanup/decodes script

MikeNeilson:
## Problem Description

After looking through the Decodes code there wasn't a great separation of concerns. This begins the slow march towards
the slow march towards making DecodesScripts easier to expand with future needs.

Additionally looking at the test in #121 that is just way too much boiler plate to run a simple test.

DecodesScriptProvider or just "FormatStatementProvider/Reader" might be a better name than DecidesScriptReader. I'm not opposed to changing it.

## Solution

- Created a DecodesScriptReader interface and some baseline implementations (some still to do)
- Created a builder class that handles getting other information set
- Made formatStatements private
   - This exposed several places that took it upon themselves to add FormatStatements (not changed in this PR, more investigation required.)

## how you tested the change

Automated test now uses the StreamDecodesReader.
The configIO now uses a newly implemented SQLDecodesScriptReader.
Test both reading and existing script in DBEdit and writing a new script for a config. Ran rs with Platform/RoutingSpec that was able to retrieve and use the script.

## Where the following done:

   - [NA] Test procedure descriptions
- [NA] Was relevant documentation updated?
- [NA] Were relevant config element (e.g. XML data) updated as appropriate

### DIFF
--- a/src/main/java/decodes/db/DecodesScript.java
+++ b/src/main/java/decodes/db/DecodesScript.java
@@ -92,7 +92,9 @@ public class DecodesScript extends IdDatabaseObject
 
         DecodesSettings settings = DecodesSettings.instance();
         if ( settings.decodesFormatLabelMode.equals("case-sensitive" ) )
+        {
             labelIsCaseSensitive = true;
+        }
         scriptName = name;
         scriptType = Constants.scriptTypeDecodes;
         platformConfig = null;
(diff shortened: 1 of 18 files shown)


mikecovesw:
FWIW, I have always proceeded with extreme caution when modifying the code for DECODES scripts. There are thousands of scripts out in the wild, some doing things I never would have anticipated.

MikeNeilson:
Fair, but that really just justifies this work more. The easier it is to setup tests the easier it will be to take that work and make sure we either don't break it in the future or can clearly communicate appropriate adjustments.

I intentionally haven't alter any script behavior in this PR. This only changes how scripts get read in, not how they work after. I considered handling merging the format statements with the same label but that seemed too much even for this first PR.

MikeNeilson:
Superseded by #138

PR #131, opened 2022-11-07
Make sure Keygenerator is initialized.

MikeNeilson:
Additionally discovered exception wasn't actually logged in a few places.

## Problem Description

Missed a stop in the connection pool fix which caused compimport to fail.

## Solution

Make sure the KeyGenerator is initialized.

## how you tested the change

Ran compimport. Currently letting other processes run (routsched,depends, comproc)

## Where the following done:

### DIFF
--- a/docs/source/dev-docs.rst
+++ b/docs/source/dev-docs.rst
@@ -1,0 +1,48 @@
+##################
+Installation Guide
+##################
+
+This Document is part of the OpenDCS Software Suite for environmental
+data acquisition and processing. The project home is:
+https://github.com/opendcs/opendcs
+
+See INTENT.md at the project home for information on licensing.
+
+.. contents. Table of Contents
+   :depth: 3
+
+
+Overview
+========
+
+This is the initial start of the documentation and thus it's still quite empty
+If you are working on OpenDCS and think of something that should be in this section
+help expanding the docs are greatly appreciated.
+
+
+MBeans
+======
+
+We have started implementing JMX MBeans for components within OpenDCS. You can connect to the process
+using the jconsole application provided with your JDK to view the information.
+
+CWMS
+====
+
+MBeans
+------
+
+The cwms connection pool implements the ConnectionPool Mbean. This MBean provides a view into the connections 
+outstanding and available. Additional each Connection returned implements a WrappedConnectionMBean that shows
+the current lifetime and can show where the connection pool was opened from.
+
+Connection pool
+---------------
+
+CwmsDb using a connection pool mechanism. Leaks are a concern, if you working against a CWMS
+system you can turn pool tracing on for an application with the following java flags:
+
+    DECJ_MAXHEAP="-Dcwms.connection.pool.trace=true" routsched ...
+
+With tracing on the WrappedConnectionMBean will show where a connection was created from. This useful for identifing 
+what code to fix for connection pool leaks.
--- a/src/main/java/decodes/cwms/CwmsTimeSeriesDb.java
+++ b/src/main/java/decodes/cwms/CwmsTimeSeriesDb.java
@@ -859,6 +859,9 @@ public class CwmsTimeSeriesDb
 				+ " with officeID=" + dbOfficeId);
 
 				postConnectInit(appName, conn); // Make sure the versions and such are set
+				setConnection(conn);
+				setupKeyGenerator(); // TODO: this should take the connection as a parameter
+				setConnection(null);
 
 				cgl.setLoginSuccess(true);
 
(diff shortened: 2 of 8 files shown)


MikeNeilson:
Okay, also found that getting this in caused a connection to be held open on the consumer initialization. But this time I think I got it. Still need to let it run a few days on the live system it's on but initial behavior appears to be correct at least.

src/main/java/decodes/cwms/CwmsTimeSeriesDb.java
```diff
@@ -859,6 +859,9 @@ public DbKey connect( String appName, Properties credentials )
 				+ " with officeID=" + dbOfficeId);
 
 				postConnectInit(appName, conn); // Make sure the versions and such are set
+				setConnection(conn);
+				setupKeyGenerator(); // TODO: this should take the connection as a parameter
+				setConnection(null);
```
adamkorynta, line 864: should this be wrapped in a `finally` block in case `setupKeyGenerator()` fails?

src/main/java/decodes/cwms/CwmsTimeSeriesDb.java
```diff
@@ -859,6 +859,9 @@ public DbKey connect( String appName, Properties credentials )
 				+ " with officeID=" + dbOfficeId);
 
 				postConnectInit(appName, conn); // Make sure the versions and such are set
+				setConnection(conn);
+				setupKeyGenerator(); // TODO: this should take the connection as a parameter
```
adamkorynta, line 863: looks easy enough to overload and deprecate

MikeNeilson: overloads... why do I always forget overloads.

adamkorynta: actually I take that back, `KeyGeneratorFactory.makeKeyGenerator` isn't even using the parameterized connection which means the `setConnection(conn)` and `setConnection(null)` aren't needed

MikeNeilson:
Okay, this ran over the last week without zero user intervention. Came back in after vacation and the routing scheduler was still chugging along and storing data.

docs/source/dev-docs.rst
```diff
@@ -0,0 +1,48 @@
...
+See INTENT.md at the project home for information on licensing.
+
+.. contents. Table of Contents
+   :depth: 3
+
+
+Overview
+========
+
+This is the initial start of the documentation and thus it's still quite empty
```
adamkorynta, line 18: maybe worth calling out the GUI main class here? maybe some other important ones

MikeNeilson: Not sure what you mean, there are several "main gui" classes.

adamkorynta: Yeah that's I guess kinda my point. Hard to know where the good starting point is for development. I was thinking just "LauncherFrame" as a start

MikeNeilson: Oh I see, yeah. I get that. The whole document could use some expansion but I'd prefer to do that later. This particular fix has been needed for a while for CWMS and it's been stopping us from doing other releases.

adamkorynta: agreed

src/main/java/decodes/cwms/CwmsTimeSeriesDb.java
adamkorynta: Still this last comment

MikeNeilson: Yeah, I'll look into that more, there's some downstream stuff that might still use it so I was a little reluctant to change it more but a simple find reference should help with that.

MikeNeilson: Oh wait, I'll still look, but if you look at say, the "OpenTSDBDB" implementation it still expects the connection to be set once, but I could probably get away with just not calling those in this implementation.

MikeNeilson: Nevermind, that was all fairly easy to change, HDB does it's own thing so setKeyGenerator is only used in two places.

src/main/java/decodes/tsdb/TimeSeriesDb.java
```diff
@@ -611,14 +614,14 @@ public void setConnection(Connection conn)
...
-	protected void setupKeyGenerator()
+	protected void setupKeyGenerator(Connection providedConn)
 		throws BadConnectException
 	{
 		String keyGenClass = DecodesSettings.instance().sqlKeyGenerator;
 		try
 		{
 			keyGenerator = KeyGeneratorFactory.makeKeyGenerator(
-				keyGenClass, conn);
+				keyGenClass, providedConn);
```
adamkorynta, line 624: This connection is not needed

MikeNeilson: huh, yeah that isn't actually used.

MikeNeilson: Usage was removed from all points.

adamkorynta (approved):

adamkorynta (approved):

PR #138, opened 2022-12-08
Decodes Script Unit Test Framework.

MikeNeilson:
## Problem Description

Contains the changes of and thus supersedes #130. Since this includes 2 different decodes script tests the value should be a little clearer.

Testing changes to the DECODES code is cumbersome at best. With only manual or the current regression tests there is limited ability to check for various regressions. 

Additionally while going through the code there were several different places that had to check if objects have been properly initialized

## Solution

Minor modifications were made to the decodes scripts class and engine to support various sources of input so that resources files can easily be read in. These test sets contain the input data, script itself, sensor configuration, and expected results.

Additional modifications were made such that once `DecodesScript.Builder#build` is called the script *is* in a valid state or the build call will throw an exception. However, existing validity checks were not removed at this time.

This all runs standalone from the gui using `ant test`. As we get more tests in place this should save a lot of time and regression pain as certain tweaks are requested.

## how you tested the change

The changes are the tests themselves. 

## Where the following done:

   - [n/a] Test procedure descriptions
- [X] Was relevant documentation updated?
- [n/a] Were relevant config element (e.g. XML data) updated as appropriate


src/main/java/decodes/db/DecodesScript.java
```diff
@@ -16,538 +16,721 @@
...
-					continue nextPM;
-				}
-			}
-			// Fell through, no matching sensor:
-			Logger.instance().info("addPMs: The platform config for '" + new String(rm.getHeader()) 
-				+ "' does not have a sensor named '" + pmname + "' -- discarded.");
-		}
-	}
-
+    public static final Logger logger = Logger.instance();
```
adamkorynta, line 31: I don't see a reason for `logger` or `EMPTY_SCRIPT` to be public at this time (and the `empty()` method makes `EMPTY_SCRIPT` redundant.

src/main/java/decodes/db/DecodesScript.java
```diff
@@ -16,538 +16,721 @@
...
+    public static final EmptyDecodesScriptReader EMPTY_SCRIPT = new EmptyDecodesScriptReader();
+    // _id is stored in the IdDatabaseObject superclass.
+
+    /**
+     * Name of this script. Must be unique within a configuration.
+     */
+    public String scriptName;
```
adamkorynta, line 38: It was too much to make `scriptName`, `scriptType`, and `platformConfig` private?

src/main/java/decodes/db/DecodesScript.java
```diff
@@ -16,538 +16,721 @@
...
+    /** Links */
+    public PlatformConfig platformConfig;
+
+    /**
+    * This stores references to the ScriptSensors.  This data member will
+    * never be null.  Note that the index
+    * of each element of this Vector does not necessarily correspond with
+    * the sensor number.
+    */
+    public Vector<ScriptSensor> scriptSensors;
```
adamkorynta, line 52: Would love to see these typed as `List` while we're making big changes to this class.

src/main/java/decodes/db/DecodesScript.java
```diff
@@ -16,538 +16,721 @@
...
+        }
+    }
+
+    /**
+     * Start a builder for a decodes script with a given reader.
+     * @param scriptReader concrete implementation that provides the required statements
+     * @return a Builder object for further operations.
+     * @since 2022-11-05
+     */
+    public static DecodesScriptBuilder from(final DecodesScriptReader scriptReader)
```
adamkorynta, line 623: `final` is unnecessary > java 7

src/main/java/decodes/db/DecodesScript.java
```diff
@@ -16,538 +16,721 @@
...
+    }
+
+    /**
+     * Utility class to help with creating DecodesScript objects 
+     * and making sure they're in a valid state.
+     * @since 2022-11-05
+     */
+    public static class DecodesScriptBuilder
+    {
+        private DecodesScriptReader scriptReader;
```
adamkorynta, line 655: final variables

src/main/java/decodes/sql/ConfigListIO.java
```diff
@@ -812,16 +815,21 @@ private void ingestRow(ResultSet rs, PlatformConfig pc)
 			if (s != null && s.length() > 0)
 				dataOrder = s.charAt(0);
 		}
-
-		DecodesScript ds;
-
-		ds = new DecodesScript(pc, name);
-		ds.setDataOrder(dataOrder);
-		ds.setId(id);
-		readFormatStatements(ds);
```
adamkorynta, line 821: `readFormatStatements` is no longer used. Method can get deleted

adamkorynta (changes requested):

src/main/java/decodes/db/DecodesScript.java
MikeNeilson: huh, I'll have to investigate, I probably just missed it.

src/main/java/decodes/db/DecodesScript.java
MikeNeilson: That's very, that should really break much... in theory.

src/main/java/decodes/db/DecodesScript.java
MikeNeilson: Thanks, apparently I mixed that up with c++'s "const" keyword. Very different meaning.

src/main/java/decodes/sql/ConfigListIO.java
MikeNeilson: done.

src/main/java/decodes/dbeditor/ConfigEditPanel.java
```diff
@@ -539,29 +541,39 @@ void deleteSensorButton_actionPerformed(ActionEvent e)
...
-		if (dsEditDlg == null)
+		try
 		{
-			dsEditDlg = new DecodingScriptEditDialog(ds, this);
-			launchDialog(dsEditDlg);
+			DecodesScript ds = DecodesScript.empty()
+											.scriptName("")
+											.platformConfig(theConfig)
+											.build();
+			ds.scriptType = Constants.scriptTypeDecodes;
```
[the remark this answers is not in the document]
MikeNeilson: maybe, there's some oddness with the xml/DecodesScriptParser. but fixing those public variables should expose that.

src/main/java/decodes/db/DecodesScript.java
MikeNeilson: for now yes, I have to run through a large chunk of the gui stuff, which I want to do, but not until more tests are already in place.

src/main/java/decodes/xml/PlatformConfigParser.java
```diff
@@ -98,11 +98,22 @@ else if (localName.equalsIgnoreCase(XmlDbTags.DecodesScript_el))
 				throw new SAXException(XmlDbTags.DecodesScript_el + " without "
 					+ XmlDbTags.DecodesScript_scriptName_at +" attribute");
 
-            DecodesScript ds = new DecodesScript(platformConfig, nm);
+			try
+			{
+				// TODO: move this to build the script after creating an XML reader
```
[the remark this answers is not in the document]
MikeNeilson: Ah, right, I'll expand that comment, basically it screws with the ability to actually encapsulate the builder. At this point the xmlparser isn't necessarily finished.

src/main/java/decodes/dbeditor/ConfigEditPanel.java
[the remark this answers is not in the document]
MikeNeilson: I'm going to leave that for an additional pull request. there's a lot of places were though public variables are used, sadly some of them even for write.

MikeNeilson:
@adamkorynta I don't disagree with some of the encapsulation changes requested; however, I attempted a few and the changes got rather out of hand.

I would like to keep this PR to the minimum required code to support better decodes script testing. There are plans to do such modifications and improvements but I would like the tests themselves in place before we go too far.

adamkorynta:
Agreed, I was hoping those were low-hanging fruit

adamkorynta (approved):

PR #142, opened 2022-12-21
Add support to change message header timestamp for Iridium parsing

DanielTOsborne:
## Problem Description
The timestamp in Iridium messages are not set to the original transmission time, but (I presume) the time they are sent from the SBD gateway. This can be a difference of many minutes.

The result is that sometimes the messages can be stored at the wrong time. This is most noticeable if you use MOFF and MINT to parse the offset and interval from the message itself. Even if you use the truncation operator to tweak the message header time, MOFF doesn't use that value, and instead retrieves the original timestamp from the raw message.

## Solution

My solution involves adding two new field type specifiers: MHD and MHT.
MHD and MHT stand for "Message Header Date" and "Message Header Time", respectively.
The syntax for those is identical to the D and T format types, however these only operate on the message header timestamp, not the sensor values. These should be used at the beginning of a decodes script.

Note, that if the Truncation operator [ T() ] is used first, MHD and MHT will reset the truncation state, since it doesn't make sense to retain a truncated timestamp after explicitly setting the timestamp. This also means that the "first use" flag in the truncation operator is also reset, so a subsequent truncation will operate on the message header time, as it if never happened the first time. For this reason, it's recommended that if any truncation operators are used, they are executed after any MHD and MHT statements. This will ensure consistent behavior. If you do not use MHD or MHT, the behavior has not changed between MOFF and T().

## How the change was tested

I started with multiple messages that demonstrated the incorrect parsing behavior, and ran those against this fix.
I also tested the same data against the original F and D field types, to ensure their behavior had not changed.
A subset of those were included into the PR as additional test case data for the existing DECODES test framework.

Here's an example Iridium message:
`ID=000000000000000,TIME=22349145804,STAT=00,MO=05571,MT=00000,CDR=1F3BB331,LAT=0.00000,LON=0.00000,RAD=3 IE:0200D0 0:Batt 10 #15 12.61 12.61 12.59 12.57 :AT 10 #15 36.0 36.3 36.5 36.7 :RH 10 #15 100 100 100 100 :Stage 10 #15 1.36 1.36 1.36 1.36 :Dewpoint 10 #15 36.03 36.31 36.47 36.65 :YB 12.60 :YN Iridium Test :YD 221215145545`

Which can be parsed correctly with this DECODEs script:

| Label | Statement |
| ----- | --------- |
| Prime | S(300,':YD',Start),4x,C(12N,ParseTimeOnly),F(MHD,A,6,1),F(MHT,A,4),1p,T(M1),>Start |
| ParseTimeOnly | F(MHT,A,4),1p,T(M1),>Start |
| Start | S(300,':',End),x,f(F,A,8D' ') |
| Batt | f(moff,a,3D' '),2x,f(mint,a,2,8),4f(s,a,7D' ',8)>Start |
| AT | f(moff,a,3D' '),2x,f(mint,a,2,2),4f(s,a,7D' ',2)>Start |
| RH | f(moff,a,3D' '),2x,f(mint,a,2,4),4f(s,a,7D' ',4)>Start |
| Stage | f(moff,a,3D' '),2x,f(mint,a,2,1),4f(s,a,7D' ',1)>Start |
| Dewpoint | f(moff,a,3D' '),2x,f(mint,a,2,7),4f(s,a,7D' ',7)>Start |
| YB | f(moff-,"60",2),f(mint,"60",2,6),T(H),f(s,a,7D' ',6)>Start |
| BL | >YB |
| End |  |
| ERROR | >Start |

## Were the following done:

- [ ✔️ ] Tests. Check all that apply:
   - [ ✔️ ] Unit tests that run during ant test
   - [ n/a ] Test procedure descriptions
- [ ✔️ ] Was relevant documentation updated?
- [ n/a ] Were relevant config element (e.g. XML data) updated as appropriate


DanielTOsborne:
@mikecovesw I added you as a reviewer mostly because I was curious about your thoughts on this change (and your name was recommended by Github). However don't feel obligated to do so if you don't wish to, someone else will most certainly review this as well. 😃

PR #150, opened 2023-01-04
add missing dependency for cwms.

MikeNeilson:
## Problem Description

CWMS Usage failed on the latest release.

## Solution

Made sure dependency was properly included.

## how you tested the change

Installed into fresh directory and run Compproc/GUI apps. Apps ran correctly instead of failing immediately.

## Where the following done:


MikeNeilson:
Okay, *after* this release, I'm going to tag with an RC or something until there's some feedback.

Perhaps in the future we'll engage with a different method of labelling releases.

MikeNeilson:
Got feedback that this fix allows CWMS to run. Merging in to make official release so we can move on.

Issue #25, opened 2021-11-19
Migrate various unit tests within existing classes to junit5 unit tests


MikeNeilson:

MikeNeilson:
The low hanging fruit anyways, from the several I saw some of them expected much more infrastructure. but the more that's in any test automation the better.

adamkorynta:
The following PR's cover migrating some low hanging fruit to JUnit 5 as well as deleting tests that no longer make sense due to unused features or exploratory JDK testing.

https://github.com/opendcs/opendcs/pull/155
https://github.com/opendcs/opendcs/pull/156
https://github.com/opendcs/opendcs/pull/157

PR #157, opened 2023-01-10
convert domsat tests to JUnit5 integration tests

adamkorynta:
## Problem Description

Test classes are within production class packages with no automated mechanism to run them.

Fixes https://github.com/opendcs/opendcs/issues/25.

## Solution

Migrate test classes into parent test directory, splitting out unit, integration, and exploratory tests with best judgement. Automate as many tests as I can to be run alongside the GitHub Actions build.

This PR specifically covers the start of the integration Domsat tests. 

## how you tested the change

Ran the existing main methods to discover functionalities and implemented basic tests of functionality.

## Where the following done:

### DIFF
--- a/ivy.xml
+++ b/ivy.xml
@@ -1,7 +1,7 @@
 
 
 <ivy-module version="2.0" xmlns:e="https//ant.apache.org/ivy/extra">
-    <info organisation="com.opendcs" 
+    <info organisation="com.opendcs"
           module="opendcs"
           revision="${version}"
           e:snapshot="a${snapshot.time}"/>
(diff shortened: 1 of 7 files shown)


adamkorynta:
Need to determine if it is a priority to get these tests running from the build system. They require the relevant native libraries to be available at runtime, so for now they are disabled. If this is a priority, we can resolve within this PR, and if not, this can be the groundwork for future testing.

MikeNeilson (approved):

MikeNeilson:
Doesn't matter if someone has the libraries, the transmission doesn't exist anymore. It's all HRIT and DRGS now. So this is fine, we'll delete them later.

PR #163, opened 2023-01-20
Fixes/iridium timestamp

DanielTOsborne:
This is the same as #142 but resubmitted against the master branch.

## Problem Description
The timestamp in Iridium messages are not set to the original transmission time, but (I presume) the time they are sent from the SBD gateway. This can be a difference of many minutes.

The result is that sometimes the messages can be stored at the wrong time. This is most noticeable if you use MOFF and MINT to parse the offset and interval from the message itself. Even if you use the truncation operator to tweak the message header time, MOFF doesn't use that value, and instead retrieves the original timestamp from the raw message.

## Solution

My solution involves adding two new field type specifiers: MHD and MHT.
MHD and MHT stand for "Message Header Date" and "Message Header Time", respectively.
The syntax for those is identical to the D and T format types, however these only operate on the message header timestamp, not the sensor values. These should be used at the beginning of a decodes script.

Note, that if the Truncation operator [ T() ] is used first, MHD and MHT will reset the truncation state, since it doesn't make sense to retain a truncated timestamp after explicitly setting the timestamp. This also means that the "first use" flag in the truncation operator is also reset, so a subsequent truncation will operate on the message header time, as it if never happened the first time. For this reason, it's recommended that if any truncation operators are used, they are executed after any MHD and MHT statements. This will ensure consistent behavior. If you do not use MHD or MHT, the behavior has not changed between MOFF and T().

## How the change was tested

I started with multiple messages that demonstrated the incorrect parsing behavior, and ran those against this fix.
I also tested the same data against the original F and D field types, to ensure their behavior had not changed.
A subset of those were included into the PR as additional test case data for the existing DECODES test framework.

Here's an example Iridium message:
`ID=000000000000000,TIME=22349145804,STAT=00,MO=05571,MT=00000,CDR=1F3BB331,LAT=0.00000,LON=0.00000,RAD=3 IE:0200D0 0:Batt 10 #15 12.61 12.61 12.59 12.57 :AT 10 #15 36.0 36.3 36.5 36.7 :RH 10 #15 100 100 100 100 :Stage 10 #15 1.36 1.36 1.36 1.36 :Dewpoint 10 #15 36.03 36.31 36.47 36.65 :YB 12.60 :YN Iridium Test :YD 221215145545`

Which can be parsed correctly with this DECODEs script:

| Label | Statement |
| ----- | --------- |
| Prime | S(300,':YD',Start),4x,C(12N,ParseTimeOnly),F(MHD,A,6,1),F(MHT,A,4),1p,T(M1),>Start |
| ParseTimeOnly | F(MHT,A,4),1p,T(M1),>Start |
| Start | S(300,':',End),x,f(F,A,8D' ') |
| Batt | f(moff,a,3D' '),2x,f(mint,a,2,8),4f(s,a,7D' ',8)>Start |
| AT | f(moff,a,3D' '),2x,f(mint,a,2,2),4f(s,a,7D' ',2)>Start |
| RH | f(moff,a,3D' '),2x,f(mint,a,2,4),4f(s,a,7D' ',4)>Start |
| Stage | f(moff,a,3D' '),2x,f(mint,a,2,1),4f(s,a,7D' ',1)>Start |
| Dewpoint | f(moff,a,3D' '),2x,f(mint,a,2,7),4f(s,a,7D' ',7)>Start |
| YB | f(moff-,"60",2),f(mint,"60",2,6),T(H),f(s,a,7D' ',6)>Start |
| BL | >YB |
| End |  |
| ERROR | >Start |

## Were the following done:

- [ ✔️ ] Tests. Check all that apply:
   - [ ✔️ ] Unit tests that run during ant test
   - [ n/a ] Test procedure descriptions
- [ ✔️ ] Was relevant documentation updated?
- [ n/a ] Were relevant config element (e.g. XML data) updated as appropriate


mikecovesw:
Art Armour and I dealt with this issue a few years back. Since the time stamp is added by the gateway, and because of retries and the way messages are bounced around through the satellites & ground systems, it may be slightly off. The solution we used was to have the DCP logger place a time stamp in the message proper. Then on the DECODES side, parse that date/time with the various field operators.

MikeNeilson:
We discussed that solution as well. Unfortunately, if I remember correctly, the logger used doesn't allow easily changing the Iridium message.

Additionally, this could likely be useful for dealing with dates scraped off a website that isn't nearly as friendly with it's formatting it could be.

src/test/java/fixtures/DecodesHelper.java
```diff
@@ -140,13 +146,53 @@ public static Arguments getScript(String testName) throws DecodesScriptException
...
+            if( input != null )
+                return input;
+        }
+        return null;
+    }
+
+    private static String getMediumFromResource(final URL input)
+    {
+        String inputStr = input.toString();
+        if( inputStr.contains("input_") ) {
```
MikeNeilson, line 169: style issues,
- { not on next line
- missing brackets (though I honestly waffle on that one.)

src/test/java/fixtures/DecodesHelper.java
```diff
@@ -160,21 +206,36 @@ public static Arguments getScript(String testName) throws DecodesScriptException
...
         String tz = "UTC";
         tmpMedium.setTimeZone(tz);
-        tmpMedium.setMediumType(Constants.medium_EDL);
+
+        tmpMedium.setMediumType(mediumType);
         rawMessage.setMediumId("11111111");
-        EdlPMParser parser = new EdlPMParser();
+
+        PMParser parser = null;
+        if( mediumType.equals(Constants.medium_Goes) )
```
MikeNeilson, line 214: brackets

src/test/java/fixtures/DecodesHelper.java
```diff
@@ -160,21 +206,36 @@ public static Arguments getScript(String testName) throws DecodesScriptException
...
+        else
+            parser = new EdlPMParser();
+
         parser.parsePerformanceMeasurements(rawMessage);
-        Date timeStamp = new Date();
-        rawMessage.setTimeStamp(timeStamp);
+
+        // If the message contained a parseable timestamp (e.g. Iridium), use that as the raw timestamp
+        Variable ts = rawMessage.getPM(GoesPMParser.MESSAGE_TIME);
+        if( ts != null )
```
MikeNeilson, line 225: brackets

MikeNeilson (changes requested):
That's pretty much all total nitpicks about language usage.
Logic and such looks good to me.

src/main/java/decodes/datasource/IridiumPMParser.java
```diff
@@ -81,15 +81,22 @@ public void parsePerformanceMeasurements(RawMessage msg)
 			int idx = hdr.indexOf("ID=");
 			if (idx == -1)
 				throw new HeaderParseException(
-					"No ID (EMEI) in Iridium Header");
+					"No ID (EMEI) in iridium header");
```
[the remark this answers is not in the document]
DanielTOsborne: Lol, ok. It was mixed casing in this file, and I think I chose the direction with fewer changes.

DanielTOsborne:
> 

I found Iridium's message editing limited. You can add a time, but not a date, which means transmissions near midnight can have the incorrect date. The python editing seemed to only let you add to the end of a message, which I did use to add the date and time in its entirety, though I did find that the time that's executed is not the same as the transmission time and can be off by 15 seconds. If the transmission happens at like :45s after the minute, then the time is still off if you use MOFF to read the offset data from the message, as it takes ~15s to transmit, so the offset is incremented an extra minute from the time written to the message. All in all, Iridium is mess IMO.

It might be possible, but I couldn't figure out how to add the full time to each message sample. I'd also like to avoid that anyway since Iridium messages are much more limited in size.

src/test/java/fixtures/DecodesHelper.java
DanielTOsborne: What bracket is missing here?

MikeNeilson: starting on line 173. for line 169 it's end of line instead of newline.

DanielTOsborne:
I missed a style change at https://github.com/opendcs/opendcs/pull/163/files#diff-4f75ac5641321310245b0df2b6b2de14751690b8fe36412463c617a5189ad476R159, so I amended the commit for it.

PR #173, opened 2023-02-01
Enchancement/auth source spi

### LINKED ISSUE #85: Additional options for handling credentials
While our current .decodes.auth file works for most uses cases we should expand to allow different styles of passing in credentails.
Such as environment variables or even plain text files to support tools like docker and kubernetes more natively.

a new `DbAuth` property should be added. with different forms:

- DbAuth=file+opendcs://$DECODES_USERDIR/.decodes.auth
- DbAuth=file+<database>://<path to file> where <database> could be cwms, hdb, etc to indicate be specific info... or just file:// and it gets passed to the database class for processing.
- DbAuth=env and each database needs to describe what parameters it requires.

This passed the control of reading and perhaps writing the credentials file to the individual database components. Which should allow a lot of flexibility in the future for things like PKI based auth or even osAuth which I saw in the properties file.

None.

MikeNeilson:
## Problem Description

Fixes #85. 

## Solution

Created a java service provider interface for authentication sources. SPI helps with seperation of concerns for different versions of things, the Auth source was good low hanging fruit for it's introduction.

NOTE: the Primary benefit of the SPI mechanism is the ability of third-party jars to introduce implementation without code change in the core unit. For example one could now create a jar dedicate to set, getting secrets out of a kubernetes property system and OpenDCS itself requires no changes.

User Configuration changes are not required, a specific format is required in the dbAuthFile property string to trigger what source to use, without a `:` in the property value the default UserAuthFile will be used.

## how you tested the change

Unit tests were created

## Where the following done:


src/main/java/decodes/launcher/DecodesPropsPanel.java
```diff
@@ -265,7 +262,7 @@ private void dbPasswordButtonPressed()
 			String afn = propsEditPanel.getProperty("DbAuthFile");
 			if (afn == null)
 				afn = "$HOME/.decodes.auth";
-			afn = EnvExpander.expand(afn);
+			afn = EnvExpander.expand(afn);
```
agilmore2, line 265: whitespace again

src/main/java/lrgs/db/LrgsDatabaseThread.java
```diff
@@ -156,27 +157,22 @@ private synchronized void attemptConnect()
 		lastConnectAttempt = System.currentTimeMillis();
 
 		// Get username & password
-		String username = "lrgs_adm";
-		String password = "";
-		String authFileName = EnvExpander.expand("$LRGSHOME/.lrgsdb.auth");
-		UserAuthFile authFile = new UserAuthFile(authFileName);
+		String username = "lrgs_adm";
```
agilmore2, line 160: Is this default user documented anywhere? I'm not sure it actually survives the following code.

agilmore2 (approved):
Exciting to see! Minor comments do not need to be addressed to merge

src/main/java/lrgs/db/LrgsDatabaseThread.java
MikeNeilson: hmm, I'm going to have to check that again. I don't think I created it, just shifted it around. But perhaps now is a good time to clean up that initial user and password generation.

MikeNeilson: you're write, it doesn't survive the logic. I think it might have originally but i've gone ahead and just removed it.

PR #183, opened 2023-02-21
created task to enable signing artifacts for maven central upload.

MikeNeilson:
## Problem Description

I'd like the opendcs.jar file (an future split up files) to be available on maven central. This requires signing the artifacts with w GPG key in a specific way.

## Solution

Created a task that would generation the appropriate bundle:

- pom file to list out dependencies
- opendcs-${version}.jar
- opendcs-${version}-javadoc.jar
- opendcs-${version}-sources.jar

And assosciated .asc files (the signature) if gpg is available and a key is available.

This also generates the javadocs.

## how you tested the change

Manual verification. No code changes.

## Where the following done:

- [n/a] Tests. Check all that apply:
- [n/a] Was relevant documentation updated?
- [n/a] Were relevant config element (e.g. XML data) updated as appropriate


common.xml
```diff
@@ -24,7 +26,9 @@
 	<property name="build.lib" value="${build.dir}/lib"/>
 	<property name="build.test.classes" value="${build.dir}/test-classes"/>
 	<property name="build.test.lib" value="${build.dir}/test-libs"/>
-	<property name="dist.jar" value="${build.lib}/opendcs.jar"/>		
+	<property name="dist.jar" value="${build.lib}/opendcs.jar"/>
+
+    <property name="build.release.dir" value="${build.dir}/release"/>
```
agilmore2, line 31: whitespace strikes again

common.xml
```diff
@@ -129,4 +128,17 @@
 		<taskdef name="junitlauncher" classname="org.apache.tools.ant.taskdefs.optional.junitlauncher.confined.JUnitLauncherTask"
 			classpath="${build.dir}/ant-junitlauncher.jar"/>
 	</target>
+
+
+    <macrodef name="gpgsignfile">
+        <attribute name="target"/>
+        <sequential>            
+            <exec executable="gpg" dir="${build.release.dir}" failonerror="true">
+                <arg value="-ab"/>
```
agilmore2, line 137: how is the GPG private key protected?

common.xml
agilmore2: there's a few more new tab count/line break oddities in here, lines 196, 216

build.xml
```diff
@@ -173,6 +173,51 @@ NOTE at this time jdk8 must be used as there is still a dependency on tools.jar
 
 	<target name="all" depends="clean,jar"
 		description="Cleans, compiles, and builds the distrubtion Jar file."/>
```
agilmore2, line 175: fix the spelling while we're here

build.xml
```diff
@@ -173,6 +173,51 @@ NOTE at this time jdk8 must be used as there is still a dependency on tools.jar
...
+		     basedir="${build.dir}/javadocs"
+			 update="true"/>
+		<jar jarfile="${build.release.dir}/opendcs-${version}-sources.jar">
+			<fileset dir="${src.dir}" includes="**/*.java"/>
+		</jar>
+		
+		<available property="have.gpg" file="gpg" filepath="${PATH}"/>
+		<if>
+            <not>
+                <isset property="have.python"/>
```
agilmore2, line 205: why checking if python is available in the middle of this gpg check?

agilmore2 (changes requested):
whitespace and a couple of questions

common.xml
MikeNeilson: It's expected to be in the user's directory. Unfortunately there's no way to for our build to enforce a developer having protections. 

The next step would be to add a KEYS file to the repository and provide instructions to those who would want to verify.

build.xml
MikeNeilson: oops, that was copy pasted xml and it appears I missed a spot.

common.xml
MikeNeilson: Cleaned up the whole file. Not even sure how/why the tabs not in there.

build.xml
MikeNeilson: done.

PR #184, opened 2023-02-22
Add exit(0) handling to python algorithm calls. Fixes #149.

### LINKED ISSUE #149: catch in PythonAlgorithm is too broad
https://github.com/opendcs/opendcs/blob/792c8ab4a9c42be14e34a430340942e9cb3f475b/src/main/java/decodes/tsdb/algo/PythonAlgorithm.java#L311

This catches even the SystemExit Exception raised when sys.exit(0) is called in the python code, which makes quiet short-circuit exits of the code impossible. 

I propose a more specific catch of SystemExit checking for a zero error code that's silent thereafter.

agilmore2:
## Problem Description
Fixes #149

## Solution

Figured out how to determine if a caught exception is the result of a python exit(0) call via
https://github.com/jython/jython/blob/master/src/org/python/core/Py.java#L282

## how you tested the change

New unit test.

## Where the following done:

   - [X] Unit tests that run during ant test
- [N/A ] Were relevant config element (e.g. XML data) updated as appropriate

### DIFF
--- a/src/main/java/decodes/tsdb/algo/PythonAlgorithm.java
+++ b/src/main/java/decodes/tsdb/algo/PythonAlgorithm.java
@@ -316,7 +316,7 @@ public class PythonAlgorithm
 					PyException pe = ((PyException) ex);
 					if (pe.type == Py.SystemExit && PyException.isExceptionInstance(pe.value)
 							&& ((PyObject) pe.value).__findattr__("code").asInt() == 0) {
-						debug3("afterScript exited with system exit and zero exit code");
+						debug3("beforeScript exited with system exit and zero exit code");
 						return;
 					}
 				}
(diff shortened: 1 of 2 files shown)


src/main/java/decodes/tsdb/algo/PythonAlgorithm.java
```diff
@@ -310,6 +312,14 @@ protected void beforeTimeSlices()
 			}
 			catch(Exception ex)
 			{
+				if (ex instanceof PyException) {
```
MikeNeilson, line 315: Would it make more sense to just catch PyException directly?

MikeNeilson (approved):

src/main/java/decodes/tsdb/algo/PythonAlgorithm.java
agilmore2: I see a couple ways this should be improved, thanks!

PR #199, opened 2023-03-13
Quick for XML decodes script processing. (that yes, I totally broke.)

MikeNeilson:
## Problem Description

Did not sufficiently implement new decodes script reader into the XML Pars

## Solution

The implementation leaves a bit to be desired, but it works. NOAA is working on updates to their decoding so getting something in quick since the XML processing is just broken.

Next steps are properly encapsulating the DecodesScript anyways (the script was allowed to be unprepared by the time it was used to decode) and this gave me a better indication of what to do.

## how you tested the change

Was provided test data from NOAA, imported and ran routespec, got null pointer, implemented code, got appropriate data.
Also verified data loaded into the GUI and saved correctly to disk.

## Where the following done:

   - [x] Test procedure descriptions

I may investigate using a new XML Parser. Not to bash on what's there, it works and is how it was done in the day, but some of the modern XML parsers might (and I'll stress might here, it's still xml) make things a little easier.)

### DIFF
--- a/src/main/java/decodes/db/FormatStatement.java
+++ b/src/main/java/decodes/db/FormatStatement.java
@@ -136,4 +136,3 @@ public class FormatStatement extends DatabaseObject
 		return decodesScript;
 	}
 }
-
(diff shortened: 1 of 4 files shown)


agilmore2:
Help us understand what was broken and how? XML parsing of Decoding Scripts during database read and XML import?

MikeNeilson:
It would read, but it was never prepared properly so it failed when actually decoding a message. But only in the XML database, the For SQLDatabases it was a lot more obvious how to write the "DecodesScriptReader" implementations. 

This PR is basically me realizing what my my TODO comment meant.

PR #204, opened 2023-03-26
Fix more broken computation retry queries.

agilmore2:
## Problem Description

Fixes HDB Support ticket #1245. Computation retries look to have been broken for some time.

We see failures with this backtrace:
WARNING 03/21/23 02:39:52 hdb java.sql.SQLException: Missing IN or OUT parameter at index:: 3
	at oracle.jdbc.driver.OraclePreparedStatement.processCompletedBindRow(OraclePreparedStatement.java:1937)
	at oracle.jdbc.driver.OraclePreparedStatement.addBatch(OraclePreparedStatement.java:9361)
	at oracle.jdbc.driver.OraclePreparedStatementWrapper.addBatch(OraclePreparedStatementWrapper.java:1069)
	at decodes.tsdb.TimeSeriesDb.releaseNewData(TimeSeriesDb.java:1060)
	at decodes.tsdb.ComputationApp.runApp(ComputationApp.java:484)
	at decodes.tsdb.TsdbAppTemplate.execute(TsdbAppTemplate.java:250)
	at decodes.tsdb.ComputationApp.main(ComputationApp.java:673)

## Solution

Fix the query to not have 3 parameters.

## how you tested the change

Build and tested.

## Were the following done:

   - [X] Unit tests that run during ant test
   - [N/A] Test procedure descriptions
- [X] Were relevant config element (e.g. XML data) updated as appropriate

### DIFF
--- a/src/main/java/decodes/tsdb/TimeSeriesDb.java
+++ b/src/main/java/decodes/tsdb/TimeSeriesDb.java
@@ -1012,7 +1012,7 @@ public abstract class TimeSeriesDb
 					+ "and ((" + curTime + " - DATE_TIME_LOADED) > " // curTimeName
 					+ inter ); //String.format(maxCompRetryTimeFrmt, maxRetries) + ")"); //
 			PreparedStatement updateFailedRetry = tcon.prepareStatement(
-				"update CP_COMP_TASKLIST set FAIL_TIME = ? where RECORD_NUM = ? and "
+				"update CP_COMP_TASKLIST set FAIL_TIME = " + curTime + " where RECORD_NUM = ? and "
 			+	"( (" + curTime + " - DATE_TIME_LOADED) <= " + inter);
 			PreparedStatement updateFailTime = tcon.prepareStatement(
 				"UPDATE CP_COMP_TASKLIST "


src/main/java/decodes/tsdb/TimeSeriesDb.java
```diff
@@ -1012,7 +1012,7 @@ public void releaseNewData(DataCollection dc, TimeSeriesDAI tsDAO)
 					+ "and ((" + curTime + " - DATE_TIME_LOADED) > " // curTimeName
 					+ inter ); //String.format(maxCompRetryTimeFrmt, maxRetries) + ")"); //
 			PreparedStatement updateFailedRetry = tcon.prepareStatement(
-				"update CP_COMP_TASKLIST set FAIL_TIME = ? where RECORD_NUM = ? and "
+				"update CP_COMP_TASKLIST set FAIL_TIME = " + curTime + " where RECORD_NUM = ? and "
```
MikeNeilson, line 1015: That needs to stay a parameter but just be included. 

A value like time, which will change every call, here will cause the the Oracle planner to constantly flush query causing huge performance problems beyond just OpenDCS usage.

MikeNeilson (changes requested):
Knowing that's the issue though I can fix that up when I do the release later.

src/main/java/decodes/tsdb/TimeSeriesDb.java
MikeNeilson: well, unless that particular value is a string and not time. that is one of the problems with the diff. they miss some context. either way I'll make sure the appropriate fix goes in.

MikeNeilson (approved):
Okay, I should finish breakfast *before* commenting on code. That is a substitution for a command. I am going to chance those variables names though.

PR #210, opened 2023-04-03
Test/regression tests

MikeNeilson:
## Problem Description

Start the process of introducing complex integration tests into the build system.
Additionally the work already helped track down some additional issues with the PlatformParser code in regards to decodes scripts.

Still need to fill in the developer documentation and verify a few things but this is certainly enough for feedback and knowledge sharing.

## Solution

Starting with the Junit5 TestTemplate system create just enough infrastructure to get an XML database working, import sites, platforms, etc. and run routespec.

A few System.exit calls that weren't required were removed. The SystemStubs library can handle calls to exit but they weren't valid in this context so for now they're just caught so the location can get rendered in the test output.

Two tests were implemented, the SimpleDecodesTest and HydroJSON test as those were the first two found that were clearly not dependent on a timeseries database.

The next iteration (different PR) will involve the setup work for our reference database. 

SPI system is used to promote separation of the implementations. Additional future work will likely include simple annotations to allow the test infrastructure to know what's valid where. 

## how you tested the change

This is the test. 

## Where the following done:


MikeNeilson:
Since #208 does get fixed by the changes here I'm going to include that data as another test before merging this in.

MikeNeilson:
I'm going to close this in favor of the dynamic version. There's far more flexibility.

PR #213, opened 2023-04-16
Test/regression tests dynamic

MikeNeilson:
## Problem Description

We need automated integration tests

## Solution

Similar to #210  but uses the junit5 DynamicTest system. Currently manually registering all tests. I expected we'll automate the search for tests (a junit5 test-engine implementation was another strategy I considered). 

Manual setup of the "configuration" is required as that will be implementation and implementation configuration under test specific. But they're automatically loaded through the service loader.

The idea is to be able to treat these tests like a "Tech Compatibility Tookit" vs specific tests. However, that will necessitate figuring out how to generalize certain concepts better in both the system and some of the tests. But, you gotta start somehow.

## how you tested the change

This change is all tests, and code changes to support testing

## Where the following done:

   - [ x Integration tests created or modified that run during integration testing


MikeNeilson:
@opendcs/opendcs-core-devs 
It's not perfect, but it's a start. Next steps are automating more of the test function search and getting the TSDBs and getting more of the tests from the regression test suite running.

FYI, I added a new parameter to the junitreports to have useful names in the output... you'll all need to update to at least ant 1.10.12 for it to work. Worth it for the integration tests otherwise you end up with a name like "DecodesTest[1][2][3][1]..." which isn't exactly useful.

@baoyuyin Merging into master. Let me know if this works and I'll do the release. Probably tomorrow morning.

PR #226, opened 2023-05-10
Added Filtering to the lists

### LINKED ISSUE #187: Allow filtering in OpenDCS DECODES Lists
It can be very difficult to have to scroll through a very large list to find a specific Platform, Site, or Config (as we continue to expand our web pulls other tabs may become large and unwieldy, too).

In certain GUIs in OpenDCS, you can filter to a location to narrow down the list that is shown.  It would be nice to have a filter at the top of each tab in DECODES, where you can type a location (doesn't have to be as sophisticated as CWMSVue) or something, and have the list filter down to only show lines that have that location in them.

hoffmanjs:
## Problem Description

Fixes #187 

## Solution
Added drop down lists to allowing filtering on the Platform, Agency, Transport-ID, and Config columns.
Describe your solution.

## Where the following done:

### DIFF
--- a/src/main/java/decodes/dbeditor/PlatformListPanel.java
+++ b/src/main/java/decodes/dbeditor/PlatformListPanel.java
@@ -4,13 +4,24 @@
 package decodes.dbeditor;
 
 import java.awt.*;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
 import javax.swing.*;
 import javax.swing.table.*;
+import java.util.stream.Collectors;
 import java.util.Vector;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Enumeration;
 import java.util.ResourceBundle;
+import java.util.Vector;
+import java.util.List;
+import java.util.ArrayList;
+
+import javax.swing.table.TableModel;
+import javax.swing.table.TableRowSorter;
+import javax.swing.JComboBox;
+import javax.swing.JLabel;
 
 import ilex.util.LoadResourceBundle;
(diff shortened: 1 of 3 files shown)


src/main/java/decodes/dbeditor/PlatformListPanel.java
```diff
@@ -80,26 +277,38 @@ public Platform getSelectedPlatform()
...
-			{
+					dbeditLabels.getString("PlatformListPanel.selectOpen"));
+		else {
+			try {
 				p.read();
 				doOpen(p);
-			}
-			catch(Exception ex)
-			{
+			} catch (Exception ex) {
```
MikeNeilson, line 307: catch should be on next line.

the try and else above should have { on the next line.

src/main/java/decodes/dbeditor/PlatformListPanel.java
```diff
@@ -123,7 +332,7 @@ public void copyPressed()
 		Platform p = platformSelectPanel.getSelectedPlatform();
 		if (p == null)
 			TopFrame.instance().showError(
-				dbeditLabels.getString("PlatformListPanel.selectCopy"));
```
MikeNeilson, line 126: Add the { and } if changing area.

MikeNeilson (changes requested):
This looks good. I only have nitpicky issues with the code formatting.
But the overall logic and structure looks good to me.

MikeNeilson (approved):
Thanks for cleaning up the formatting!

PR #236, opened 2023-05-28
Testing/more integrationtests

MikeNeilson:
## Problem Description

Need at least some automated regression tests for the SQL databases

## Solution

Began process of implementing configuration hooks for the Postgres built in database.

## how you tested the change

The changes are the test, or to allow testing and thus tested.

## Where the following done:


MikeNeilson:
@opendcs/opendcs-core-devs 

Okay, so in the present state, this technically breaks the installer files. As i've modified the Postgres .sql set to be easier to handle within the test code.

My inclination is to also go ahead and just do a full migration to Flyway, barring any objections.
The short term goal is just to make testing and upgrades easier. The midterm is to unify the database schemas slightly and let the implementations feed some overriding configuration.

Flyway has a suitable default for the initial efforts, but also has hooks to so just about anything can be overriden. So work example we could implement the changes as a template.

That said, I'm not set on it beyond any change, it's just the tool I'm most familiar with and I like that it's easy to embed, so for example the initial db install could be done in the installer, or so could an upgrade.

It appears our choices are teh following:
1. [Flyway](https://flywaydb.org/)
2. [Liquibase](https://www.liquibase.org/)
3. Completely custom

3 is my least preferred option. 1 is my vote, as I like that' it's still generally just SQL (we may just need to work in some relatively minor implementation differences.)

agilmore2:
I'm onboard with the database representation migration, but I would like to see a stable testing setup beforehand.

MikeNeilson:
Conveniently those two go hand in hand.

MikeNeilson:
But perhaps for the moment, as I still need to get computation tests in place, it may be worth the effort of maintaining yet another schema directory so we can go the test without interfering with current operations.

Normally I wouldn't want to do that but for the moment we've got a lot going on, and our schema doesn't change that much.

MikeNeilson:
FYI, I'm going to have the SQL be separate in the test, and use flyway in the 7.x branch series. Since we're intentionally make breaking changes in 8.0 I'll move that to be flyway only so we can see and review a full sequence of things and it gets under test.

PR #239, opened 2023-06-14
Feature/lrgs health

### LINKED ISSUE #235: Simple HTTP health endpoint
Cloud services generally want a simple system to check the health of containers.

A simple HTTP services embedded with not generally customary "/health" and extra information.
It is also practical to add "/metrics" at the same time and at least expose limited JVM metrics while we begin adding others.

The docker file could have a health/status script embedded that various container engines can just use. However, we have other efforts that will benefit from having an HTTP service in place so I think just starting that is a good solution. The effort will be small so we should be able to easily revert if it proves problematic.

**Additional context**
I'm requesting infrastructure for some internal work and realized I need a way to access the status of each running instance and existing tools, such as complocklist and the event monitor aren't practical in the AWS ECR/ECS environment.

MikeNeilson:
## Problem Description

Fixes #235 . 

## Solution

Created simple javalin server and two end points. One is a simple 200 OK if the Lrgs is in a usable state or 500 otherwise.
The other endpoint is the normal HTML status page served over HTTP.

HTTP service is implemented as a LoadableLrgsInput.

## how you tested the change

Manually

## Where the following done:

### DIFF
--- a/docs/source/lrgs-userguide.rst
+++ b/docs/source/lrgs-userguide.rst
@@ -2160,6 +2160,36 @@ The display shows the following columns:
 -  **LRGS Version**: Version of the LRGS software running on the remote
    server.
 
+LRGS Web Interface
+------------------
+
+The LRGS now has a simple Healthcheck end point and access to the status page over HTTPs.
+At present the design is simple, readonly, and meant to be access from behind some sort of proxy 
+if TLS is required.
+
+To enable add the following elements to your lrgs.conf file
+
+   LrgsInput.httpStatus.class=org.opendcs.lrgs.http.LrgsHttpInterface
+   LrgsInput.httpStatus.enabled=true
+   LrgsInput.httpStatus.port=7000
+
+7000 is the default port and can be left off. Include the property and set the value is 
+required in your environment.
+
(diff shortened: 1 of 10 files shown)


docs/source/lrgs-userguide.rst
```diff
@@ -2160,6 +2160,36 @@ The display shows the following columns:
...
+if TLS is required.
+
+To enable add the following elements to your lrgs.conf file
+
+   LrgsInput.httpStatus.class=org.opendcs.lrgs.http.LrgsHttpInterface
+   LrgsInput.httpStatus.enabled=true
+   LrgsInput.httpStatus.port=7000
+
+7000 is the default port and can be left off. Include the property and set the value is 
+required in your environment.
```
agilmore2, line 2177: unclear, "if" required?

agilmore2 (commented):
Looks simple enough. I have little context to judge, but looks good to me.

docs/source/lrgs-userguide.rst
MikeNeilson: yep, good catch.

PR #242, opened 2023-06-15
Refactor/logging

### LINKED ISSUE #161: Logging -> JUL, keep what we have but allow more control?
Would anyone think I'm totally insane if I just went and did one big pull request to convert the logging to java.util.logging?

Assuming the following:

- I can find a clean way to handle the GUI elements (like the compedit/run/trace dialog
- Additional configuration *will not* be required. E.g. puting -dN and -l will behave as normal, just mapping as appropriate to the JUL bits.

I just went looking through the code to see about creating a logging system that can keep a number of files and, well, it's there already. But the echo system is not setup to let the user really user it. 

I'm thinking default mappings from the usual parameters, but if logging.properties exists in say, LRGSHOME or DCSTOOL_USERDIR, pick it up and use it. (Mapping the name of course. That may mean stealing the Tomcat JUL implementation instead of just using the default.)

We can discuss other frameworks later. However my inclination is no, as while currently all of OpenDCS is pure application, it would be good to leave the modules, like DECODES scripts and the like, open for consumption in other systems. 

Alternatively I could just keep going with #100, but honestly why maintain our own logging system if we don't have to?

NOTE: I have no idea when I would do this, but I don't suspect it would be difficult, just a smidge time consuming and drudgeworky.

MikeNeilson:
## Problem Description

Fixes #161 . 

## Solution

Replace logging statements with SLF4j statements. SLF4j is a frontend, it is not responsible for the logging, only making it easier (in some contexts otherwise really the same) to format and gather log information and pass it to an actual logging framework. 

Current design will not require changes to user configuration. For example -l and -dX and log size will behave as normal, however since the applications will be using JUL as the actual logging tool, a logging.properties file can be configured or picked up to allow a user more granularity.

Focusing on just LRGS for now. These changes are extremely initial and only done for the one place to allow easier conceptual review.

Things to do/figure out:
- DacqEvents 
- Trace dialog
- Date based log file naming? (I prefer it, JUL only does numbered, curious what the group thinks.)

Good news is, shouldn't be too hard. Just a combination of JUL Filters and Handlers, but they do need to be develops

NOTE: Currently outputs to a different log filename so I can verify I actually *have* gotten all the messages. Once I'm done "lrgslog" for example should be empty.

## how you tested the change

Manual verification, it's the log system; will try to figure out some tests though.

## Where the following done:


MikeNeilson:
Quick NOTE: I'm intennding this to go in after the lrgs health endpoint PR so to save myself some config hassle while testing that I branched off of that branch, ignore the Javalin related code while commenting.

Though I'll note that once I got the Cobra/LoboBrowser jar fiasco out Javalin was able to actually log since it uses SLF4j (actually a lot of our dependencies do.)

MikeNeilson:
Much like the java docs, this effort has to touch literally everything. I've modified the JavaLoggerAdapter class to be an always pass through and do thing a little more clean. It was easier than manipulating things the other way.

Once we've removed instances of logging through the ilex logger we can reintroduce using JUL for the actual file information. But I think ti would be better for that to take some time. Especially as I would like to pivot back to the database interaction issues.

MikeNeilson:
Okay, yet another change. I figured out this morning that creating a bridge logger to JUL wouldn't be that bad.

I considering going through SLF4J interface, which is doable, but somehow this seemed a bit more sensible. The idea is that this at least least JUL manage the logs but we don't have to change everything all at once.

If this seems reasonable to everyone I'll also make the required changes to the OpenTsdbApp template and command line args. 

Then we can slowly migrate classes to SLF4j over time as we update them and do other bug fixes.

MikeNeilson:
Well, this PR has really gotten away from me. I think I've implemented 3, or more, different ways to output the logging to a standard framework but i've also change too many classes using the logs to keep track off.

I'm going to leave this here for reference and comment though. But probably week after next I'm going to revisit and simplify.
I'm pretty happy with the tweaks I made to have the existing classes become a proxy to JUL but I think they should probably be wrappers to SLF4j so that the actual log creation can be handle outside of time.

Unfortunately tools like the LRGS have several real loggers attached to them; I'd like to find a way to keep that behavior while also allowing it to be override by a logging.properties file. 

However, step one is to go can and do just the tweaks to log management and creation, and then covert something a little easier to using SFL4j, like the connection pool class I made, or something. But made it does need to just be this drastic, I'm just really hoping not.

PR #246, opened 2023-06-20
Batch 1 of javadoc updates and simple cleanup.

MikeNeilson:
## Problem Description

During the building of releases the javadoc task flags a lot of errors. Most of them are just due to how javadoc works (can't use > or < has to be &gt; and &lt; now) and others were just information that got stale over time.

## Solution

Updated javadoc as appropriate, filled in what I could where things just didn't exist.
Also since I was doing through every file anyways I cleared out dead code and unused imports.

## how you tested the change

No functional changes.

## Where the following done:

### DIFF
--- a/src/main/java/decodes/comp/AreaComputation.java
+++ b/src/main/java/decodes/comp/AreaComputation.java
@@ -13,7 +13,7 @@ import java.util.Enumeration;
 
 /**
 * Implements area table computations.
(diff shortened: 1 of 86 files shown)


MikeNeilson:
I was going to do one big PR with everything, but I realized that would just get out-of-hand. So here's batch one.

agilmore2 (approved):

PR #259, opened 2023-07-13
Make sure tests actually fail the build if they fail.

MikeNeilson:
## Problem Description

while testing something I discovered that I copy pasted and forgot to change in one location a variable required for the `ant test` command to actually fail the build properly.

## Solution

Set test failure property variable to the correct value.

## how you tested the change

A test that was failing didn't fail the build. Once the property was correct the build failed as expected for that code.

## Where the following done:

### DIFF
--- a/src/test/java/org/opendcs/authentication/UserAuthFileTest.java
+++ b/src/test/java/org/opendcs/authentication/UserAuthFileTest.java
@@ -19,6 +19,7 @@ public class UserAuthFileTest
     @Test
     public void test_user_auth_file_default(@TempDir Path configDir) throws Exception
     {
+        System.out.println(configDir.resolve("test.txt").toString());
         final String USERNAME = "Testuser";
         final String PASSWORD = "Testpassword";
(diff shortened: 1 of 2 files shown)


src/test/java/org/opendcs/authentication/UserAuthFileTest.java
```diff
@@ -19,6 +19,7 @@ public class UserAuthFileTest
     @Test
     public void test_user_auth_file_default(@TempDir Path configDir) throws Exception
     {
+        System.out.println(configDir.resolve("test.txt").toString());
```
agilmore2, line 22: debug? If not, should have a little more context output.

agilmore2 (approved):

src/test/java/org/opendcs/authentication/UserAuthFileTest.java
MikeNeilson: Oops, that was supposed to be removed. Was trying to get some initial feedback as to what exactly was going on.

PR #260, opened 2023-07-14
Initial Release process documentation.

MikeNeilson:
## Problem Description

I realized we discussed the release process in the project discussion and came to a conclusion but I never actually wrote up the process.

## Solution

Create a RELEASES.md file that discusses what releases will be along with the steps for anyone with appropriate permissions to do so.

This is currently a work in process. The actual steps need more detail.. and possibly just an ant target; will investigate later.
But It seems to be at a good state to make sure I covered our intent from https://github.com/opendcs/opendcs/discussions/84.

## how you tested the change

I will eventually be doing the steps as I write it up.

## Where the following done:

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,0 +1,61 @@
+# OpenDCS Release Process
+
+This will document the release process. The initial sections will document user expectations
+for when and how releases are made. The remaining sections will detail how a release is generated
+so that the process is repeatable by multiple developers.
+
+
+## Releases
+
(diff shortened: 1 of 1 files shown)


ktarbet (approved):

ktarbet:
Looks OK to me.

MikeNeilson:
just some notes from doing the process in a new environment:

need:
-  gpg key and gnupg
- gnu make (not properly indicated in error message)
- python venv (indicated in error message)
- latex and latexpdf (indicated in error message)

PR #266, opened 2023-08-01
Refactor/logging attempt 2

### LINKED ISSUE #161: Logging -> JUL, keep what we have but allow more control?

MikeNeilson:
## Problem Description

Fixes #161 . 

Okay, this is the new attempt. Unfortunately it's still a lot of files. However; all of the changes mostly constrained to the actually Logging system and command line args parsers vs actually just implementing SLF4j everywhere so it should be easier to review.

## Solution

Implemented required hooks and changes to allow the ilex.util.Logging system to foward to SLF4j.
Implemented appropriate java.util.logging handlers and formatters to handle writing the log to files converting between the ilexLogging and JUL/SLF4j values and systems as required.

JavaLoggingAdapter no only does anything if java.util.logging.config.file is null. Otherwise we'd be overriding user requirements.

TODO (this PR):

[] - Implement a new Handler that will take the file name and create dated files OR keep the default behavior
[] - Verify all GUI elements still work. (Initial things did, Process monitor showed nothing - though it might have been doing that before.)
[] - update developer docs.

TODO (future PRs):

- Migrate the QueueLogger and TeeLogger systems to be JUL Handlers... or something. Need to investigate a little more.

## how you tested the change

Ran programs compproc, compdepends, lrgs, routingscheduler, laucher_start to verify logs were outputting correctly.

## Where the following done:


ktarbet:
Using the branch refactor/logging_atempt_2, I created a new installer, installed and launched (on windows)

Here is the output.

```
C:\project\opendcs>c:\bin\OPENDCS-7.0.9\bin\launcher_start.bat
Exception in thread "main" java.lang.NoSuchMethodError: 'org.slf4j.spi.LoggingEventBuilder org.slf4j.Logger.makeLoggingEventBuilder(org.slf4j.event.Level)'
        at org.opendcs.logging.IlexToSlf4jBridge.doLog(IlexToSlf4jBridge.java:27)
        at ilex.util.Logger.log(Logger.java:173)
        at ilex.util.Logger.debug1(Logger.java:357)
        at ilex.cmdline.ApplicationSettings.parseArgs(ApplicationSettings.java:98)
        at decodes.util.CmdLineArgs.parseArgs(CmdLineArgs.java:243)
        at decodes.launcher.LauncherFrame.main(LauncherFrame.java:1192)

**Comment by MikeNeilson** [comment]
check the install and make sure LoboBrowser and Cobra jars will be at the end of the classpath. That's required because both jar include class files for their dependencies and it's the wrong version of slf4j... amongst others.

**Comment by ktarbet** [comment]
I see those dependencies, however also zzzCobra without .jar extension.

```
C:\bin\OPENDCS-7.0.9\dep>dir /b *Cob*
Cobra-1.0.2.jar
zzzCobra-1.0.2


C:\bin\OPENDCS-7.0.9\dep>dir *Lobo* /b
zzzLoboBrowser-1.0.0.jar

MikeNeilson:
Hmm, might be a quirk of the windows build.

build.xml
```diff
@@ -364,7 +364,7 @@ NOTE at this time jdk8 must be used as there is still a dependency on tools.jar
                 </restrict>
                 <chainedmapper>
                     <flattenmapper/>
-                    <regexpmapper from="^(LoboBrowser-1.0.0.jar)" to="zzz\1"/>
+                    <regexpmapper from="^(Cobra-1.0.2|LoboBrowser-1.0.0.jar)" to="zzz\1"/>
```
ktarbet, line 367: need  .jar on the Corbra-1.0.2

build.xml
```diff
@@ -374,7 +374,7 @@ NOTE at this time jdk8 must be used as there is still a dependency on tools.jar
                 </restrict>
                 <chainedmapper>
                     <flattenmapper/>
-                    <regexpmapper from="^(?!LoboBrowser-1.0.0.jar)(.*)" to="\1"/>
+                    <regexpmapper from="^(?!Cobra-1.0.2|LoboBrowser-1.0.0.jar)(.*)" to="\1"/>
```
ktarbet, line 377: need  .jar on the Corbra-1.0.2   (that fixed my issue on windows)

MikeNeilson: oh, oops. I must have though that the .jar was outside the grouping. Good find.

src/main/java/decodes/dcpmon/XRWriteThread.java
```diff
@@ -77,6 +77,7 @@ class XRWrapper
 
 	public XRWriteThread()
 	{
+		super("XRWriteThread");
```
ktarbet, line 80: For debugging would we need a specific thread name for each Instance?

MikeNeilson: For this particular component there's only ever one instance. But having a name anyways helps with sorting through threads.

src/main/java/ilex/util/FileLogger.java
```diff
@@ -19,197 +28,136 @@
...
+    * to an existing file when it starts up.
+    * Set to false to cause the constructor to zero and start a new log.
+    */
+    public static boolean appendFlag = true;    
+
+    /**
+    * Construct with a process name and a filename.
+    * @param procName Name of this process, to appear in each log message.
+    * @param filename Name of log file.
+    * @throws FileNotFoundException if can't open file
```
ktarbet, line 64: -- comment update
* @throws IOException if can't open file

src/main/java/ilex/util/SequenceFileLogger.java
```diff
@@ -16,55 +16,26 @@
...
 {
-	private int numOldLogs;
 
 	/**
 	* Construct with a process name and a filename.
 	* @param procName Name of this process, to appear in each log message.
 	* @param filename Name of log file.
+	* @param maxLength how many bytes each log can be
+	* @param count how many total files we will keep
 	* @throws FileNotFoundException if can't open file
```
ktarbet, line 26: comment should use IOException

src/main/java/ilex/util/SequenceFileLogger.java
ktarbet, line 1: It feels like SequenceFileLogger class isn't really needed, and could be swapped out for FileLogger;  but you are probably doing that in steps.

src/main/java/ilex/util/JavaLoggerAdapter.java
```diff
@@ -106,60 +114,69 @@ public static void initialize(ilex.util.Logger ilexLogger, boolean _doForward, S
...
-				logger.addHandler(instance());
-				logger.setLevel(Level.ALL);
 			}
+
+			for(String path : HIDE_BY_DEFAULT)
+			{
+				setSevere(path, globalLogger);				
+			}
+
+			boolean trace = Boolean.parseBoolean(System.getProperty(CwmsConnectionPool.POOL_TRACE_PROPERTY,"false"));
```
ktarbet, line 141: It feels odd to be referencing something with the name CWMS (CwmsConnectionPool) in a utility area of the code.   I would expect Utility code to avoid that type of dependency.

src/main/java/lrgs/lrgsmon/LrgsMonCmdLineArgs.java
```diff
@@ -180,20 +180,13 @@ public void parseArgs(String args[])
 		if (fn != null && fn.length() > 0)
 		{
 			String procname = Logger.instance().getProcName();
```
ktarbet, line 182: This pattern of setting up a FileLogger is also in LqmCmdLineArgs, and DrgsCmdLineArgs,  ...    perhaps a base or utility class could take this on.

src/main/java/org/opendcs/logging/IlexToSlf4jBridge.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    public void doLog(int priority, String text)
+    {
+        Level level = ilexLevelToSLF4J(priority);
+        realLogger.makeLoggingEventBuilder(level)
+                  .setMessage(text)
+                  .log();
+    }
+
+    private Level ilexLevelToSLF4J(int priority)
```
ktarbet, line 32: Consider making ilexLevelToSLF4J a static method.

ktarbet: Consider replacing with a call to JULHelpers.ilexLevelToSLF4J.  However the levels are a little different,  For example ilex.util.Logger.E_Debug3 returns Trace here, but Finest in JULHelpers.

src/main/java/org/opendcs/logging/jul/StaticLogName.java
ktarbet, line 1: Consider  naming StaticLogNamer  (add 'r' to end) 
consider class level Java doc, with something like:

```java
/**
StaticLogNamer is a LogNamer that keeps the same logfile name.  It doesn't rotate log file extensions 
**/

src/main/java/org/opendcs/logging/jul/OpenDcsFilter.java
ktarbet, line 1: The class feels CWMS specific using Property called CwmsConnectionPool.
Consider having the HIDE_*  arrays  in a config file, or and option to show everything if really wanted.

MikeNeilson: One of the next steps is moving the CwmsConnectionPool trace to a more generic location as the plan is for all of the implementations to use a java.sql.DataSource going forward.

src/main/java/org/opendcs/logging/IlexToSlf4jBridge.java
MikeNeilson: I'll have to review that but there are some slight differences. Though it might just be because of changes that were made as I went.

src/main/java/lrgs/lrgsmon/LrgsMonCmdLineArgs.java
MikeNeilson: Hmm, that's a good idea. The duplication was bugging me but wasn't sure what to do.
I'll see about a utility class.

src/main/java/ilex/util/JavaLoggerAdapter.java
MikeNeilson: Yes, the property started within the CwmsConnectionPool and for this purpose I just needed a way to get it and see the results. Long term, as said, in a comment below the "trace connections" property will be moved somewhere generic.

src/main/java/ilex/util/SequenceFileLogger.java
MikeNeilson: Yes, it's still technically used by at least one specific system, but we should be able to remove it eventually.

MikeNeilson:
I'm going to focus on #394 ; I think it's a cleaner path towards the goal.
And it would be good to separate the file logic from the actually logging.

PR #275, opened 2023-08-04
Fixed U1 timestamp issue

baoyuyin:
## Problem Description

1Minute U1 data was decoded with 1 day shift for the 00:00 to 00:05Z time frame. 

## Solution

The dataTime in the java code was shifted back 6 minutes in order to calculate the redundantDataTime, but it failed to change it back to its original time. The fix is to add the 6 minutes back.

## how you tested the change

Decode U1 data with the 00:00 to 00:05Z time frame to confirm

## Where the following done:

   - [ x] Unit tests created or modified that run during ant test.
   - [ x] Integration tests created or modified that run during integration testing

### DIFF
--- a/src/main/java/decodes/decoder/Nos6Min.java
+++ b/src/main/java/decodes/decoder/Nos6Min.java
@@ -3,6 +3,9 @@
 *
 * $Log$
 *
+* 2023/08/04 Baoyu Yin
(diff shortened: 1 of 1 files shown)


baoyuyin:
Checked in a new version of Nos6Min.java to fix a U1 timestamp issue.

agilmore2 (approved):

Issue #283, opened 2023-08-17
Location selector sorting messes up result


MikeNeilson:
**Describe the bug**
If one sorts by any column in the location editor a different location is returned

**To Reproduce**
1. Open CompEdit
2. Open Computation
3. select and edit a parameter
4. Sort by base
5. See the base location change in "Add Mask By" but not match the rendered row

**Expected behavior**

Correct element should be presented to downstream uses

**Screenshots**
![image](https://github.com/opendcs/opendcs/assets/7980296/6c715304-75fe-4a69-90b0-ddb26676681e)

**Desktop/ServerSide (please complete the following information):**
 - OS: Windows 10
 - Version: 7.0.8

**Additional context**
This looks to be identical to the items fixed in https://github.com/opendcs/opendcs/pull/265/files, considering these were working I suspect some Java update "corrected" behavior that we either relied on or worked around.

Additionally it appears that JTable can handle sorting on it's own now, we just need to make sure the table model is probably defined and so far as I can tell it is other than getColumnClass not be configured to return a Class appropriate to the actual value.

MikeNeilson:
This also appears to be a good candidate for getting the "Database" to a constructor instead of using Database.getDb() and actually setting up tests without too much difficulty.

While the SiteSelectPanel and Dialog are used in multiple places they would all already have a reference to the Database.

hoffmanjs:
Fixed all (Location, Param, Version, Units, and Time Series Lookup)  of the "Select" button options.  
![image](https://github.com/opendcs/opendcs/assets/122313426/a3259860-f491-41dc-b8d2-f906ddc77089)

When each button is selected it brings up a different dialog box.  All of the dialog boxes had similar sorting issues.  All dialog sorting issues mentioned above were resolved.

MikeNeilson:
@agilmore2 FYI, this likely affects HDB as well, when I was looking through the code a different dialog for location selection is opened. However, it was the dbedit SiteList panel which I believe Scott also fixed so while it needs a quick check it's probably working in the current versions.

agilmore2:
It was messed up when I tried it a week ago from current build, thanks!

DanielTOsborne:
I noticed myself that in the computation list, if you double click a comp, it opens correctly, but if you click "Copy" to duplicate it, the wrong one is copied.
This was with 7.0.11.

msweier:
[...]

PR #285, opened 2023-08-22
Dialog selector sorting messes up result.

hoffmanjs:
## Problem Description

Fixes #.  Location selector sorting messes up result #283

## Solution
Converted the selected row in the table to the selected row in the table model.  This will selected the correct result even if the table is sorted.

## how you tested the change
Test as shown and described in the PR 

## Where the following done:


MikeNeilson (approved):
This looks good to me. I'm sure there are more but we gotta start somewhere.

Issue #297, opened 2023-09-01
OpenDCS 7.0.8 Gui Fix does not show all decoding scripts


JasonDSS:
**Describe the bug**
OpenDCS 7.0.8 Gui Fix does not show all decoding scripts when an older version 6.8.14 does.

**To Reproduce**
1. Copy edit-db folder from 6.8.14 and replace the one in 7.0.8
2. Launch the GUI interface for 7.0.8 and click on DECODES Database Editor
3. Go to configs tab and click on any config with multiple decoding scripts
4. 7.0.8 does not show all decoding scripts, but they're in the xml file. Attached example where the older version sees ST and RD decoding scripts, but 7.0.8 sees only RD

**Expected behavior**
Should show all decoding scripts

**Screenshots**
OpenDCS 7.0.8 not showing the ST Decoding script:
![image](https://github.com/opendcs/opendcs/assets/143829980/cf485345-55fd-4dd9-929f-6a3ef04c92b2)

ST Decoding script picked up by an older OpenDCS, version 6.8.14:
![image](https://github.com/opendcs/opendcs/assets/143829980/9f851e72-54c4-46c0-afd9-4e7e65559ef5)

**Desktop/ServerSide (please complete the following information):**
 - OS: Windows
 - Version 10
[DA-Omnisat-019.txt](https://github.com/opendcs/opendcs/files/12500310/DA-Omnisat-019.txt)

**Additional context**

JasonDSS:
Tried several versions, and it appears version 7.0.5 is the last one that doesn't have this issue.

MikeNeilson:
That tracks. I think i roghly kniw what the issue it.

Fyi it may have been my phone download, but the export is cutoff at the end
and doesnt have the other script.

MikeNeilson:
oh quick question just to make sure I don't run down the wrong path. what this with getting reading from oracle with 7.0.8 or an XML based database?

JasonDSS:
This is 7.0.8 trying to read an XML database

hoffmanjs:
I compared version 6.8.14 and 7.0.8.  Both versions worked exactly the same.  Both versions reading from an oracle database.

MikeNeilson:
That make sense. The way the XML data is read is different and I probably missed an edge case when I was fixing it from the first time I made a tweak.

ktarbet:
I've created a test that I believe reproduces the issue.  However, I need advice how to use a simple XML test database.  
[...]

PR #305, opened 2023-09-14
add XmlDatabaseTest.java with failing test for issue #297

ktarbet:
Draft PR - Documents  #297  with junit test.

I was able to re-work the test to create a database from scratch, and reproduce the issue.  No input files required.

### DIFF
--- a/src/main/java/decodes/xml/PlatformConfigParser.java
+++ b/src/main/java/decodes/xml/PlatformConfigParser.java
@@ -5,7 +5,8 @@ package decodes.xml;
 
 import org.xml.sax.Attributes;
 import org.xml.sax.SAXException;
-import java.util.Collection;
+
+import java.util.ArrayList;
 import java.util.Iterator;
 import decodes.db.*;
 import decodes.db.DecodesScript.DecodesScriptBuilder;
@@ -13,6 +14,8 @@ import ilex.util.TextUtil;
 import ilex.util.AsciiUtil;
 import ilex.util.Logger;
 import java.io.IOException;
+import java.util.List;
+
 import ilex.xml.*;
 
 /**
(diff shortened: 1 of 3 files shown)


MikeNeilson:
The structure of the test looks great to me. Do you have time to try and fix the actual issue or do you need to pass the buck at this point?

ktarbet:
I'll spend some time on the actual issue.

ktarbet:
The potential solution in https://github.com/opendcs/opendcs/pull/305/commits/a38eb28e690dd0a1baa0aa50aaefd55743e07c7d 
uses lists of  DecodesScriptParser and DecodesScriptBuilder, in the PlatformConfigParser.  When the endTag </PlatformConfig> is reached everything is ready to parse in the endElement() method

src/main/java/decodes/xml/PlatformConfigParser.java
```diff
@@ -24,8 +27,8 @@ public class PlatformConfigParser implements XmlObjectParser, XmlObjectWriter, T
 	/**
 	 * Reference to the DecodesScriptBuilder so we can finalized after we've read it in.
 	 */
-	private DecodesScriptBuilder scriptBuilder; 
-	private DecodesScriptParser scriptParser;
+	private List<DecodesScriptParser> decodesScriptParserList = new ArrayList<>();
+	private List<DecodesScriptBuilder> decodesScriptBuilderList = new ArrayList<>();
```
MikeNeilson, line 31: This looks like a reasonable solution to me. Reading through it I suspect we could limit this to just an array of `DecodeScriptBuilder`s. I think my original work in this design it seemed impractical to do so.

But that can be addressed later.

MikeNeilson (approved):
This looks good to me.

However, I will leave it a day or two for others to review.

PR #316, opened 2023-09-26
Bugfix/screening connection retrieval

MikeNeilson:
## Problem Description

ScreeningDAO wasn't closing connections after calls to getConnection as no references were held.

## Solution

Move getConnection to an assignment within a try-with-resources block.
Changed DAO/DAI returns to List instead of ArrayList to allow use of the new SQL query wrappers.

## how you tested the change

Manually, leaving Screening open and performing work.

## Where the following done:

### DIFF
--- a/src/main/java/decodes/cwms/validation/dao/ScreeningDAO.java
+++ b/src/main/java/decodes/cwms/validation/dao/ScreeningDAO.java
@@ -345,9 +345,9 @@ public class ScreeningDAO
             doQuery(q, rs -> rsAddCriteria(rs, ret),screeningCode);
 
             q = "select distinct " + durMagColumns + " from " + durMagTable
-                + " where screening_code = ?"// + screeningCode
+                + " where screening_code = ?"
                 + " and unit_system = 'SI'"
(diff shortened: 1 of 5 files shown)


src/main/java/decodes/cwms/validation/dao/ScreeningDAO.java
ktarbet, line 1: I heard talk about getting into a 'broken' state and not failing.  This may not be related, but here is an observation:

ScreeningDAO extends DaoBase and implements ScreeningDAI .


the close() method in DaoBase seems to satisfy the AutoCloseable of ScreeningDAI .  However, that close() swallows some query exceptions so that could be a point of non-failure.

ktarbet (approved):

src/main/java/decodes/cwms/validation/dao/ScreeningDAO.java
MikeNeilson: Eating errors on close it's fairly normal in the context of how that's implemented.

This isn't used in the routingscheduler and the broken state there appears to now be about the data source (like an LRGS datasource) connections than the database connection.

PR #327, opened 2023-10-03
Devops/integration test organization

MikeNeilson:
## Problem Description

I've been struggling with how to handle certain integration level tests for each of the different implementations. For example the XML database can't do any of the time series things.

With this new framework in place it *should* be easier to standup the new lower level regressions tests (those not from the very high level direct application tests such as the existing DecodesTests) and reduce some boiler plate to get key instances of classes for specific testing.

## Solution

- Extracted the existing AppTestBase class to a Extension that implements a few different components to handle test database and configuration creation.
- Added logging to the integration tests to better see issues.
- Added a trivial test to verify resources were attempting to be recreated.

## how you tested the change

Running integration-tests. Only changes were to the build.xml and sources under `src/test-integration`

## Where the following done:

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -131,6 +131,7 @@ NOTE at this time jdk8 must be used as there is still a dependency on tools.jar
                 <!--<pathelement location="${build.classes}"/>-->
                 <path refid="runtime.classpath"/>
                 <path refid="test.classpath"/>
+                <path refid="junit.platform.libs.classpath"/>
             </classpath>
         </javac>
         <sync todir="${build.test-integration.resources}">
@@ -722,6 +723,10 @@ NOTE at this time jdk8 must be used as there is still a dependency on tools.jar
                     }
 
                     project.getReference("test.classpath").each {
+                        classpathentry(kind:"lib", path:it)
+                    }
+
+                    project.getReference("junit.platform.libs.classpath").each {
                         classpathentry(kind:"lib", path:it)
                     }
                 }
--- a/build.xml
+++ b/build.xml
@@ -225,6 +225,7 @@ NOTE at this time jdk8 must be used as there is still a dependency on tools.jar
                     <jvmarg value="-Dresource.dir=${build.test-integration.resources}"/>
                     <jvmarg value="-DDCSTOOL_HOME=${stage.dir}"/>
                     <jvmarg value="-Dopendcs.test.engine=${opendcs.test.engine}"/>
+                    <jvmarg value="-Djava.util.logging.config.file=${build.dir}/../src/test-integration/test-config/logging.properties"/>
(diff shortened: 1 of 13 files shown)


docs/source/dev-docs.rst
```diff
@@ -127,6 +127,81 @@ to run each do the following:
...
+
+OpenDCS now contains a framework for running integration tests. See the folder src/test-integration for the code.
+
+Framework
+---------
+
+There is set of code under :code:`org.opendcs.fixtures` that allows configuration and setup to take place, determine if a given 
+test should be enabled or not and other per test tasks.
+
+All new tests classes should derive from :code:`org.opendcs.fixtures.ApptestBase`. This class is marked with the :code:`OpenDCSTestConfigExtension`
```
ktarbet, line 141: All new *integration* tests...

docs/source/dev-docs.rst
```diff
@@ -127,6 +127,95 @@ to run each do the following:
...
+an instance of any field annotated with :code:`@ConfiguredField` as seen in the table 
+above for the configuration.
+
+The only other injected field is a :code:`TimeSeriesDb` which is Provided by the Configuration
+and will already be valid and can be used directly for things like testing DaoObjects or null which 
+indicates the implementation under test doesn't use the any of the timeseries database components.
+
+The sample :code:`LoadingAppDaoTestIT` uses the :code:`@EnableIfSql` :code:`ExecutionCondition`
+ to determine if the test should be run or not.
+
```
ktarbet, line 203: The sample :code:`LoadingAppDaoTestIT` uses the :code:`@EnableIfSql`  **_annotation that extends from Junit's_**  :code:`ExecutionCondition`
 to determine if the test should be run or not.

docs/source/dev-docs.rst
```diff
@@ -127,6 +127,95 @@ to run each do the following:
...
+Additional ExecutionConditions and parameter injection will be added in the future as needed and as
+we better identify concepts to map to vs implementation details.
+
+
+Caveats
+-------
+
+OpenDCS supports several implementations, the XML database, the baseline Postgres and Oracle database,
```
ktarbet, line 211: **_and two Oracle databases:_**
USBR's HDB and USACE's CWMS.

docs/source/dev-docs.rst
```diff
@@ -127,6 +127,95 @@ to run each do the following:
...
+USBR's HDB and USACE's CWMS. While each share the same fundamental concepts portions of the implementation, like Site names
```
ktarbet, line 212: < line feed >


Each implementation shares the same fundamental concepts.  However, portions of the

ktarbet (approved):
I like this framework for testing!

More example integration tests will lower the learning curve of adding a new integration test.  I suspect the first test a developer, such as me, makes will take some effort, but immediately pay off in setup code saved.

Issue #318, opened 2023-09-29
Computation list from cp_comp_depends table


MikeNeilson:
**Component to be refactored**
decodes/tsdb/TimeSeriesDb.java
opendcs/dao/COmpDependsDAO.java

**Purpose of refactoring**
the `select computation_id from cp_comp_depends...` query used in both has been identified as one of the most called queries in the database, the lack of bind variables is causing plans to be pushed out of shared memory affecting total performance.

**Proposed method**
Migrate to bind variables and appropriate new handlers for SQL.

NOTE: CompDependsDAO uses an "in (item1,item2,...itemN) " list. No known JDBC implementations appropriately handle this. The use of temporary tables so the query to be change to ... `in (select column from the_temp_table)...` however the use across Postgres and Oracle will need to be investigated and adapted to.

**Additional context**
- https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-COMPATIBILITY
- oracle follows the SQL standard of the table always existing; just fresh for each session.

PR #347, opened 2023-10-24
Refactor/loading app dao

### LINKED ISSUE #276: LoadingAppData usage of getConnection causes connections to remain open and not be returned to the pool.
**Describe the bug**
LoadingAppDAO is leaking connections from the tool

**To Reproduce**
1. Using CWMS
2. Run compproc
3. eventually run out of connections

**Expected behavior**
Connections are always returned when database work is finished.

Additionally DAO should not hold a reference to prepared statements, just always call Connection::prepareStatement and let the JDBC driver sort it out.

**Screenshots**

 - OS: Solaris, likely all.
 - Version 7.0.x

**Additional context**

MikeNeilson:
## Problem Description

Fixes #276. 

## Solution

Establish use of bind variable and new SQL handlers.
I looked into handling setting properties as a larger join; I think that should wait until JDBI usage is established.
Given the current design (both existing and the new handlers) it's not clear to me how to go about it cleanly.

The `select ... in (...)` clause is also a bit tricky. I think temp tables are the best solution, however the syntax/setup between Oracle and Postgres needs to be reviewed (see #318 for current notes)

This is the start of infrastructure for Computation/Algorithm/etc tests. My plan is to 
reduce the amount of boiler plate required for any given computation test itself likely be adding mechanisms as
part of the extension so a test can assert "I need X running" but in separate pull requests.
This at least gets us A test of compproc so we know at least a few things are still working. (Actually caught a couple of mistyped sql statements with the new integration test.

## how you tested the change

Running loading apps manually to verify operation.
Created infrastructure to run CompDepends and instance of CompProc to verify all operations.
Created simple tests of LoadingApp dao for the basics

## Where the following done:


MikeNeilson:
Okay, so after working on the annotations for the decodes data I decided I would just create the system for any TsDbApp as well. Mostly just initial work for CompProc itself.

MikeNeilson:
I'll leave future improvement to the next computation test that gets added.

src/main/java/opendcs/dao/LoadingAppDao.java
```diff
@@ -81,713 +76,680 @@
...
+            doModify(q, app.getKey());
+            if (DecodesSettings.instance().editDatabaseTypeCode == DecodesSettings.DB_OPENTSDB
+             && db.getDecodesDatabaseVersion() >= DecodesDatabaseVersion.DECODES_DB_15)
+            {
+                q = "delete from "
+                    + (db.getDecodesDatabaseVersion() >= DecodesDatabaseVersion.DECODES_DB_17 ? "ALARM_EVENT"
+                        : "ALARM_DEF")
+                    + " where LOADING_APPLICATION_ID = ?";
+                doModify(q, app.getKey());
+                q = "delete from PROCESS_MONITOR where LOADING_APPLICATION_ID = ?";
```
agilmore2, line 483: Process Monitor is only installed in OpenTSDB?

build.xml
```diff
@@ -284,6 +284,11 @@ more information about each.
         <mkdir dir="${build.dir}/test-integration/${opendcs.test.engine}/results"/>
         <mkdir dir="${build.dir}/test-integration/${opendcs.test.engine}/tmp"/>
         <jacoco:agent property="coverage.agent.param.it" destfile="${coverage.dir}/jacoco.test-it-${opendcs.test.engine}.exec"/>
+        <path id="opendcs.runtime">
+            <pathelement location="${stage.dir}/bin/opendcs.jar"/>
+            <fileset dir="${stage.dir}/dep"/>
+        </path>
+        <pathconvert property="opendcs.test.classpath" refid="opendcs.runtime"/>
```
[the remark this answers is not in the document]
MikeNeilson: That was to support testing. When moved the applications to their own process instanced I discovered the path wasn't quite fully set right for them to run.... Probably good to put a comment about that though.

src/test-integration/java/org/opendcs/fixtures/annotations/TsdbAppsRequired.java
ktarbet, line 1: Since TsdbAppsRequired.java and TsdbAppRequired.java are so close in names, javadoc on both would be helpful.

src/test-integration/java/org/opendcs/fixtures/configurations/opendcs/pg/OpenDCSPGConfiguration.java
```diff
@@ -74,30 +79,45 @@ public Map<String,String> getEnvironment()
...
+        Jdbi jdbi = Jdbi.create(db.getJdbcUrl(),db.getUsername(),db.getPassword());
         jdbi.useHandle(h->{
             log.info("Creating application user.");
-            h.execute("create user dcs_proc with password 'dcs_proc'");
+            h.execute("DO $do$ begin create user dcs_proc with password 'dcs_proc'; exception when duplicate_object then raise notice 'user exists'; end; $do$");
             h.execute("GRANT \"OTSDB_ADMIN\" TO dcs_proc");
             h.execute("GRANT \"OTSDB_MGR\" TO dcs_proc");
             log.info("Setting authentication environment vars.");
+            environmentVars.put("DB_USERNAME","dcs_proc");
+            environmentVars.put("DB_PASSWORD","dcs_proc");
```
ktarbet, line 119: I would normally expect reading password from environment instead of putting in environment, but I think this is for testing in a container only?

ktarbet (commented):
I've created a PR with code review suggestions:

https://github.com/opendcs/opendcs/pull/411

MikeNeilson:
Okay, I think this one is done. There are some additional improvements in #332 that will be easier to review once this is merged in.

src/test-integration/java/org/opendcs/fixtures/OpenDCSTestConfigExtension.java
```diff
@@ -127,36 +192,72 @@ else if (f.getType().equals(TimeSeriesDb.class) && configuration.isRunning())
...
+            for(String file: decodesConfig.value())
             {
-                Programs.DbImport(setupLog(), configuration.getPropertiesFile(), env, exit, decodesConfig.value());
+                files.add(AppTestBase.getResource(file));
             }
         }
+
+        if (!files.isEmpty())
+        {
+           Programs.DbImport(setupLog("decodes-setup.log"), configuration.getPropertiesFile(), env, exit, files.toArray(new String[0]));
```
ktarbet, line 217: This may limit different tests from having different logs?

src/test-integration/java/org/opendcs/fixtures/OpenDCSTestConfigExtension.java
```diff
@@ -127,36 +192,72 @@ else if (f.getType().equals(TimeSeriesDb.class) && configuration.isRunning())
...
+        {
+            for(String file: compConfig.value())
+            {
+                files.add(AppTestBase.getResource(file));
+            }
+        }
+
+        if (!files.isEmpty())
+        {
+            Programs.CompImport(setupLog("computation-setup.log"), configuration.getPropertiesFile(), env, exit, files.toArray(new String[0]));
```
ktarbet, line 252: will this log file be overwritten by multiple tests?

ktarbet (approved):

PR #366, opened 2023-10-30
Add annotation for setting required setup data.

MikeNeilson:
## Problem Description

While working on the LoadingAppDao PR I thought setting up annotation instead of using before handlers would make more sense, it would certainly alleviate the additional "if sql/tsdb" checks within those blocks.

## Solution

Created a new annotation for the developer to specify decodes database setup files required for these tests
Additionally, I setup code coverage for helping to track test progress.

I'd like to add a status badge, as I did for the actual build success status, but still reviewing the best way.

## how you tested the change

Manually ran OpenDCS-XML and OpenDCS-Postgres integration tests

## Where the following done:

### DIFF
--- a/docs/source/dev-docs.rst
+++ b/docs/source/dev-docs.rst
@@ -236,6 +236,7 @@ to run each do the following:
 Only CPD is fast. checkstyle and SpotBugs are rather slow.
 
 .. _integration_test_infra:
+
 Integration Test infrastructure
 ===============================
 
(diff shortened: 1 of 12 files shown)


docs/source/dev-docs.rst
```diff
@@ -301,6 +302,20 @@ Integration tests inherit from :code:AppTestBase. This simplifies access to reso
...
 +--------------------------------------------+--------------------------------+
  
+
+At the Class and method level the following annotations are available.
+
++--------------------------------------------+--------------------------------+
+|Annotation                                  |Description                     |
++============================================+================================+
+|DecodesConfigurationRequired                |List of database import files   |
+|                                            |needs for tests to succeed.     |
```
ktarbet, line 312:
```suggestion
|                                            |needed for tests to succeed.     |
```

ktarbet (approved):

PR #387, opened 2023-11-11
Feature/export test extension

### LINKED ISSUE #386: export test extensions to supprot downstream development
The test extension created for the integration tests was designed so that downstream systems,  such as https://github.com/opendcs/rest_api and https://github.com/opendcs/dcpmon can use it easily for their own test automation.

Create a jar for the extension, pretty much everything under "fixtures" for use downstream.

duplicating all of that manually... I didn't really consider it, but it's worth mentioning as a terrible option.

**Additional context**
I've been poking the rest_api, Need to make some changes and it's already clear that having tests in place is quite simply going to be required.

MikeNeilson:
## Problem Description

Closes #386 . 

## Solution

Create additional output artifacts so that projects like DCPMon, rest_api, and any other users can

## how you tested the change

TODO: need to install locally and manually test whether the concept works.

## Where the following done:


MikeNeilson:
Need to review the dev docs and javadocs for each new element but I think the final structure is there.

I'm also considering moving that initialization out of the "configuration" classes since it's actually identical for that initial database.

MikeNeilson:
Given the flyway change merged in I don't think this is the way forward for downstream apps, like DcpMon and the REST API.

Issue #384, opened 2023-11-08
specific SHEF .E records not importing properly


ktarbet:
**Describe the bug**
The following SHEF data is not imported correctly.
```
.ER WOOR1 20231023 E DC202310231053/DH14/HGIFF/DIH6
:STAGE FCST  /   8AM /   2PM /   8PM /   2AM 
.E1  :1023:          /   3.8 /   3.7 /   3.5 
.E2  :1024:  /   3.4 /   3.3 /   3.2 /   3.2 
.E3  :1025:  /   3.1 /   3.1 /   3.1 /   3.0 
.E4  :1026:  /   3.0 

```

**To Reproduce**
[export.zip](https://github.com/opendcs/opendcs/files/13300002/export.zip)

1. Configure a Decodes Script to import using  shefprocess() method (see attached export.zip)
2. Decode the sample message above using Decode Script Editor.
3. See:  'missing' record values
<html>
<body>

2023/10/20-00:00:00 | missing
-- | --
2023/10/20-06:00:00 | 3
2023/10/20-12:00:00 | 3.1
2023/10/20-18:00:00 | 3
2023/10/21-00:00:00 | 3.1
2023/10/21-06:00:00 | 3.1
2023/10/21-12:00:00 | 3.2
2023/10/21-18:00:00 | missing
2023/10/22-00:00:00 | 3.3
2023/10/22-06:00:00 | 3.2
2023/10/22-12:00:00 | missing
2023/10/22-18:00:00 | 3.4
2023/10/23-00:00:00 | 3.7
2023/10/23-06:00:00 | 3.5
2023/10/23-18:00:00 | 3.8
</body>
</html>

**Expected behavior**
should parse same as NWS shefit program as shown below:
```
C:\incoming\shef>shefit shef2.txt
STATION   OBSERVATION-DATE/TM  CREATION-DATE/TIME   PARM-CD     DATA-VALUE Q  PROB-CD DURTN R T  GROUP-NM  QUOTE
aaaaaaaa  yyyy mm dd hh nn ss  yyyy mm dd hh nn ss  pedtsep     rrrrr.dddd a  rrr.ddd iiiii i i  aaaaaaaa  "aa..."

WOOR1     2023-10-23 18:00:00  2023-10-23 14:53:00  HGIFFZ          3.8000 Z   -1.000  0000 1 1            " "
[...]

PR #388, opened 2023-11-13
Test/shef parser

ktarbet:
Exploring Fix for #384

## Solution

Attempting to create a shef parsing test using the test/java/fixtures/DecodesHelper. 

I've made some modifications to DecodesHelper, and made some progress.  now the ShefProcess is confused because the SensorNumber = -1, so data is being dropped.

Considering if I should use integration test instead...


ktarbet:
@MikeNeilson Do you know the design for the .sensors testing input file?  I see a  ```TODO: lookup algo and parts``` in DecodesHelper  , and I'm not sure what the purpose of the ';' semi-colon is.

I would like to include DataType information in the .sensors file.  I've hacked an experiment in commit [5c4d2e1](https://github.com/opendcs/opendcs/pull/388/commits/5c4d2e10a9e265210608339fcd95627d9c399f60)

ktarbet:
The SHEF decoding is nearly working in the test.  However, the last value (of the SHEF time-series) is missing when using the Junit/DecodesHelper.  When using the UI the last value is OK.

MikeNeilson:
It was probably to be a separator for the algorithms, but for the life of me I don't remember why i put it there.
And I like what you setup here so let's stick with your change and we'll figure out the conversion algorithms later.

MikeNeilson:
The last value being missing is odd, the test is calling `decodesScript.decodeMessage` just as the GUI would. maybe add an extra newline to the input file?

ktarbet:
Thanks- extra newline fixed the issue.

MikeNeilson:
Givwen that fixed it that is in itself a bug. I dont think a blank newline
shpuld be required.

src/main/java/decodes/decoder/ShefProcess.java
```diff
@@ -218,31 +218,30 @@ else if (c == '\n')
 				if (c == ':')
 				{
 					lineContext = LineContext.InComment;
-					debug(idx, "start of commnet");
-					continue;
+					debug(idx, "start of comment");
 				}
 				else if (Character.isWhitespace(c) || c == '/')
```
MikeNeilson, line 223: This all looks reasonable but would probably be good to also include a simple test with the .A format to verify we aren't going to be adding a regression there.

ktarbet: The header parsing for shef  (ShefPMParser.java) doesn't check for comment lines!

If I remove the comment lines from the sample .A data it appears to be happy.

src/main/java/decodes/datasource/ShefPMParser.java
```diff
@@ -196,6 +147,56 @@ else if (dateField.length() == 8)
 		msg.setPM(GoesPMParser.SITE_NAME, new Variable(stationName));
```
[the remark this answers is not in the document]
ktarbet: Depending on the data set we have options.
We could subtract leading comment lines from the MESSAGE_LENGTH . 

For a simple .E format the header is a little more concrete.

Preferably, we have definition for header length setHeaderLength() , that is compatible with MESSAGE_LENGTH .

src/main/java/decodes/datasource/ShefPMParser.java
```diff
@@ -191,11 +133,85 @@ else if (dateField.length() == 8)
...
 		Logger.instance().debug3("SHEF Message Date=" + ccyymmdd.format(msgTime));
 		msg.setPM(PM_TIMEZONE, new Variable(tz.getID()));
-		msg.setPM(GoesPMParser.MESSAGE_LENGTH, new Variable(data.length - headerEnd));
+		msg.setPM(GoesPMParser.MESSAGE_LENGTH, new Variable(data.length));
 		msg.setPM(GoesPMParser.FAILURE_CODE, new Variable('G'));
 		msg.setPM(GoesPMParser.SITE_NAME, new Variable(stationName));
 	}
 
+
+	private Date parseDate(String dateField, TimeZone tz) throws Exception {
```
MikeNeilson, line 142: Sorry, I didn't catch this before. the opening `{` needs to be on the next line.

src/test/java/fixtures/DecodesHelper.java
```diff
@@ -112,6 +115,15 @@ public static Arguments getScript(String testName, String resourcePath) throws D
                 decodesScript.scriptSensors.add(stage);
                 ConfigSensor configSensor = new ConfigSensor(decodesScript.platformConfig, Integer.parseInt(parts[0]));
                 configSensor.sensorName = parts[1];
+
+                if( parts.length >= 5) {
```
MikeNeilson, line 119: `{` on next line.

src/test/java/fixtures/DecodesHelper.java
```diff
@@ -292,17 +312,41 @@ public static DecodedSample sampleFor(int sensor, ZonedDateTime sampleTime,
...
         }
-
+        String info = samplesToString(sensor, samples);
         throw new NoSuchElementException("Expected Sample for sensor " 
                                      + sensor + " at time "
-                                     + sampleTime + " doesn't exist");
+                                     + sampleTime + " doesn't exist\nDecoded "+samples.size()+" samples\n"+info);
+    }
+
+    private static String samplesToString(int sensor, ArrayList<DecodedSample> samples) {
```
MikeNeilson, line 328: `{` on next line

MikeNeilson (changes requested):
Minor formatting changes required. Otherwise the logic of this looks great to me.

MikeNeilson (approved):

PR #394, opened 2023-11-22
Refactor/logging attempt 3

MikeNeilson:
## Problem Description

Starts #161 . 

## Solution

Instead of immediately yanking out the current custom logging this implements a SLF4J handler so we can
modernize the logging over time.

Given the way this and SLF4j work we can later determine the best time to shim in the logger itself. #266 Used an SLF4JBridge, 
when the time comes to change the logging backend, we can implement the bridge to avoid having to change *every* file at once.

The current implementation is slightly problematic for downstream library usage (only one `SLF4ServiceProvider` can exist on the classpath.

The simple solution is to package that code up as it's own jar so it's included in our distribution but the artifact on maven central can be pulled in without it. I will do that after initial review as it would've just added noise to the particular task at hand. This may also simplify the next challenge which is dealing with things like Trace in the decode script panel and computation execution as we can just decide our GUI implementation requires this custom implementation.

## how you tested the change

Manually to verify messages showed up as expected.

## Where the following done:

### DIFF
--- a/src/main/java/ilex/util/ThreadLogger.java
+++ b/src/main/java/ilex/util/ThreadLogger.java
@@ -136,7 +136,7 @@ public class ThreadLogger extends Logger
 	 * its own priority filtering.
 	 */
 	@Override
-	public synchronized void log( int priority, String text )
+	public void log( int priority, String text )
 	{
 		Logger tl = getCurrentThreadLogger();
 		if (tl != this)
@@ -198,4 +198,3 @@ public class ThreadLogger extends Logger
 		this.moduleName = moduleName;
 	}
 }
-
--- a/src/main/java/org/opendcs/utils/sql/SqlSettings.java
+++ b/src/main/java/org/opendcs/utils/sql/SqlSettings.java
@@ -1,0 +1,22 @@
+package org.opendcs.utils.sql;
+
+/**
+ * This classes is to old developer or advanced configuration settings
+ * Set as properties as opposed to intended user facing settings.
+ */
+public class SqlSettings
+{
+    /**
+     * Determines whether or not to maintain and print track traces of 
+     * various java.sql.Connection operations. Primarily @see opendcs.util.sql.WrappedConnection
+     * to find instances of createStatement and print stacktraces showing the location.
+     *
+     * @since 7.0.11 (cwms.connection.pool.trace to be removed)
+     */
+    public static final boolean TRACE_CONNECTIONS = 
+        Boolean.parseBoolean(
+            System.getProperty("opendcs.connection.poo..trace",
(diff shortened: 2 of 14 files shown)


src/main/java/ilex/util/ThreadLogger.java
```diff
@@ -136,7 +136,7 @@ public void close( )
 	 * its own priority filtering.
 	 */
 	@Override
-	public synchronized void log( int priority, String text )
```
ktarbet, line 139: does slf4j take care of thread sharing, if synchronized is removed here?

src/main/java/org/opendcs/utils/sql/SqlSettings.java
```diff
@@ -0,0 +1,22 @@
...
+    /**
+     * Determines whether or not to maintain and print track traces of 
+     * various java.sql.Connection operations. Primarily @see opendcs.util.sql.WrappedConnection
+     * to find instances of createStatement and print stacktraces showing the location.
+     *
+     * @since 7.0.11 (cwms.connection.pool.trace to be removed)
+     */
+    public static final boolean TRACE_CONNECTIONS = 
+        Boolean.parseBoolean(
+            System.getProperty("opendcs.connection.poo..trace",
```
ktarbet, line 18: is the   '..'   a typo?

src/main/java/ilex/util/ThreadLogger.java
MikeNeilson: For where it's used yes. otherwise that synchronization should happen exactly at point of writing.
But I'm fairly certain it isn't necessary. It's going to require a bit more testing for sure, but having synchronized here was the cause of a dead lock. due to the user of ilex.util.Logger within the DacqEventDao; it's possible migrating that class to slf4j would allow synchronize to go back, but to be blunt having synchronized on a log statement is asking for performance problems.

This shouldn't get merged in before the next release though so we have some time.

src/main/java/org/opendcs/utils/sql/SqlSettings.java
MikeNeilson: Yep.

ktarbet (approved):

MikeNeilson:
@opendcs/opendcs-core-devs  not that I don't trust Karl but given this change begins the effort that affects everyone at least a quick stop check for sanity would be appreciated.

A known thing to be addressed: The GUIs render log information - there needs to be a way to handle this somehow. given this still uses the ilex logger under the hood I wasn't going to worry about it in this PR.

MikeNeilson:
I don't think I ever realized just how physically long SqlDatabaseIO.java is until I scrolled through to update all those log messages.

PR #399, opened 2023-11-27
Correct all but 2 class usages of getConnection not using connection …

MikeNeilson:
…in try-with.

- DaoBase Can't be fixed, fucntion marked deprecated
- XmitRecordDAO, needs enough work that it should just get the full migration to new handler treatment.

## Problem Description

We're still getting unxpected "connection closed" error in some of the handlers. This will make the closing happen more immediate and be more obvious what's going on.

## Solution

anywhere that had `getConnection().<some method>()` was changed to have and try-with-resources to get the connection first except in DaoBase and XmitRecordDAO (challenges noted in first commit of this PR.)

## how you tested the change

- Integration Test, OpenDCS-Postgres. Will test CWMS today.
- Manual, CWMS, RoutingScheduler, DCPMon, CompProc, CompDepends. I expect DCPMon to eventually fail for the reason of these changes (uses XmitRecordDAO). The others should failure fairly quickly if something was wrong.
- Manual, OpenDCS-Postgres, CompProc, RoutingScheduler - made it correctly through data cycles.

## Where the following done:

### DIFF
--- a/src/main/java/decodes/hdb/HdbTimeSeriesDAO.java
+++ b/src/main/java/decodes/hdb/HdbTimeSeriesDAO.java
@@ -6,6 +6,7 @@ import ilex.var.NoConversionException;
 import ilex.var.TimedVariable;
 
 import java.sql.CallableStatement;
+import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.text.SimpleDateFormat;
@@ -50,12 +51,12 @@ import opendcs.dao.DbObjectCache.CacheIterator;
 public class HdbTimeSeriesDAO extends DaoBase implements TimeSeriesDAI
 {
 	// TODO: Integrate the cache with the methods below.
-	protected static DbObjectCache<TimeSeriesIdentifier> cache = 
+	protected static DbObjectCache<TimeSeriesIdentifier> cache =
 		new DbObjectCache<TimeSeriesIdentifier>(60 * 60 * 1000L, false);
 	private SiteDAI siteDAO = null;
 	protected DataTypeDAI dataTypeDAO = null;
 	private static SimpleDateFormat rwdf = null;
-	
+
 	// MJM 2016/1/8 Calls to reloadTsIdCache only does a refresh for time series
 	// since the last call.
 	// Once per hour only it does a full load.
@@ -63,8 +64,8 @@ public class HdbTimeSeriesDAO extends DaoBase implements TimeSeriesDAI
 	private static final long CACHE_RELOAD_INTERVAL = 3600000L;
 	private static long lastCacheRefresh = 0L;
 	private static final long CACHE_REFRESH_OVERLAP = 120000L;
-	
-	private String tsidQuery = 
+
+	private String tsidQuery =
 		"SELECT a.ts_id, a.site_datatype_id, a.interval, a.table_selector, a.model_id, "
 		+ "b.SITE_ID, b.DATATYPE_ID, d.UNIT_COMMON_NAME, e.site_common_name, c.datatype_common_name "
 		+ "FROM CP_TS_ID a, HDB_SITE_DATATYPE b, HDB_DATATYPE c, HDB_UNIT d, HDB_SITE e ";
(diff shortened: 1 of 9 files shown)


MikeNeilson:
Assuming nothing drastic comes up, I think this is suitable for the final 7.0.10 branch. But I'm also good with just waiting.

src/main/java/opendcs/dao/ComputationDAO.java
```diff
@@ -252,11 +252,11 @@ public DbComputation getComputationById(DbKey compId) throws DbIoException,
...
 
-		try(
-			PreparedStatement getComp = getConnection().prepareStatement(
+		try(Connection c = getConnection();
+			PreparedStatement getComp = c.prepareStatement(
 				"select " + compTableColumns + " from CP_COMPUTATION where COMPUTATION_ID = ?"
 			);
-			PreparedStatement getAppId = getConnection().prepareStatement(
+			PreparedStatement getAppId = c.prepareStatement(
 				"select LOADING_APPLICATION_NAME from HDB_LOADING_APPLICATION where LOADING_APPLICATION_ID = ?"
```
ktarbet, line 260: Consider a future refactor  of the getAppId query into HdbComputation?

src/main/java/opendcs/dao/LoadingAppDao.java
```diff
@@ -531,69 +532,78 @@ public TsdbCompLock obtainCompProcLock(CompAppInfo appInfo, int pid, String host
...
-		try( PreparedStatement getLock =
-				getConnection().prepareStatement(
-					"SELECT * from CP_COMP_PROC_LOCK WHERE LOADING_APPLICATION_ID = ?"
-				);
-		)
-		{
-			getLock.setLong(1,appInfo.getAppId().getValue());
-			try(ResultSet rs = getLock.executeQuery();){
-				if ( rs.next() && (lock = rs2lock(rs)) != null)
+		try (Connection c = getConnection();)
```
ktarbet, line 535: Is the nested try necessary?  can both connection and getLock be created in same try?

ktarbet (approved):
This PR is a good advertisement for try-with-resources pattern.

Consider testing with CWMS and HDB.

src/main/java/decodes/hdb/HdbTimeSeriesDAO.java
```diff
@@ -831,23 +829,24 @@ private void doDelete(CTimeSeries ts)
...
 		String idstr = "SDI=" + ts.getSDI() + ", intv=" + ts.getInterval();
 		String q =  "{ call delete_from_hdb(?,?,?,?,?,?)}";
-		oracle.jdbc.OracleCallableStatement cstmt = null;
+
 		GregorianCalendar gc = new GregorianCalendar(
 			TimeZone.getTimeZone(DecodesSettings.instance().sqlTimeZone));
-		try
+		try (Connection c = getConnection();
+			 oracle.jdbc.OracleCallableStatement cstmt =
+			 	(OracleCallableStatement)c.unwrap(oracle.jdbc.OracleConnection.class).prepareCall(q);
```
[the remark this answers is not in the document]
MikeNeilson: Yeah. @agilmore2 will need to test that. The unwrap might not actually be needed here, but he was getting issues with casts so I thought it would be good to leave one example somewhere.

src/main/java/opendcs/dao/LoadingAppDao.java
MikeNeilson: No, because you can't do the setLong in the try(...) block. Only assignments to an object implementing AutoClosable.

At least that's been my experience. This is actually one of the reasons I created the new handles and tools like JDBI exist... the nesting gets really annoying sometimes.

src/main/java/decodes/tsdb/ImageTest.java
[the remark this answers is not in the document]
MikeNeilson: might actually be able to just delete it. checking the file again I'm not even sure the table referenced even exists.

src/main/java/opendcs/dao/ComputationDAO.java
agilmore2: Think this table is OpenDCS-wide?

src/main/java/decodes/hdb/HdbTimeSeriesDAO.java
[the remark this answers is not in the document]
agilmore2: Can test this tonight. I had to make some of these changes locally just to get 7.0.10 working for HDB at all. I'll be interested to see how your changes match up. ;)

src/main/java/opendcs/dao/ComputationDAO.java
MikeNeilson: yep. the tablename kept hdb in it, we *could* change it to "loading_applcation" but honestly then we'd just have to have logic in place to detect the right name for backwards compatibility and it's still there anyways.

Maybe that can be an 8.0 thing.

Issue #269, opened 2023-08-03
Create Apache Calcite modules for XML SQL Database


MikeNeilson:
Actually create the Apache Calcite infrastructure to operate on the XML database as though it was a SQL database.

Apache Calcite allows for the defining of structure, location, and I/O of any type of data and that handles all of the required SQL query planning.

MikeNeilson:
#520 let us find away around needs to do this.

PR #402, opened 2023-11-28
Feature/xml over sql

MikeNeilson:
## Problem Description

Starts #269  

## Solution

Implement the required elements for Handling the XML database using SQL.
This is preliminary and there is frankly a lot of work todo. However, I think this is a good start to get the idea across and start discussion.

It can read from an EnumList.xml file but then getting the values file.
Interesting challenge there, the values are read into the DbEnum objects. so technically in this system they're already loaded.

Still TODO: (though maybe not in this PR)

- Determine a better way than hard coding everything; worked fine for the first one, not the next 100 "tables"
   - Add annotations to types?
   - Use the model.json schema to manually define everything
   - pure java.lang.reflect.
   - Some combination 
   - Could this be a later on just loading an XML database. E.G. passing things to down to an instance of XmlDatabase vs manually loading. (I suspect this will be the easiest path forward actually.)
- Ability to write data.
- Is this even a good idea?
   - I'm pretty sure if we don't do this we won't be able to consolidate the connections and properly implement pool behavior.

Thus far I've only spent about 5 hours creating this, partly by just running the tests, seeing the errors and comparing to the example CsvProject: https://github.com/apache/calcite/tree/main/file/src/main/java/org/apache/calcite/adapter/file

An additional possiblity this opens is using FasterXML, JAXB, or another more direct parser instead of the current SAX parser.

## how you tested the change

Created integration test provider. Tests will need to be expanded to cover all edge cases.
While the lower level tests would be faster, the integration test system is a good place to keep these tests as DbImport is called which has to read and write *everything* so we should get really good test coverage.

## Where the following done:


MikeNeilson:
My current goal with this PR, which we can then decide if it should be merged in and we really start the process, is to have read and write of enums working.

But for now I just wanted to expose people to the code of the concept.

MikeNeilson:
So this does now "successfully" read the enums and the queries don't fail. However the XML data doesn't have any index columns (except maybe platforms) so need to figure out a way to deal with that.

src/main/java/decodes/db/HasIterator.java
```diff
@@ -0,0 +1,8 @@
+package decodes.db;
```
MikeNeilson, line 1: This isn't the right package location. Also I think I need a few more interfaces to help simplify the Calcite code a little.

Once more discussion has happened we'll figure out a better location.

src/main/java/opendcs/dao/EnumSqlDao.java
```diff
@@ -87,7 +87,7 @@ public DbEnum getEnum(String enumName)
 				return ret;
 			
 			int dbVer = db.getDecodesDatabaseVersion();
-			String q = "SELECT " + getEnumColumns(dbVer) + " FROM Enum";
```
MikeNeilson, line 90: oh, this wasn't actually needed, just had to figure out how to set the Calcite shema to not be case sensitive.

MikeNeilson:
Well, there's certainly remove for improvement (especially some simplification) but this gets the idea across.

It appears a class per table *is* actually the simplest solution as it allows for just that basic need of implementing an add and the specific columns that would be provided. But it's not nearly as much code as I expected so there's that.

MikeNeilson:
Slight snafu, not all operations are as directly supported as I originally thought. once i got it to actually call UPDATE i discovered some issues. 

I'll continue slowly experimenting but I *think* it might be issues with the way that the Calcite jdbc driver was implemented; but looking at the mailing list and issues others are in a similar boat and they wouldn't be opposed to changes.

MikeNeilson:
After dealing with this, and reading the XmlLoadingApp and XmlScheduleEntry DAOs I don't think this will be needed.
Certainly an interesting idea, I think we may have an easier path forward just making XML DAO and possibly some minor tweaks to the XML database itself.

PR #412, opened 2023-12-13
Bugfix/profile handling

### LINKED ISSUE #289: Profiles handling within LauncherFrame

LauncherFrame/Profiles

**Purpose of refactoring**

The current Profile system has two different but identical(ish) functions for loading the profiles.
Additionally the current code while it looks at the DCSTOOL_USERDIR assumes "decodes.properties" instead of "user.properties"
as the default profile path.

**Proposed method**

- Add type safety, ProcCallBack takes an Object instead of a Generic and then does conversions, this can cause unexpected and confusing class cast exceptions
  - Given the functional interfaces and futures of Java 8+ it also may no longer be necessary; if so new standards should be used
- Currently the LauncherFrame detects on button click if a non-default profile is used and goes down a different code path.
  - I think all "button clicks" should just go down the path of getting launched that way. This would reduce mental load while debugging and make it easier to say, put in a "fake" ProfileServerThread so testing can verify commands are sent correctly without actually having to start anything.
  - BasicServer is used, but this really doesn't need to use this style of IPC and should be investigated.

The short version is:
Instead of profileCombo being `JComboBox<String>` it will become `JComboBox<Profile>` and every button click can just get the current Profile and not have to worry about determining the correct path later, it would already be set.

The Profile class should get a static member `List<Profile>` getProfiles(Path searchDir)`

The Profile Server Thread would also just take this object and use it for lookup on whatever method it ends up using.

The model in the profile viewer/creator could then also be extended as the Profile Model can hold more data such as the database type.

**Additional context**

### LINKED ISSUE #405: OpenDCS Toolkit Profiles
**Describe the bug**
When you open the toolkit (using dcstool_start.bat or launcher_start.bat), if you have profiles enabled there is a dropdown on top to choose which profile to log in with.  NWDM has this enabled to make COOPing easier - the user can select a profile rather than editing their decodes.properties file every time.  We have two profiles, NWO and NWK.  There is also the "default" profile.  In older versions of OpenDCS, utilizing the profiles worked as expected.  In the newest version of the GUI, our added profiles do not work.  The toolkit allows you to select them, but clicking on a component takes you to the login screen and then disappears without ever popping up a new GUI of that component.  Keeping the Profile as "default" will work to open the additional components.

**To Reproduce**
1. Go to dcstool_start.bat or launcher_start.bat, double-click to open toolkit
2. On the Profile dropdown, select NWO-T7 (rather than the default of "default")
image
4. Click on DECODES Database Editor, or another component
image
5. Enter username/password.  The loading screen and login screens disappear, and no additional GUI ever opens.

**Expected behavior**
The additional component GUI should open after login.

 - OS: Windows 10
 - Version OPENDCS 7.0.10

**Additional Context**
When using the "default" profile, log information is populated in the command window that opens with the batch file.  When using the user-defined profile, no log information is populated.

MikeNeilson:
## Problem Description

Fixes #405.
Closes #289. 

## Solution

- Modified ProcWaitThread to dump the streams to the logs applications behaved.
  - Applications run from the launcher were starting but would than hang. The issue appears to have been the OutputStreams (stdout, stderr) of the applications being ignored and then writes waiting forever (even jstack,jconsole, and visualvm couldn't not connect and would also hang.)
- Created a "Profile" object to hold the simple name, full path (java.io.File), and default/profile determination. Replaced duplicate code with new Profile functions to retrieve all available or a specific profile.
- LauncherFrame `profileCombo` is now typed on `Profile`
- Updated `ProfileManager` to use standard `javax.swing` interfaces
  - Uses defaultTableSorter
  - usages of getRow/Column index properly convert to appropriate model index
  - Cells are rendered with proper type
  - TableModel is typed on `Profile`
  - Added ToolTip to show `File::getAbsolutePath` for the Profile
  - Simplified operation logic given new Objects and TableModel design.

NOTE: I did whitespace formatting fixes (tabs and spaces at EOL), turn on GitHubs "ignore whitespace" feature during review.

TODO:
 - [x] Additional test of the ProfileManagerFrame. This particular table panel is completely isolated from the databases and is suitable for a proper automated test demonstration. I may also not be doing certain operations on the right thread and assertj-swing will be very direct about it.
 - [-] Delete? LauncherAction classes and implementation. They appear to be from an attempt to simplify the logic and do not appear to be needed. (Already too many changed files, will handle in a later PR)

## how you tested the change

## Where the following done:

   - [X] Integration tests created or modified that run during integration testing

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,9 +37,18 @@ jobs:
           ant integration-test -D"no.docs=true" -D"opendcs.test.engine=OpenDCS-Postgres"
       - name: Gui Tests
         if: matrix.platform == 'ubuntu-latest'
+        # we need time kill Xvfb and let ffmpeg write out the video file
+        shell: bash {0}
         run: |
-          sudo apt install xvfb
-          xvfb-run ant gui-test -D"no.docs=true"
+          sudo apt install xvfb ffmpeg
+          export DISPLAY=:99.0
+          Xvfb $DISPLAY -screen 0 1920x1080x24 &
+          ffmpeg -nostdin -y -f x11grab -video_size 1920x1080 -i "$DISPLAY" build/test-gui.webm &
+          ant gui-test -D"no.docs=true"
+          exit_val=$?
+          pkill Xvfb
+          sleep 1
+          exit $exit_val
       - name: Code Coverage
         run: |
           ant coverage.report
@@ -54,6 +63,7 @@ jobs:
           path: |
             build/reports/**
             build/test-integration/**/tmp/configs**
+            build/test-gui.webm
 #docker-images:
 #  runs-on: ubuntu-latest
 #  steps:
--- a/build.xml
+++ b/build.xml
@@ -392,7 +392,7 @@ more information about each.
             <fileset dir="${build.dir}/test-gui/results">
                 <include name="TEST-*.xml"/>
             </fileset>
-            <report format="frames" todir="${junit.html.output.dir}/test-gui"/>
+            <report format="noframes" todir="${junit.html.output.dir}/test-gui"/>
         </junitreport>
         <if>
             <isset property="gui.failed"/>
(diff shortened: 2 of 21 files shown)


ktarbet (approved):
Looks good.  I have some suggestions here:

https://github.com/opendcs/opendcs/pull/418

PR #414, opened 2023-12-14
Bugfix/413 decodes message browser

### LINKED ISSUE #413: Decoding issues using the GUI Interface
While completing initial testing on 7.0.11 I found issues decoding data via the GUI interface using platforms and decoding configurations that worked properly on 7.0.8. After talking with our IT team they are also completing similar testing and have not experienced the issue when using scripts to run the decoding config. As a result, the problem appears to only occur when using the GUI interface

1. Setup a know Platform with a decoding configuration that has worked using the GUI interface on previous versions
2. Use the GUI interface to bring up the DCP Message Browser
3. Run the message browser to show both Raw and Decoded values for the Platform
Example of 7.0.11 not decoding properly (doesn't appear to find the Platform in general)
image

**Expected behavior**
Example of 7.0.8 properly decoding with the same Platform and Decoding Config:
image

**Desktop/ServerSide:**
image

Platform and decoding configuration (in TXT format) that were used for testing both 7.0.8 and 7.0.11:

[p00130.txt](https://github.com/opendcs/opendcs/files/13676049/p00130.txt)

[Nos6Min.txt](https://github.com/opendcs/opendcs/files/13676052/Nos6Min.txt)

MikeNeilson:
## Problem Description

Fixes #413. 

## Solution

Inject Instance of Database. 
NOTE: Still using the Deprecated `Database::getDb`, removal of such usage will require larger structural changes to
that that interface.

## how you tested the change

1. Opened Message Browser (launcher_start/dcstool_start or `msgaccess`) connecting to a database of some sort
2. Configure search criteria to known data (you have platforms and decodes scripts configured)
3. Change "Show" to `Raw and Decoded` or `Decoded`
4. Choose display all or next and see appropriately decoded output.

## Where the following done:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,15 +13,20 @@ jobs:
         platform: [ubuntu-latest,macos-latest,windows-latest]
     runs-on: ${{matrix.platform}}
     steps:
-      - uses: actions/checkout@v3.5.2
+      - uses: actions/checkout@v4.1.1
       - name: Set up JDK
-        uses: actions/setup-java@v3.11.0
+        uses: actions/setup-java@v4.0.0
         with:
           java-version: '8'
-          distribution: adopt
-      - uses: actions/setup-python@v3.1.3
+          distribution: temurin
+      - uses: actions/setup-python@v4.0.0
         with:
           python-version: "3.8"
+      - uses: actions/cache@v3.3.2
+        with:
+          path: ~/.ant
+          key: ${{ matrix.platform }}-ivy-${{hashFiles('ivy.xml')}}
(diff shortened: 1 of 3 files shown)


MikeNeilson:
This is done, but I accidentally based in on the profile handling PR. Will mark ready after that merges in.

MikeNeilson:
I've updated the github actions versions. Especially for update the performance change was drastic.

Also migrated to Temurin JDK (the old adoptjdk was picked up by Eclipse and rename, may help with the flaky swing test if there was a specific bug in that previous version. Was using OpenJDK locally.)

ktarbet (approved):

PR #417, opened 2023-12-15
Cache .ant directory to cache ivy dependencies for build.

MikeNeilson:
## Problem Description

Builds are slower than they need to be, often have issues downloading the CWMS Deps due to network issues.

## Solution

Cache the ivy dependencies so that there is less network traffic during the builds.

## how you tested the change

Will repeat action manually after it runs during this PR.

## Where the following done:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,6 +22,11 @@ jobs:
       - uses: actions/setup-python@v3.1.3
         with:
           python-version: "3.8"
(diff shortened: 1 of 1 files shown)


MikeNeilson:
Wow, that was drastic. Between the cache and updating the upload-artifact action that went from like 11 minutes per build to < 3 minutes per build.

Once #412 and #414  are merged in I'll rebase this and set it for ready.... or just make the changes as part of those if I need to make additional changes.

PR #431, opened 2023-12-21
Bugfix/429 configeditor over commiting objects

### LINKED ISSUE #429: Committing a config file also commits all platforms associated with that config
**Describe the bug**
When committing an open config file, all the platforms connected to that config are also committed, even if no platforms are open in DECODES and in spite of not being on that tab.

**To Reproduce**
1. Go to the NWDM DECODES Database
2. Activate the Configs tab, and then click on and open a config that is connected to a platform that does not have a Site associated with it.  When we found this but, it was with the web-CODWR-Flow config file.  Those platforms have been deleted from the database now, but the NWK-CLIM-001 config would allow you to reproduce this error now with only one platform instead of multiple platforms.
3. Hit Commit on the open config file.  An error GUI will pop up stating "Cannot save platform without site assignment".  In NWDM's case, this Error pop-up occurred multiple times (there were multiple platforms without sites connected to this config).  In the example case that you can use in testing, there is only one platform connected to this GUI without a site and the error only pops up once before allowing you to move around in the editor again.
4. See error screenshot below.
image

**Expected behavior**
The expected behavior of a "Commit" on a specific file is that it would only commit that open file.  Not any other files on that tab or on any other tabs.  File --> Commit All is what we would expect would commit all files (or all open files).

**Screenshots**
Open NWK-CLIM-001 config file.
image

Hit Commit, see error:
image

See related platform that is causing the error:
image

 - OS: Windows 10, OpenDCS 7.0.1
 - Version 6.8.14

**Additional context**

MikeNeilson:
## Problem Description

Fixes #429. 

The SQL databases use an integer foreign key to reference the configs and scripts; when updating a config it's not needed to alter the platform definition in any way.
However, the XML database embeds the full config within the platform definition instead of providing a reference of some sort.

## Solution

Detect if XML database, which requires this level of write out, and only saves platforms if it is the XML database.

### Manually (XML and Postgres)

1. Open dbedit (have existing database of any size)
2. Open Config list
3. Edit a config and add a new script
4. Close scripteditor
5. Click commit
6. Go to platforms tab and open platform that uses the above config
  a.  XML -> See that Last Modified has changed
  b. Postgres (or any sql) -> See that Last Modified has *NOT* changed.
8. Edit a transport medium and see the the new config in the drop down list

TODO: test the NWDM dataset that lost data because of this to really verify platforms are unnecessarily saved.

## Where the following done:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,6 +45,7 @@ jobs:
         # we need time kill Xvfb and let ffmpeg write out the video file
         shell: bash {0}
         run: |
+          sudo apt update -y && sudo apt upgrade -y
           sudo apt install xvfb ffmpeg
           # Precompile the test code so the recording isn't as long.
           ant compile-test-gui -D"no.docs=true"
--- a/src/main/java/decodes/dbeditor/ConfigEditPanel.java
+++ b/src/main/java/decodes/dbeditor/ConfigEditPanel.java
@@ -40,6 +40,7 @@ import decodes.db.DecodesScript;
 import decodes.db.DecodesScriptException;
 import decodes.db.UnitConverterDb;
 import decodes.util.TimeOfDay;
+import decodes.xml.XmlDatabaseIO;
 import decodes.db.Database;
 import decodes.db.DatabaseException;
 import decodes.db.Platform;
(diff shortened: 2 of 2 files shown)


ktarbet (approved):

MikeNeilson:
Verified with the discovered data set that the Platforms *aren't* accidentally deleted with the SQL database.

PR #435, opened 2024-01-02
Additional support for multiple offices.

MikeNeilson:
## Problem Description

The CWMS implementation includes an office_code column on any directly referenced tables.
Previously a default office code was set. This worked fine for reading data, however when writing, regardless of which offices was set in user.properties, the default would always be used preventing any actual multi office use.

## Solution

Create an insert/update trigger on each table to verify that either a value is provided or pull it from the current session.

## how you tested the change

Tested on a database database inserting data to verify the correct office was set.

## Where the following done:

### DIFF
--- a/schema/cwms/update68to_70.sql
+++ b/schema/cwms/update68to_70.sql
@@ -1,0 +1,32 @@
+create or replace procedure set_office_column(new_office IN OUT cwms_20.av_office.office_code%type,
+                                              old_office cwms_20.av_office.office_code%type)
+is
+begin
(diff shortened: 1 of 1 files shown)


ktarbet (approved):
Looks ok to me.

MikeNeilson:
Thanks, does need to be rebase and rewritten for the new directory structure and flyway scheme.

MikeNeilson:
Yeah, I need to alter it to a follow on migration now that that's merged in.

PR #433, opened 2023-12-28
Feature/421 migrate pg schema flyway

MikeNeilson:
## Problem Description

Fixes 421. 

Describe the problem you are trying to solve.

## Solution

Move custom shell script and SQL files to use the Flyway schema migration mechanism for installing the database.

JDBI is more directly introduced in this PR; and extremely simple DataSource was created to get this running.
Additionally because this code is intended to be executed *before* there is a DECODES/TSDB database available
all existing DAO/TimeseriesDb/SQLdatabaseIO helpers are ignored to avoid difficult to diagnose edge cases.

TODO:
- [x] Move flyway.migrate call code out of integration test code to main code
- [x] Create CLI application to allow users to perform schema install/update
- [] Baseline import for Decodes Data
  - [x] with no versioning support
  - [ ] with versioning support (will be moving to the flyway infrastructure now that some operations are cleaned up.) 
- [] Baseline import for Computation Data
  - [x]  with no versioning support
  - [ ] with versioning support
- [x] ~Create GUI? application to allow users to perform schema install/update~ Going to do GUI as a separate PR; this is already big enough.
- [x] ~Create IZPACK panel to allow GUI to allow user to perform schema install/update~ Actually this can happen later, if at all.

Those are all listed as other issues but with the current scripts now deleted it would be difficult (though not impossible) for a user to actually, you know, install the database for the OpenDCS-Postgres version.

## how you tested the change

The Schema migrate is currently used in the integration test; this is simply moving it to a new location and better organizing all the requirements.

Manually tested the migration app for the case of new database and existing database.

## Where the following done:

- [ ] Tests. Check all that apply:
   - [ ] Unit tests created or modified that run during ant test.
   - [x] Integration tests created or modified that run during integration testing
         (Formerly called regression tests.)
   - [ ] Test procedure descriptions for manual testing
- [x] Was relevant documentation updated?
- [ ] Were relevant config element (e.g. XML data) updated as appropriate

If you aren't sure leave unchecked and we will help guide you to want needs changing where.

PR #439, opened 2024-01-05
remove duplicate sequence creation

adamkorynta:
## Problem Description

PostgreSQL schema installation has a duplicate TSDB_DATA_SOURCEIdSeq sequence creation.

## Solution

Only create the sequence once

## how you tested the change

Running schema installation against postgresql docker container.

## Where the following done:

If there are tests that I can update/create for this, can you point me at them?


MikeNeilson:
This is getting fixed in #433 . I just verified it's not duplicate in the source source schema file.

PR #441, opened 2024-01-05
Refactor/comp depends improvements

MikeNeilson:
## Problem Description

NOTE: I needed the transaction support, this PR is based off of #432 

CompDepends runs continuously and uses small queries that don't use bind vars. Additionally there are options on multiple tables that should handle within a single transaction to ensure proper state of the database.

## Solution

Use the new handlers, move SQL operations out of CpCompDependsUpdater into the DAO object.
Allow the dao to start a transaction of itself.

## how you tested the change

## Where the following done:


MikeNeilson:
This clearly needs a pile of specific tests; at the least it did not break the current integration tests which do depend on this set of data.

If anyone knows of a well defined set of data please point me at it.

src/test-gui/java/decodes/launcher/ProfileManagerFrameTest.java
```diff
@@ -21,7 +21,10 @@
 import org.assertj.swing.timing.Timeout;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
```
agilmore2, line 24: Is this import needed now?

MikeNeilson: Nope, thanks. that got pulled in right before I switched to the EnableIfSystemProperty.

src/main/java/decodes/tsdb/CpCompDependsUpdater.java
```diff
@@ -239,10 +242,8 @@ public void runApp( )
 		String action="";
 		while(!shutdownFlag)
 		{
-			LoadingAppDAI loadingAppDAO = theDb.makeLoadingAppDAO();
-			CompDependsDAI compDependsDAO = theDb.makeCompDependsDAO();
-
-			try
+			try(LoadingAppDAI loadingAppDAO = theDb.makeLoadingAppDAO();
+				CompDependsDAI compDependsDAO = theDb.makeCompDependsDAO();)
```
agilmore2, line 246: Indention off?

src/main/java/decodes/tsdb/CpCompDependsUpdater.java
```diff
@@ -620,15 +614,12 @@ private boolean tsCreated(DbKey tsKey)
 		{
 			String msg = "Error processing TS_CREATED for key=" + tsKey + ": " + ex;
 			warning(msg);
-			System.err.println(msg);
-			ex.printStackTrace(System.err);
+			StringWriter sw = new StringWriter();
+			PrintWriter pw = new PrintWriter(sw);
+			ex.printStackTrace(pw);
+			warning(sw.toString());
```
agilmore2, line 620: sw never initialized to anything here?

MikeNeilson: `StringWriter sw = new StringWriter()` is the initialization. StringWriter is a "StringBuffer" but for a print stream.

src/main/java/decodes/tsdb/CpCompDependsUpdater.java
```diff
@@ -671,21 +672,22 @@ private boolean tsDeleted(DbKey tsKey)
...
 								+ (theDb.getTsdbVersion() >= TsdbDatabaseVersion.VERSION_9 ? "ts_id" : "data_id")
-								+ " = " + tsKey;
+								+ " = ?";
 							try
 							{
-								tsGroupDAO.doModify(q);
+								dao.doModify(q,grp.getGroupId(), tsKey);
 							}
-							catch (DbIoException ex)
+							catch (SQLException ex)
```
agilmore2, line 690: Probably need to print error with the specific keys as well.

src/main/java/decodes/tsdb/CpCompDependsUpdater.java
MikeNeilson: Yep, that file is still in tabs.

src/main/java/decodes/tsdb/CpCompDependsUpdater.java
```diff
@@ -709,13 +712,10 @@ private boolean tsDeleted(DbKey tsKey)
...
 	
-			// Delete from cp_comp_depends table any tupple with this ts_id.
-			q = "delete from CP_COMP_DEPENDS " + "where TS_ID = " + tsKey;
-
-			tsGroupDAO.doModify(q);
+			compDependsDao.deleteCompDependsForTsKey(tsKey);
 			return true;
 		}
-		catch (DbIoException ex)
+		catch (SQLException | DbIoException ex)
```
agilmore2, line 718: Likewise here

src/main/java/decodes/tsdb/CpCompDependsUpdater.java
MikeNeilson: oh, yeah, now that it's just in the query string.

src/main/java/decodes/tsdb/CpCompDependsUpdater.java
agilmore2: right, but does printStackTrace(pw) cause sw to get a value as well? Sorry, don't know these interfaces well.

src/main/java/decodes/tsdb/CpCompDependsUpdater.java
agilmore2: Also odd we're replacing system.err.print above, but not here?

src/main/java/decodes/tsdb/CpCompDependsUpdater.java
```diff
@@ -1000,18 +986,29 @@ public void evalComp(DbComputation comp)
...
 			{
-				try { writeToAdd2Db(comp.getId()); }
-				catch(DbIoException ex) { /* do nothing -- err msg already logged. */ }
+				try
+				{
+					writeToAdd2Db(comp.getId());
+				}
+				catch(DbIoException ex)
+				{
+					/* do nothing -- err msg already logged. */
```
agilmore2: except we are doing something?

src/main/java/decodes/tsdb/CpCompDependsUpdater.java
MikeNeilson: no, the PrintWriter is initalized with the StringWriter, which initialized to the empty string.
StringWriter is the low level interface, PrintWriter provides the same interface as System.out basically. So `::printStackTrace(pw)` will write to the StringWriter, through the PrinterWriter, as though it was just outputing to stdout.

As to way Java decided they needed that many levels of indirection for this particular case; you'd have to ask the language/library designers.

src/main/java/decodes/tsdb/CpCompDependsUpdater.java
MikeNeilson: To be fair to myself, that was already in place and wasn't entirely sure if I wanted to deal with replacing it right now.
Probably good to at least log a warning message though in case it *wasn't* actually already logged.

src/main/java/decodes/tsdb/CpCompDependsUpdater.java
MikeNeilson: That was probably just me being lazy, but I'll go through and replace all the instances within this file.

src/main/java/opendcs/dao/CompDependsDAO.java
```diff
@@ -259,27 +241,38 @@ public void deleteCompDependsForTsKey(DbKey timeSeriesKey)
 		throws DbIoException
 	{
 		if (db.isHdb())
+		{
```
agilmore2, line 244: What's different about the HDB instances that we're not doing deletions from the depends table?

src/main/java/opendcs/dao/CompDependsNotifyDAO.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+        catch (SQLException ex)
+        {
+            throw new DbIoException("Unable to retrieve notification records.", ex);
+        }
+    }
+
+    @Override
+    public CpDependsNotify getNextRecord() throws DbIoException {
+        if (db.getTsdbVersion() < TsdbDatabaseVersion.VERSION_8)
```
agilmore2, line 52: indention here is messed up?

src/main/java/opendcs/dao/CompDependsDAO.java
MikeNeilson: I have absolutely no idea which is why I left it in place.

agilmore2: It's only called from a chain that deletes a Timeseries completely. We don't do that here.

PR #444, opened 2024-01-10
DACQ Event DAO improvements

### LINKED ISSUE #443: DacqEventDAO running query more then necessary.
**Describe the bug**
DacqEventDAO in the constructor does a pythonesque try-query-to-determine-state but does not cache the result in a 
static variable.

**To Reproduce**
1. Run routingscheduler
2. See database issues, such as performance

**Expected behavior**
System should not make such a repeated query as the columns aren't likely to change in the middle of operations.

**Screenshots**

 - OS: AWS Oracle RDS/ECS
 - Version: 7.0.11

**Additional context**
This is probably not the main issue given a recent total performance slump we saw; but it was at the top of the report.

MikeNeilson:
## Problem Description

**** NOTE ****

Apparently I'm not as awake as I thought I was this morning and pushed this directly to master. I'll be adding the tests in the PR, the changes are already in.

Fixes #443. 

## Solution

Static var for column check to reduce database queries.
Bind vars for all queries

## how you tested the change
will make tests.

## Where the following done:

### DIFF
--- a/src/main/java/opendcs/dao/DacqEventDAO.java
+++ b/src/main/java/opendcs/dao/DacqEventDAO.java
@@ -136,7 +136,7 @@ public class DacqEventDAO extends DaoBase implements DacqEventDAI
 
 		ArrayList<Object> parameters = new ArrayList<>();
 		String q = "SELECT " + dacqEventColumns + " FROM " + dacqEventTableName
-			+ " WHERE EVENT_TEXT LIKE '%' || '?' || '%'";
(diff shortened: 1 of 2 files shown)


ktarbet (approved):
How is the  CP_COMP_TASKLIST different than DACQ_EVENT?

MikeNeilson:
cp_comp_tasklist is time series database data stored events.

dacq_event is for data processing events so certain programs can see the log. I'd actually prefer to rip it out entirely in favor of pushing the same data to an external logger for that sort of monitoring.

PR #456, opened 2024-01-25
Bugfix/lrgs lock file

MikeNeilson:
## Problem Description

The lock file isn't required in container environment so allow user specify such.

## Solution

Moved ServerLock to be an interface and original became "FileServerLock" and greated a "DockerServerLock"
That basically asserts it's always locked.

NOTE: do not merge during the 7.0.12 release cycle.

## how you tested the change

Manually locally and within USACE environment.

## Where the following done:

### DIFF
--- a/docs/source/dev-docs.rst
+++ b/docs/source/dev-docs.rst
@@ -512,6 +512,8 @@ lrgs.sh handles first time setup, copy default config, initial admin user, and s
 The lrgs.lock file is currently ignored and docker just kills the process. Currently investigating better ways to 
 handle shutdown. Will likely just add a flag to remove the lock file entirely.
 
+The docker environment no uses the special sequence `lrgsStart -F -k -` to use the NoOpServerLock which causes
+the applications using that Lock to assume it's always valid for them.
 
 Database Scripts
 ================
--- a/docs/source/dev-docs.rst
+++ b/docs/source/dev-docs.rst
@@ -512,7 +512,7 @@ lrgs.sh handles first time setup, copy default config, initial admin user, and s
 The lrgs.lock file is currently ignored and docker just kills the process. Currently investigating better ways to 
 handle shutdown. Will likely just add a flag to remove the lock file entirely.
 
-The docker environment no uses the special sequence `lrgsStart -F -k -` to use the NoOpServerLock which causes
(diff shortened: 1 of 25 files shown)


src/main/java/lrgs/lrgsmain/LrgsMain.java
```diff
@@ -171,14 +173,22 @@ public void run()
         // Establish a server lock file & start the server lock monitor
 
         String lockName = EnvExpander.expand(cmdLineArgs.getLockFile());
-        myServerLock = new ServerLock(lockName);
-        if (!myServerLock.obtainLock(this))
+        Logger.instance().info("Lock File =" + lockName);
+        if (lockName.equals("-"))
```
ktarbet, line 177: consider documenting the special hyphen argument in the argument usage.

src/main/java/ilex/util/DockerServerLock.java
ktarbet, line 1: If this could be used outside the Docker context, it could be called something like NoOpServerLock.java

MikeNeilson: That's a good idea.

src/main/java/lrgs/lrgsmain/LrgsMain.java
MikeNeilson: I have, though in the dev docs for now.

docs/source/dev-docs.rst
```diff
@@ -512,6 +512,8 @@ lrgs.sh handles first time setup, copy default config, initial admin user, and s
 The lrgs.lock file is currently ignored and docker just kills the process. Currently investigating better ways to 
 handle shutdown. Will likely just add a flag to remove the lock file entirely.
 
+The docker environment no uses the special sequence `lrgsStart -F -k -` to use the NoOpServerLock which causes
+the applications using that Lock to assume it's always valid for them.
```
ktarbet, line 516:
```suggestion
The docker environment now uses the special sequence `lrgsStart -F -k -` to run in the foreground (-F) and use the NoOpServerLock file (-k -)  which causes
the applications using that Lock to assume it's always valid for them.
```

ktarbet (approved):
approved, with suggestion in dev-docs.rst

PR #457, opened 2024-01-26
Gui Apps in CWMS will now use auth source regardless of environment.

MikeNeilson:
## Problem Description

Primarily this is to avoid needing users to enter credentials on a shared system. However it also helps reduce certain edge cases.
e.g. TsdbAppTemplate shouldn't have a clue that "CWMS" even exists.

## Solution

Allow successful retreieval of "AuthSource" credentials to be used for any type of application.
NOTE: current solution requires user to login twice. This is due to LoginDialog not caching credentials and the system needs to login twice. the long term solution is to not have a "DECODES" and "TimeSeries" database but instead "A database."
In the meantime we have have to cache credentials. I will investigate later. 

NOTE: Do not merge during 7.0.12 cycle, this is primarily for commentary and CWMS usage.

## how you tested the change

Manually in VM/Linux.

## Where the following done:

### DIFF
--- a/src/main/java/decodes/tsdb/TsdbAppTemplate.java
+++ b/src/main/java/decodes/tsdb/TsdbAppTemplate.java
@@ -73,7 +73,11 @@
 package decodes.tsdb;
 
 import java.io.PrintStream;
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.util.Properties;
+
+import javax.swing.JOptionPane;
 
 import org.opendcs.authentication.AuthSourceService;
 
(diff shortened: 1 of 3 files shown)


src/main/java/decodes/tsdb/TsdbAppTemplate.java
```diff
@@ -244,7 +248,12 @@ public void execute(String args[])
 			}
 			catch(BadConnectException ex)
 			{
+				JOptionPane.showMessageDialog(null,ex.getLocalizedMessage());
```
ktarbet, line 251: If this is running as service , the dialog may cause an issue?

MikeNeilson: Oh yeah, TsdbAppTemplate isn't gui only, I should check before popping that open. I'm not super worried about the current usage though.

MikeNeilson: I'll be going through and cleaning this up after the current release cycle.

If I'm lucky this might actually be the path forward on properly "merging" the DECODES and Time Series database within the code.

ktarbet (approved):

ktarbet:
[x] I tested this on a windows system - The .decodes.auth file was read correctly, and no password prompt occurred.

MikeNeilson:
overcome by events.

PR #458, opened 2024-01-29
Fixes #449 -  Python Alogorithm Scripts - Column sorting issue

### LINKED ISSUE #449: Python Alogorithm Scripts - Column sorting issue
**Describe the bug**
Sorting confuses the 'Select CP Functions Function' dialog
**To Reproduce**
using windows Client (building from master 7.0.12), connected to a CWMS Database

1. open Computations
2. Open a python algorithm
3. Click 'CCP Built-in' button
4. Click Name column to sort.
5.Click the 'screening' under Name.

Before Sorting Description is correct as shown for screening for example:
image

After Sorting
image

ktarbet:
Fixes #449 

## Solution

Using basic JTable features to get the job done, while removing code.

## how you tested the change

Startup OpenDCS , Click Alogrithms, pick a PythonAlgorithm,  Click the Python Button,  Click Python Built-in.
Select functions and insert, try sorting and verify descriptions stay in sync.

### DIFF
--- a/src/main/java/decodes/tsdb/compedit/PyFuncSelectDialog.java
+++ b/src/main/java/decodes/tsdb/compedit/PyFuncSelectDialog.java
@@ -58,12 +58,18 @@ public class PyFuncSelectDialog extends JDialog
 		cancelButton.addActionListener(e -> cancelPressed());
 		buttonPanel.add(cancelButton);
 
-		model = new DefaultTableModel(new String[]{"Name", "Signature","Function"},0);
+		model = new DefaultTableModel(new String[]{"Name", "Signature","Function"},0){
+			@Override
+			public boolean isCellEditable(int row, int column) {
+				return false; // Disable cell editing
+			}
+		};
 		for (PyFunction item : funcList.getList()) {
 			model.addRow(new Object[]{item.getName(), item.getSignature(), item});
 		}
 		tab = new JTable(model);
-		tab.setCellSelectionEnabled(false);
+		tab.setRowSelectionAllowed(true);
+		tab.setColumnSelectionAllowed(false);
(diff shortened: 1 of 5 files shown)


src/main/java/decodes/tsdb/compedit/PyFuncSelectDialog.java
```diff
@@ -77,31 +50,24 @@ public PyFuncSelectDialog(JFrame theFrame, PyFuncList funcList)
...
-			}
-		});
+		cancelButton.addActionListener(e -> cancelPressed());
 		buttonPanel.add(cancelButton);
 
-		// Set up the table
-		if (model == null)
-			model = new FuncListTableModel(funcList);
-		tab = new SortingListTable(model, new int[] {30,70});
+		model = new DefaultTableModel(new String[]{"Name", "Signature","Function"},0);
```
MikeNeilson, line 61: Since we're making this change anyways we should put these labels in the resource bundle: https://github.com/opendcs/opendcs/blob/master/src/main/resources/decodes/resources/compedit_en_US.properties

For future internationalization.

MikeNeilson (changes requested):
This looks good. The labels from a resource can happen later, but if it seems easier to fix now we should just do it.

MikeNeilson:
I'm going to argue these table sorting fixes are an "already identified" issues and can and should go in during the release candidate window.

src/main/java/decodes/tsdb/compedit/PyFuncSelectDialog.java
ktarbet: done in latest commit

MikeNeilson (approved):

PR #465, opened 2024-02-02
Refactor/comp manual run to background

MikeNeilson:
## Problem Description

When someone runs a group computation with a lot of groups it's hard to know if CompEdit is hung/frozen.

## Solution

Moved computation progressing to SwingWorker instance and created a progress bar.

Additionally, since the GuiThread is no longer occupied, the TraceDialog can be opened immediately immediately and support for auto scroll was added.

## how you tested the change

Manually:
- Ran Group comp with many inputs
- Closed comp, found another, ran again.
- Enabled/Disabled autoscroll

Pictures:
Showing Progress -with Error message

Additional Progress and More Messages in the trace log

The progress bar location is frankly terrible but it was the easiest location to just insert for initial testing.
Frankly I'm just stoke that this tiny portion wasn't that difficult

## Where the following done:

### DIFF
--- a/src/main/resources/decodes/resources/comprun_en_US.properties
+++ b/src/main/resources/decodes/resources/comprun_en_US.properties
@@ -15,6 +15,7 @@ RunComputationsFrame.saveOutputButton = Save Output Data
 RunComputationsFrame.timeLabel = Time
 RunComputationsFrame.noCompSelectedErr = No computations selected to run!
 RunComputationsFrame.saveCompOutput = Save Computation Output?
+RunComputationsFrame.cancelComputationExecution = Cancel Computation Execution?
 TimeSeriesTable.dateTimeColumnLabel = Date/Time
 RunComputationsFrame.inputLabel = I:
 RunComputationsFrame.outputLabel = O:
--- a/build.xml
+++ b/build.xml
@@ -1,19 +1,11 @@
 <?xml version="1.0"?>
-
 <!--
-
 This is the ANT build file for OpenDCS 6.7 RC02 and Later.
 The intent is that this will eventually be replaced with a Maven pom.xml.
-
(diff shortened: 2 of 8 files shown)


MikeNeilson:
@hoffmanjs @opendcs/opendcs-core-devs 

FYI, for those that don't already know, the Eclipse WindowBuilder plugin works to edit the exist GUI code. 
So far as I know it's the only GUI Builder for Swing that is free, and the only one that doesn't require a specific format.

I'm still going to use VSCode for my main code development but I was able to update the eclipse-ide-task to better generate the required .project and .classpath files so that should work better in both eclipse and vscode after this gets merged.

Also it's not perfect, the panel for manual comps renders a placeholder at least, but it doesn't show up. and I did cause it to freeze a few times (eclipse itself).

Something to keep in made for anyone tweaking the actual GUIs.

ktarbet (approved):

src/main/resources/decodes/resources/comprun_en_US.properties
```diff
@@ -15,6 +15,7 @@ RunComputationsFrame.saveOutputButton = Save Output Data
 RunComputationsFrame.timeLabel = Time
 RunComputationsFrame.noCompSelectedErr = No computations selected to run!
 RunComputationsFrame.saveCompOutput = Save Computation Output?
+RunComputationsFrame.cancelComputationExecution = Cancel Computation Execution?
```
MikeNeilson, line 18: @ktarbet if you can paste the appropriate Spanish for this as a reply I can add it to the es_US.properties file.

MikeNeilson:
I'm going to wait to merge this in until the flyway branch is merged in. I made some changes here to logging that will need a rebase.

PR #469, opened 2024-02-05
Fixed column sorting issues for Copy and Delete functions for Schedule Entry

hoffmanjs:
## Problem Description
Column sorting gives erroneous results when trying to copy or delete in schedule entry.

In Schedule Entry, sorting columns results in the wrong row being selected when trying to copy or delete a row.

## Solution
Modified the code to use the table model when selecting a row.

## how you tested the change
Open Decodes Database Editor, select the Schedule Entry tab.  Sort columns and then select a row to copy or delete.  Verify the correct row is selected for the copy or delete.

## Where the following done:


MikeNeilson (approved):
You want a laugh, I just had that bite me testing something and made the sameish fix.

PR #471, opened 2024-02-05
Bugfix/row model sorting

MikeNeilson:
## Problem Description

We keep running into problems where sorting tables causes wrong elements to open/delete/etc.

## Solution

Slogged through the the code and changed all of them.

## how you tested the change

Will be writing a set of manual tests to cover the behavior of the of all of the tables... which is the next slog.

NOTE: This is way too much, I'm not really expecting to get this all done before the 7.0.12 release cycle. But who knows, we may get lucky.

## Where the following done:


MikeNeilson:
well, I did find at least one place where adding this conversion breaks things. More work required. Probably not going to get it done this week.

MikeNeilson:
In hindsight i went too far and ended up breaking several.

Issue #476, opened 2024-02-08
DCSTOOL_USERDIR consistency issues


ktarbet:
**Describe the bug**
The default saving location for .decodes.auth is different than the default reading location.

**To Reproduce**

1. create simple decodes.properties with EditDatabaseType:OPENTSDB  *without* defining DbAuthFile  
2. run setDecodesUser   - config is saved to $HOME/.decodes.auth   (example: c:\users\ktarbet\.decodes.auth )
3.  run dbedit    -- It looks in $DCSTOOL_USERDIR/.decodes.auth   (example: C:\test\OPENDCS7.0.12-RC04/.decodes.auth )

```text
Cannot read username and password from 'C:\test\OPENDCS7.0.12-RC04/.decodes.auth' (run setDecodesUser first): ilex.util.AuthException: Unable to read auth file
Exception in thread "main" ...
```

**Expected behavior**
default locations should be consistent.

**Additional context**
The DCSTOOL_USERDIR env variable has many possible definitions.  A more consistent approach will be easier to document and debug. 

Unix ./decj script has four potential settings of DCSTOOL_USERDIR listed in order of preference:

1.  User defined env
2. $DCSTOOL_HOME  if the running user owns the install directory
3.  $HOME/opendcs_user  if that directory exists
4.  $HOME  otherwise
   

Windows decj.bat hardcodes DCSTOOL_USERDIR=$INSTALL_PATH

** proposed fix **

If the DCSTOOL_USERDIR variable is not set default to $HOME/.opendcs/ for .decodes.auth, user.properties,  my-special.profile, and log files; do this  for all applications including setDecodesUser.

Additionally investigate if DCSTOOL_HOME and DECODES_INSTALL_DIR can be combined.

ktarbet:
@agilmore2 - Do you have any suggestions on this issue?

ktarbet:
Note 1:
 DecodePropsPanel.Java  has a default hardcoded in the UI for DbAuthFile
 "$HOME/.decodes.auth";   in method dbPasswordButtonPressed()


Note 2:

On windows, in some cases, you must specify UserAuthFile: (provider syntax with colon) to overcome windows file prefix such as: C:  

```text
DbAuthFile=UserAuthFile:$HOME/.opendcs/.decodes.auth
```

ktarbet:
another note:

$DECODES_INSTALL_DIR/hubhome.txt    -- usage in SyncGui.java

agilmore2:
[...]

PR #483, opened 2024-02-15
Bugfix/launcher default profile

MikeNeilson:
## Problem Description

The launcher wasn't opening the Setup properties editor correctly in the case of no, or the default, profile. Additionally discovered that if you launch the launcher with a profile it was only partially used as the default profile.

## Solution

Altered the behavior of launcher, cmdline args, and DecodesSetupFrame to take a profile instead of guessing what it should be.
Additionally (partially) centralized the logic for loading the DecodesSettings information from the profiles.

LauncherFrame and DecodesSetupFrame now require a valid Profile as an input parameter. The LauncherProfile stores the commandline profiled Profile as the "default" for the purposes for determining how to open the given applications.

## how you tested the change

Manually.
1. Open launcher (no profile set) and click Setup, default (user or decodes.properties) profile opens correctly
2. Open launcher (with profile set) and click Setup, selected profile opens correctly
3. Open launcher (with profile set), select (default) in the combo, selected profile opens correctly.

## Where the following done:


src/main/java/decodes/launcher/DecodesPropsPanel.java
```diff
@@ -217,44 +169,46 @@ public void loadFromSettings(DecodesSettings settings)
...
+            typ == DecodesSettings.DB_NWIS ? 2 :
+            typ == DecodesSettings.DB_CWMS ? 3 :
+            typ == DecodesSettings.DB_OPENTSDB ? 4 :
             typ == DecodesSettings.DB_HDB ? 5 : 0);
         editDbLocationField.setText(settings.editDatabaseLocation);
     }
 
-    public void saveToSettings(DecodesSettings settings)
+    public DecodesSettings saveToSettings()
     {
```
ktarbet, line 192:
```suggestion
    DecodesSettings saveToSettings()
```

src/main/java/decodes/launcher/LauncherFrame.java
```diff
@@ -2166,13 +2052,13 @@ protected void startProfileManager()
      */
     public Profile getSelectedProfile()
     {
-        return profileCombo == null ? null : (Profile)profileCombo.getSelectedItem();
+        return profileCombo == null ? launchProfile : (Profile)profileCombo.getSelectedItem();
     }
 
     private boolean trySendToProfileLauncher(String cmd)
     {
         profile = getSelectedProfile();
```
ktarbet, line 2060: profile here is null with a fresh XML database install.

The following error can be found with fresh XML install after clicking : Setup, or 'Decodes Database Editor' from the start_launcher main screen (without using profiles)

```text
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
        at decodes.launcher.LauncherFrame.trySendToProfileLauncher(LauncherFrame.java:2061)
        at decodes.launcher.LauncherFrame.setupPressed(LauncherFrame.java:962)
        at decodes.launcher.LauncherFrame$26.actionPerformed(LauncherFrame.java:677)
```

ktarbet (changes requested):

src/main/java/decodes/launcher/LauncherFrame.java
MikeNeilson: Yep, the else clause in the profileCombo handling left the combo empty. I changed it to add the "launchProfile" item and set it as selected. Seems to behave now.

ktarbet (approved):

MikeNeilson:
Given https://github.com/opendcs/opendcs/issues/476#issuecomment-1949114933 I'm going to wait to merge this in until I fix that particular behavior as well.

Partially a similar fix, needs to just use the profile as-provided, but also check that DbAuth *is* UserAuthFile as the others aren't supposed to be writable from the GUI. Might be worth adding a function to the AuthSourceProvider interface for checking writable and writing.

MikeNeilson:
@agilmore2 

So I think the right approach to #476 is to just make a new izpack panel and prompt the user during install. Going through the docs that's going to involve creating a custom izpack panel. I don't think that's a huge deal but I don't think it makes sense
to do that in izpack 4.

I don't suggest inplace updates, but always using a fresh directory, so I feel like migrating to izpack5 is the correct to do now. 
But like a separate PR, this one is already a bit big and is normalizing at least some behavior between OSes.

MikeNeilson:
I'm going to do the installer work as part of another PR, this at least improves operations once things have been configured.

PR #488, opened 2024-02-21
windows support for spaces in install dir.

### LINKED ISSUE #79: Space in install directory causes issues
**Describe the bug**
At least on windows if spaces are put in the installation path name the programs won't start and the message isn't super clear

**To Reproduce**
1. download opendcs-ot-<version>.jar and run it
2. set an installation path with a space in it
3. finish install, attempt run launcher_start.bat
4. See error (only if run from command line)

**Expected behavior**
We either shouldn't allow spaces and warn the user during install or the batch and shell files should be robust enough to handle it.

**Desktop**
 - OS: Windows
 - Version 6.8.14 and future and previous versions

**Additional context**
User was setting up new version and wasn't getting any error message but even launcher_start failed to open.

ktarbet:
This replaces most instances of $INSTALL_PATH with non-hardcoded alternative for .bat file.

## Problem Description

Fixes #79

## Solution

In windows using double quotes around paths/variables that may have spaces.

We take advantage of the fact that batch files in ./bin reference decj.bat which is in the same directory as all batch files; relative paths can be used and eliminate need to hard-code paths.  

## how you tested the change

Installed opendcs in directories with spaces,  such as ```c:\test\test 5\opendcs ```  and verify batch files work from within bin directory or from other directories.


install/bin/alarmdelete.bat
```diff
@@ -1 +1 @@
-@$INSTALL_PATH\bin\decj decodes.tsdb.alarm.AlarmDelete %*%
+@"%~dp0\decj" decodes.tsdb.alarm.AlarmDelete %*%
```
MikeNeilson, line 1: I'm not familiar with `%~dp0` is that just part of the windows shell or something that izpack inserts/replaces?

ktarbet: That is built into windows batch files.

ktarbet: I would like to do something similar to the Linux scripts such as:

```
SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```

https://stackoverflow.com/questions/59895/how-do-i-get-the-directory-where-a-bash-script-is-located-from-within-the-script

ktarbet: The installer builder, izpack, is using the opendcs-install.xml to decide what files to install.  

However, for batch files (.bat - windows) every batch file is installed regardless off what components are selected during the install.
This is because of this part of the config:

```xml
	<fileset dir="bin" targetdir="$INSTALL_PATH/bin" defaultexcludes="yes" 
				includes="*.exe,*.sql,*.bat,opendcs.jar" os="windows">
	</fileset>
```			

If this super-simple setup desired, over a hundred lines can be deleted in the rest of the opendcs-install.xml since they have no effect -- especially since I'm updating the batch files so they don't need any parsing during the install process.

MikeNeilson: We should probably go in the other direction and make sure those are properly part of a pack.

ktarbet (commented):
suggestion for future updates:

* izpack doesn't need to parse the batch files

install/bin/decj.bat
```diff
@@ -2,17 +2,22 @@
...
+ for /R "%CP_SHARED_JAR_DIR%" %%a in (*.jar) do (
+   set "CLASSPATH=!CLASSPATH!;%%a"
  )
 )
 
-set CLASSPATH="$INSTALL_PATH/bin/opendcs.jar;$INSTALL_PATH/bin/hibernate.cfg.xml;
-for /R $INSTALL_PATH/dep %%a in (*.jar) do (
-  set CLASSPATH=!CLASSPATH!;%%a
-)
+set "CLASSPATH=%BIN_PATH%opendcs.jar;%BIN_PATH%hibernate.cfg.xml;"
```
MikeNeilson, line 16: It looks like we're already setting the classpath on lines 10-13 and then here we completely ignore that and start the classpath fresh.

Or at least we could if that CP_SHARED_JAR_DIR is set.

@agilmore2 is that variable (CP_SHARED_JAR_DIR) used by any HDB users?

MikeNeilson (commented):
My only concern is the duplicate classpath processing. I suspect adding !CLASSPATH! to the second instance would fix it, but given that it's always been that way be may be able to remove that first loop.

Or I'm misunderstanding how batch works again.

install/bin/decj.bat
ktarbet: Yes, that is a issue.  I didn't test with CP_SHARED_JAR_DIR.

MikeNeilson (approved):

ktarbet:
I tested with install in a directory with spaces:
```
C:\test\install 42\OPENDCS\bin
```

And then moved opendcs.jar to  "C:\tmp\path with spaces\opendcs.jar"

and 
```
set CP_SHARED_JAR_DIR=C:\tmp\path with spaces
```

PR #489, opened 2024-02-21
Initial stab at all databases deriving from Database.

MikeNeilson:
## Problem Description

We technically have two database implementations within OpenDCS. One for "DECODES only" and another for Timeseries databases. It would greatly simplify the code to have a single implementation who's instance can be passed around when needed. (Like the constructors of the GUI frames).

## Solution

Taking notes from  https://github.com/opendcs/opendcs/wiki/Project-%E2%80%90-Merging-the-DECODES-Database-and-TimeseriesDb-classes this is an initial start on merging the systems into a single database implementation that has "requisite capabilities"

This starts that process. I'd like to get a few items flushed out and working before actually merging in like the getDao function
and also perhaps an improvement to how makeDatabaseIO is called.

The general goal is to turn `class Database` into `interface Database` and everything derives from that, such that the XML implementation could just be XmlDatabase.

## how you tested the change

At this point just running the tests we have to see if anything breaks.

## Where the following done:

### DIFF
--- a/src/main/java/decodes/db/Database.java
+++ b/src/main/java/decodes/db/Database.java
@@ -276,4 +276,10 @@ public class Database extends DatabaseObject
 		}
 		return presentationGroupList;
 	}
+
+
+	public <DaoType> DaoType getDao(Class<? extends DaoType> daoClazz)
+	{
+		return null;
+	}
 }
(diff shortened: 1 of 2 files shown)


ktarbet:
I don't see any notes in the wiki link provide above.

MikeNeilson:
Page title had a period at the end that didn't play nice with the automatic linking system. Should be fixed now.

MikeNeilson:
whelp, this initial excursion now fails with the most recent changes. But I think I've thought of another path forward.

PR #491, opened 2024-02-21
Give error for non-time-series editDatabaseType

ktarbet:
Avoids Null Pointer Exception when running alarmedit.bat  (AlarmEditor.java) with a xml database.  This will help other programs/batch-files give better error messages.

### DIFF
--- a/src/main/java/decodes/tsdb/TsdbAppTemplate.java
+++ b/src/main/java/decodes/tsdb/TsdbAppTemplate.java
@@ -261,7 +261,8 @@ public abstract class TsdbAppTemplate
 		throws ClassNotFoundException,
 		InstantiationException, IllegalAccessException
 	{
-		if (theDb != null) {
+		if (theDb != null)
+		{
 			return;
 		}
 
--- a/src/main/java/decodes/tsdb/TsdbAppTemplate.java
+++ b/src/main/java/decodes/tsdb/TsdbAppTemplate.java
@@ -1,80 +1,5 @@
-/*
-*  $Id: TsdbAppTemplate.java,v 1.18 2020/05/07 13:53:03 mmaloney Exp $
-*  
-*  Open Source Software 
-*  
-*  $Log: TsdbAppTemplate.java,v $
-*  Revision 1.18  2020/05/07 13:53:03  mmaloney
-*  app name defaults to "utility"
-*
-*  Revision 1.17  2019/12/11 14:32:44  mmaloney
-*  Allow for noInit to facilitate launcher.
-*
-*  Revision 1.16  2019/11/13 15:14:31  mmaloney
-*  Rm unneeded imports.
-*
-*  Revision 1.15  2019/09/12 12:56:05  mmaloney
-*  Added getCmdLineArgs
-*
-*  Revision 1.14  2018/03/30 14:13:32  mmaloney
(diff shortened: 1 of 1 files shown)


src/main/java/decodes/tsdb/TsdbAppTemplate.java
```diff
@@ -105,511 +33,518 @@
...
+            return;
+
+//        String className = DecodesSettings.instance().dbClassName;
+        String className = DecodesSettings.instance().getTsdbClassName();
+
+        if (className == null)
+        {
+            String message = "Error: can't create time series database with editDatabaseType= '"
+                    + DecodesSettings.instance().editDatabaseType + "'";
+            Logger.instance().fatal(message);
```
MikeNeilson, line 277: We should change this to slf4j and using the string formatting features.

MikeNeilson (changes requested):

src/main/java/decodes/tsdb/TsdbAppTemplate.java
ktarbet: If we construct a message (with concatenation) for purpose of a exception constructor message **and** logging, like the example above  -- Is there still  benefit to string formatting?

Other cases in this file are more clear cut and the slf4j {} formatting is clearly helpful.

MikeNeilson: If I'm doing both I'll usually use String.format. But I've run into this problem myself, should it even do both? Or just throw the exception?

src/main/java/decodes/tsdb/TsdbAppTemplate.java
```diff
@@ -384,24 +369,14 @@ public void tryConnect()
...
+     * @param afn auth filename
+     * @param ex exception
      */
     protected void authFileEx(String afn, Exception ex)
     {
-        String msg = "Cannot read DB auth from file '" + afn + "': " + ex;
-        System.err.println(msg);
-        Logger.instance().failure(msg);
-try { throw new Exception(""); } catch (Exception ex2) { ex2.printStackTrace(); }
+        log.error("Cannot read DB auth from file '{}'", afn ,ex);
```
MikeNeilson, line 376: I believe this form doesn't work and the .atError().setCause(ex).log(...) needs to be used.
But easy to test with a bad auth file.

MikeNeilson (commented):
Looking good. Thanks for adjusting the whole file.

src/main/java/decodes/tsdb/TsdbAppTemplate.java
ktarbet: Thanks, I'm trying String.format

MikeNeilson (approved):

src/main/java/decodes/tsdb/TsdbAppTemplate.java
```diff
@@ -340,42 +261,40 @@ public synchronized void createDatabase()
 		throws ClassNotFoundException,
 		InstantiationException, IllegalAccessException
 	{
-		if (theDb != null)
+		if (theDb != null) {
```
MikeNeilson, line 264: { on next line.

MikeNeilson (changes requested):
One nitpicky bracket formatting issue.

MikeNeilson (approved):

PR #490, opened 2024-02-21
Refactor/324 cwms timeseries dao

MikeNeilson:
## Problem Description

Fixes #324 . 
Fixes some minor issues and issues with recent PRs that were merged in.

## Solution

- Slogged through to use bind variables.
- Removed general caching. list all was still allowed to as it was getting everything anyways.
- Added batch modify operations
- Implemented a getFirstResult vs getSingleResult (single result throws on more than one)
- Added a very simple log filter feature (CWMS JOOQ log messages were all happening at info making the log impossible to read.)

## how you tested the change

Existing and new tests. Running on local machine, running on HEC Test machine against real-world data.

## Where the following done:

- [ ] Tests. Check all that apply:
   - [ ] Unit tests created or modified that run during ant test.
   - [x] Integration tests created or modified that run during integration testing
         (Formerly called regression tests.)
   - [ ] Test procedure descriptions for manual testing
- [x] Was relevant documentation updated?
- [ ] Were relevant config element (e.g. XML data) updated as appropriate

If you aren't sure leave unchecked and we will help guide you to want needs changing where.

PR #497, opened 2024-02-26
Refactor/tasklist operation design

MikeNeilson:
## Problem Description

Fixes #. 

During #490 I discovered it was difficult to write a generic test for tasklist operations.
While we have the TaskListRecord object it must have logic for every possible implementation.

This begins the process of making a more generic interface. The Goal, beyond just making tests easier, is to also allow additional concepts, like using "select for update" and allowing automatic horizontal scaling at some point in the future.

## Solution

Created a DAO for Tasklist and created abstract functions for baseline operations.
Created an interface for tasklist entries that maps the concepts that apply to all implementations (e.g. what CompProc itself requires.)

ts_code vs site_datatype_id + tabsel, etc has been mapped to TimeSeriesIdentifier leaving such logic to the implementation.

## how you tested the change

In-progress, will be updating the now disabled CompProcIT test for the tasklist, and will add additional operations.
Additionally since each implementation will be using this the exiting computation system test will make use of it.

## Where the following done:


MikeNeilson:
@opendcs/opendcs-core-devs I started this this morning and realized it was going to take at least the rest of the week (and I need to pivot to other work for a day or two.) But I think this is a decent start to get feedback on the ideas.

Eventually this would be used in the existing getNewData function that builds the DataCollection.

src/main/java/opendcs/dao/TaskListDao.java
```diff
@@ -0,0 +1,117 @@
...
+    public TaskListDao(DatabaseConnectionOwner dco, String module)
+    {
+        this(dco, module, null);
+    }
+
+    public TaskListDao(DatabaseConnectionOwner dco, String module, Connection con)
+    {
+        super(dco, module, con);
+        curTimeSqlCommand = dco.isOracle() ? "sysdate" : "current_timestamp" ;
+        intervalSqlCommand = dco.isOracle() ? " ?/24 )" : " ? * INTERVAL '1' hour )" ; // Oracle date math
```
ktarbet, line 36: regarding DatabaseConnectionOwner.java 
 
As the code continues to refactor, Hopefully methods like : sOracle(), isHdb(), isCwms(), isOpenTSDB() could be eliminated, and instead each implementation can answer more specific questions, as needed, such as   getCurrentTimeStamp() .     

Another note: The name 'DatabaseConnectionOwner' feels like it should have the Connection if it is the owner, so why do we need a separate Connection argument.

MikeNeilson: Yeah, that's the goal. At the moment I'm more focus on at least centralizing those things so that's easier to actually do.

As for DatabaseConnectionOwner, that should really be called DatabaseConnectionProvider but I'm fairly certain long term it's going to go away in favor of just "Database" so I'm not sure it's worth the effort to rename.

src/main/java/opendcs/dai/TaskListDAI.java
```diff
@@ -0,0 +1,47 @@
...
+     * Retrieve a set of entries encoded into appropriate TaskListEntry objects
+     *
+     * @param appId Current Design: Loading application that we want data for.
+     *              Future Design: Marker for parallel instances to retrieve arbitrary data? (more investigation is required.)
+     * @param amount How much data to retrieve to process this batch.
+     * @param includeFailed Whether data with fail_time set should be retrieved.
+     * @return A list of TaskListEntries that can be further processed.
+     * @throws DbIoException If anything goes wrong during the retrieval.
+     */
+    public List<? extends TaskListEntry> getEntriesFor(DbKey appId, int amount, boolean includeFailed) throws DbIoException;
```
ktarbet, line 24: Consider renaming amount to  'numberOfEntries' if that is what it is?

MikeNeilson: hmm, that works. it's basically the batch size of tasks to process.

src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
```diff
@@ -1489,7 +1489,8 @@ else if (!TextUtil.strEqualIgnoreCase(recUnitsAbbr, ctsUnitsAbbr))
         }
         if (rrhandle != null)
         {
-            rrhandle.addRecNum(rec.getRecordNum());
+            // TODO: implement new Scheme
+            //rrhandle.addRecNum(rec.getRecordNum());
```
ktarbet, line 1493:
```suggestion
            //rrhandle.addRecNum(rec.getRecordNum());
            throw new UnsupportedOperationException("rrhandle.addRecNum(rec.getRecordNum())");
```

src/main/java/decodes/tsdb/RecordRangeHandle.java
```diff
@@ -20,92 +20,113 @@
...
-	public void addRecNum(int recordNumber)
+	public void addRecNum(TaskListEntry entry)
 	{
-		recnums.add(recordNumber);
+		recNums.add(entry);
 	}
 
 	/**
 	 * Returns a string containing comma-separated integer record numbers.
 	 * The returned integers are also removed from the list.
```
ktarbet, line 60:
```suggestion
	 * removes and returns a list of TaskListEntry 
```
consider calling this method:  RemoveEntries(max)

src/main/java/decodes/tsdb/RecordRangeHandle.java
```diff
@@ -20,92 +20,113 @@
...
-	 * @param max the maximum number of integers to return.
+	 * @param max the maximum number of integers to return. (ignored)
```
ktarbet, line 61:
```suggestion
	 * @param max the maximum number of TaskListEntry to remove and return.
```

src/main/java/decodes/tsdb/RecordRangeHandle.java
```diff
@@ -20,92 +20,113 @@
...
 		}
-		return sb.toString();
+		return entries;
 	}
 
 	/**
 	 * Returns a string containing comma-separated integer failed rec-numbers.
 	 * The returned integers are also removed from the list.
 	 * @param max the maximum number of integers to return.
 	 */
```
ktarbet, line 90:
```suggestion
	 * Removes and returns a list of FailedTaskListEntry
	 * @param max the maximum number of items to remove and return.
	 */
```

src/main/java/decodes/tsdb/RecordRangeHandle.java
```diff
@@ -20,92 +20,113 @@
...
-	public String getFailedRecNumList(int max)
+	public List<FailedTaskListEntry> getFailedRecNumList(long max)
```
ktarbet, line 91: consider naming this method:   removeFailedEntries(...)

ktarbet (commented):
Great demonstration of formalizing the task list!

side note:
'OpenTimeSeries' and 'OpenHydro' seemed to be two names for the same thing?  This can be confusing.

src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
MikeNeilson: I should have been clearer in the message, these need to be replace *before* this gets merged in. The entire computation process system fails without these records. I just wanted to allow for initial review before adding that much code.

src/main/java/decodes/tsdb/RecordRangeHandle.java
MikeNeilson: That's a good point on the naming, I wouldn't expected removal on the current name alone.

src/main/java/decodes/tsdb/RecordRangeHandle.java
MikeNeilson: good idea.

MikeNeilson:
> Side note:
> 'OpenTimeSeries' and 'OpenHydro' seemed to be two names for the same thing? This can be confusing

Yes, I mostly wanted to try out the new name. I like it, and it's easier to read in the code for me for whatever reason. I'll rename the new classes to OpenTsdb before this is merged in and do the actual rename in a separate PR.

PR #499, opened 2024-02-28
updating-legacy-doc-images

zamusacelre:
- adding to routing spec
- start on screening
- lots of renaming of legacy images - cleaned out if image not used, and fixed a couple


docs/source/legacy-dev-docs.rst
```diff
@@ -151,13 +151,11 @@ system you can turn pool tracing on for an application with the following java f
 
 .. code-block:: bash
 
-    DECJ_MAXHEAP="-Dopendcs.connection.pool.trace=true" routsched ...
```
MikeNeilson, line 154: this should be opendcs not cwms. I need to sanity check but this should've been renamed as part of a branch that was recently merged it.

(also I move the dev docs out of the legacy mark as they aren't, you'll want to rebase onto the most recent master branch)

docs/source/legacy-dev-docs.rst
```diff
@@ -213,103 +211,6 @@ To acquire the configured credentials the following can be used:
     }
     ...
 
-PropertyProvider
```
MikeNeilson, line 216: This shouldn't be removed, as above you'll just need to pull the latest changes from master in.

docs/source/start-decoding.rst
```diff
@@ -0,0 +1,38 @@
...
+A major bulk of the examples here are likely not typically
+full raw messages from sources.  The sample messages are 
+displayed for the purposes of providing examples of how 
+the format statements function and decoding information.
+
+Further details and advanced topics on the DECODING can be 
+found in the DECODING manual ::ref DECODES Guide <./legacy-decoding-guide.rst>
+
+
+How do I launch the Screening Editor?
```
MikeNeilson, line 19: This appears to be in the wrong file.

MikeNeilson (changes requested):
Looking good. Needs to keep/pickup some recent changes that are in the master branch.

docs/source/dev-docs.rst
```diff
@@ -151,13 +151,11 @@ system you can turn pool tracing on for an application with the following java f
 
 .. code-block:: bash
 
-    DECJ_MAXHEAP="-Dopendcs.connection.pool.trace=true" routsched ...
+    DECJ_MAXHEAP="-Dcwms.connection.pool.trace=true" routsched ...
```
zamusacelre, line 154:
```suggestion
    DECJ_MAXHEAP="-Dopendcs.connection.pool.trace=true" routsched ...
```

docs/source/legacy-dev-docs.rst
zamusacelre: This is I believe the one example of where changed/edits were made on the master. I rebased. However when I try including the content in PropertyProvider, I get an warning/error on the build.  Unsure how to proceed on this one.

MikeNeilson (approved):

MikeNeilson (approved):

PR #501, opened 2024-02-29
Improvements to handling operations within a transaction.

MikeNeilson:
## Problem Description

Came up as part of #497 

There are certain delete operations that work on multiple tables. These queries are independent (often between different DAO objects) and auto commit is usually turned on by default (just in generally by JDBC.) This means that users can be left with a database in an inconsistent and sometimes difficult to correct state.

## Solution

Make it easier for the DAO classes to start and use a transanction and make sure each other DAO used is using the same connection and thus the same transaction.

## how you tested the change

- CompDepends uses transaction to manipulate the tasklist tables with a sequence of operations.
- [x] test for delete timeseries which at least in the OpenDCS Postgres implementation calls another DAO and operates directly on a few different tables.

## Where the following done:


src/main/java/opendcs/dao/DaoBase.java
```diff
@@ -99,6 +100,20 @@ public DaoBase(DatabaseConnectionOwner tsdb, String module, Connection con)
...
+     * @param other
+     * @throws IllegalStateException if the other Dao does not already have a connection open this operation is not valid.
+     */
+    public void  inTransactionOf(DaoBase other) throws IllegalStateException
+    {
+        if (other.myCon == null)
+        {
+            throw new IllegalStateException("Provided DAO does not currently have a valid connection that would be in a transaction.");
+        }
+        this.setManualConnection(other.myCon);
```
ktarbet, line 114: When this/DaoBase assumes connection in another Dao, by calling setManualConnection(other.myCon).

Should there be cleanup when finished, to get back to original connection?

src/main/java/opendcs/opentsdb/OpenTimeSeriesDAO.java
```diff
@@ -830,72 +830,90 @@ public void deleteTimeSeriesRange(CTimeSeries ts, Date from, Date until)
...
+				}
+				catch (Exception ex)
+				{
+					throw new Exception("Error deleting ts_spec entry", ex);
+				}
+
+				try
+				{
+					String q = "select num_ts_present, est_annual_values from storage_table_list "
+					+ "where table_num = ?" + ctsid.getStorageTable();
```
ktarbet, line 887:
```suggestion
					String q = "select num_ts_present, est_annual_values from storage_table_list "
					+ "where table_num = ?";
```
Doesn't seem quite right, but not sure if this is the proper fix.

MikeNeilson: You're suggested change is correct, I missed actually deleting that particular string append.

src/main/java/opendcs/dao/DaoBase.java
MikeNeilson: I'm going to have to think about that, The way I've been using them it doesn't matter as they only live during the transaction block itself. But there could be a reasonable way to allow that without any specific work on the end user.

MikeNeilson:
I'm going to hold of on further changes to this one until #505 is merged in. There's changes there to make reading in time series easier and this needs a few tests against actual, reasonably larger time series.

src/main/java/opendcs/dao/DaoBase.java
MikeNeilson: There's not really a good way to handle this, I put a not in java doc that the Dao should be considered invalid after calling this.

src/main/java/opendcs/opentsdb/OpenTimeSeriesDAO.java
```diff
@@ -410,50 +447,68 @@ public int fillTimeSeries(CTimeSeries ts, Date from, Date until,
...
 					EngineeringUnit.getEngineeringUnit(ts.getUnitsAbbr()));
+		}
 
 		String tableName = makeDataTableName(ctsid);
-		String q = "select " + ts_columns + " from " + tableName
-			+ " where ts_id = " + ctsid.getKey();
+		List<Object> parameters = new ArrayList<>();
+		StringBuilder q = new StringBuilder();
+		q.append("select " + ts_columns + " from " + tableName
+			+ " where ts_id = ?");// + ctsid.getKey();
```
ktarbet, line 469:
```suggestion
			+ " where ts_id = ?");
```

MikeNeilson: oops. good catch.

PR #520, opened 2024-03-15
Refactor/auth and datasource before database object creation

### LINKED ISSUE #221: SPI for Database Implementation
Database Instantiation

**Purpose of refactoring**
Reduce manual curation of Database system implementations. Allows additional implementations in the future.
DecodesSettings.java doesn't need to know about every possible database type, just how to search for them.
Class name not required in configuration files

**Proposed method**
Use the Java ServiceProviderInterface mechanism to search for and create instances of required database connections.
Would also be used by "CwmsConsumer" to pull a valid instance.

**Additional context**
I've already done this in a previous branch. It's not a huge change; config files already have an "edit-db" type field and we can reuse those names.

MikeNeilson:
## Problem Description

Fixes #221 . 

Merges Database concepts into a single "source of truth" for a connected instance.

The distinction between the Decodes and Timeseries datbases, and the history of the XML implementation make many operations rather obtuse. Additionally work generally needs to be duplicated to account for various differences.

This will make future expansion rather difficult.

## Solution

1. DatabaseIO and TimeSeriesDb now all require a javax.sql.DataSource and DecodesSettings
2. a JDBC driver was implemented just enough to handle passing the folder location.
3. TimeSeriesDb now derives from Database
4. TimeSeriesDb implementations are responsible for setting up their Decodes DbIo, and now we can slowly create Dao objects to replace that behavior
5. References to Database.getDb() within the DAOs can be removed (may need to add function to DatabaseConnectionOwner for DAOs to access, or perhaps new interface), additionally starts that path with the settings as well.
6. There is an SPI system for database creation that handles creating the datasource and using the settings to get auth if required.

TODO:
- [] XML, CWMS and HDB SPI for Database
- [] More tests
- [] determination of exactly when the Database::instance can be removed.
- [] A proper data source, such as HakiriCP, to handle pooling
- [] remove CwmsConnectionPool

## how you tested the change

Integrations, still much work to be done.

## Where the following done:


src/main/java/decodes/cwms/CwmsSqlDatabaseIO.java
```diff
@@ -49,9 +48,7 @@
  * USACE (U.S. Army Corps of Engineers) CWMS (Corps Water Management System)
  * database, which is hosted on an Oracle DBMS.<p>
  */
-public class CwmsSqlDatabaseIO
-	extends SqlDatabaseIO
-	implements DatabaseConnectionOwner
+public class CwmsSqlDatabaseIO extends SqlDatabaseIO
 {
 	public final static String module = "CwmsSqlDatabaseIO";
 	/** The office ID associated with this connection. This implicitely
```
ktarbet, line 54:
```suggestion
	/** The office ID associated with this connection. This implicitly
```

src/main/java/decodes/cwms/CwmsSqlDatabaseIO.java
```diff
@@ -70,20 +67,15 @@ public class CwmsSqlDatabaseIO
 	* where hostname and dbname specify the Oracle CWMS database.
 	* @param sqlDbLocation the location string from decodes.properties file
```
ktarbet, line 68: suggested javadoc:

```
	/**
 	* Constructor for CwmsSqlDatabaseIO
	 * @param dataSource the DataSource for connecting to the database
	 * @param settings the DecodesSettings
	 * @throws DatabaseException if there is an issue with the database connection
 	*/
```

src/main/java/decodes/cwms/CwmsTimeSeriesDb.java
```diff
@@ -765,9 +67,9 @@ public class CwmsTimeSeriesDb
 	 * No args constructor required because this is instantiated from
 	 * the class name.
 	 */
-	public CwmsTimeSeriesDb()
+	public CwmsTimeSeriesDb(String appName, javax.sql.DataSource dataSource, DecodesSettings settings) throws DatabaseException
```
ktarbet, line 70: suggested javadoc:

```java
/**
 * Constructs a CwmsTimeSeriesDb object with the specified parameters.
 *
 * @param appName the name of the application using the database
 * @param dataSource the javax.sql.DataSource object for the database connection
 * @param settings the DecodesSettings object containing database settings
 * @throws DatabaseException if there is an issue with the database connection
 */

File: src/main/java/decodes/db/DatabaseIO.java [hunk 4]
```diff
@@ -37,47 +48,45 @@ public abstract class DatabaseIO
...
 		
-		try 
+		try
 		{
-			if (type == DecodesSettings.DB_XML)
-				return new XmlDatabaseIO(location);
+			final String location = locOverride.startsWith("jdbc:") ? locOverride : "jdbc:xml:" + locOverride;
+			final int type = settings.editDatabaseTypeCode;
+
+			AuthSource auth = AuthSourceService.getFromString(settings.DbAuthFile);
```
ktarbet, line 70: We may neeed a AuthSource for the XML database?   Testing with initial install gives:

```
C:\test\OPENDCS-auth\bin>dbedit
Exception in thread "main" java.lang.NullPointerException
        at org.opendcs.authentication.AuthSourceService.getFromString(AuthSourceService.java:35)
        at org.opendcs.database.DatabaseService.getDatabaseFor(DatabaseService.java:24)
        at decodes.dbeditor.DecodesDbEditor.main(DecodesDbEditor.java:132)
```

MikeNeilson: That would be the NoOp datasource that you can see in the tests. Though it should've tried to default to that.

src/main/java/decodes/db/Netlist2Lrgs.java
```diff
@@ -67,12 +67,11 @@ public static void main(String args[])
 		String dbloc = dbLocArg.getValue();
 		if (dbloc.length() > 0)
 		{
-			dbio = DatabaseIO.makeDatabaseIO(settings.DB_XML, dbloc);
+			dbio = DatabaseIO.makeDatabaseIO(settings, dbloc);
```
ktarbet, line 70: also getting error here:

```
:\test\OPENDCS-auth\bin>nl2lrgs
Exception in thread "main" java.lang.NullPointerException
        at org.opendcs.authentication.AuthSourceService.getFromString(AuthSourceService.java:35)
        at decodes.db.DatabaseIO.makeDatabaseIO(DatabaseIO.java:70)
        at decodes.db.DatabaseIO.makeDatabaseIO(DatabaseIO.java:58)
        at decodes.dbimport.LrgsNetlist.main(LrgsNetlist.java:70)
```

MikeNeilson:
So, at least for the Postgres reference implementation, the routing scheduler is now stable.
There are issues in this branch for CWMS I was working on before vacation, but that's at least motivation to keep going and fix it.

MikeNeilson:
So some good news/bad news again.

Good news: The tests pass for CWMS.
The Bad news: Everything runs a *lot* slower if you have a large database as moving the "Decodes Database"::read into the initial creation causes *every* decodes element to be read in at application start.

We needed to start moving everything towards the DAO style anyways so I don't think this is a total disaster, but it does mean this is going to take some time.

src/main/java/decodes/cwms/CwmsDatabaseProvider.java
```diff
@@ -0,0 +1,46 @@
...
+import decodes.util.DecodesSettings;
+import usace.cwms.db.dao.util.connection.ConnectionLoginInfoImpl;
+
+public class CwmsDatabaseProvider implements DatabaseProvider
+{
+
+    @Override
+    public boolean canCreate(DecodesSettings settings)
+    {
+        return "CWMS".equals(settings.editDatabaseType);
```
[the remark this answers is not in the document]
MikeNeilson: Actually that should be a constant within this Provider. Putting it in "Contants" implies that the core system *should* have knowledge of CWMS, it shouldn't.

PR #522, opened 2024-03-19
Copy Parity error in copy constructor.

### LINKED ISSUE #401: RoutingSpec Messages with parity error (?) are not deleted
### Discussed in https://github.com/opendcs/opendcs/discussions/396

<div type='discussions-op-text'>

<sup>Originally posted by **berrio-23** November 23, 2023</sup>
Hello Community

Currently, we have interference in GOES transmissions and some messages have a parity error, code ?; In the opendcs 7.0.8 version, in the decodes module and routing spec, I have changed the parity option with the same result: It does not eliminate the message received with a parity error. Do you have any suggestions? Because despite the thresholds used for the variables, garbage values are passed into the database as a result of interference. It looks like a bug on oppendcs 7.0.8 version. Please, see attachment.
[Messages with parity error.docx](https://github.com/opendcs/opendcs/files/13451587/Messages.with.parity.error.docx)
PD. Opendcs 7.0.6 has the same bug.
Thanks for your comments,
Gustavo</div>

MikeNeilson:
## Problem Description

Fixes #401. 

## Solution

Required property was not getting copied and was thus never send to the server to filter data as requested by the user.
Value is now properly assigned in copy constructor.

## how you tested the change

Manually. Set routing spec to only required "bad" parity message and received no messages.
Logic flow for "only" good is the same and the particular logic block check appears reasonable.

## Where the following done:


src/test/java/lrgs/common/SearchCriteriaTest.java
```diff
@@ -0,0 +1,66 @@
...
+    public void test_time_evaluation() throws Exception
+    {
+        sc.setLrgsSince("now - 1 day");
+        sc.setLrgsUntil("now");
+        final Date since = sc.evaluateLrgsSinceTime();
+        final Date nowMinus1Day = new Date(System.currentTimeMillis()-86400*1000);
+        assertTrue(Math.abs(since.getTime() - nowMinus1Day.getTime()) < (3600*1000+2000), "since time not within reasonable deviation.");
+        final Date now = new Date();
+        final Date evalNow = sc.evaluateLrgsUntilTime();
+        assertTrue(Math.abs(now.getTime() - evalNow.getTime())< 2000, "until time not within reasonable deviation.");
```
ktarbet, line 64:
```suggestion
        int millisPerDay = 86400*1000;
        int millisPerHour = 3600*1000;
        int millisTolerance= 2000;
        final Date nowMinus1Day = new Date(System.currentTimeMillis()-millisPerDay);
        assertTrue(Math.abs(since.getTime() - nowMinus1Day.getTime()) < (millisPerHour+millisTolerance), "since time not within reasonable deviation.");
        final Date now = new Date();
        final Date evalNow = sc.evaluateLrgsUntilTime();
        assertTrue(Math.abs(now.getTime() - evalNow.getTime())< millisTolerance, "until time not within reasonable deviation.");

```

ktarbet (commented):
Looks good.

I have one suggested readability improvement in the test.

ktarbet (approved):

PR #524, opened 2024-03-19
Update PlatformListIO.java with Bind Variables for Performance Enhancement

### LINKED ISSUE #523: PlatformListIO should use bind variables

PlatformListiO.java needs refactoring for bind variables, primary for performance reasons.

**Proposed method**

Follow other examples in the code that has moved to bind variables.   SiteDAO for example, simplifies binding  work by using a method doQuery(...) , inherited from DaoBase.   

**Additional context**

PlatformListIO extends SqlDbObjIo, so I'll study this code. However, I welcome advice on how to approach my first bind-variable refactor in OpenDCS.

ktarbet:
## Problem Description

Fixes #523 

## how you tested the change

## Where the following done:

### DIFF
--- a/src/main/java/decodes/sql/PlatformListIO.java
+++ b/src/main/java/decodes/sql/PlatformListIO.java
@@ -2,6 +2,7 @@ package decodes.sql;
 
 import java.sql.Connection;
 import java.sql.ResultSet;
+import java.sql.SQLDataException;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
@@ -9,6 +10,7 @@ import java.util.Date;
 import java.util.Iterator;
 import java.util.Vector;
 
+import decodes.db.Database;
 import opendcs.dai.DacqEventDAI;
 import opendcs.dai.PlatformStatusDAI;
 import opendcs.dai.PropertiesDAI;
@@ -27,10 +29,12 @@ import decodes.db.Site;
 import decodes.tsdb.DbIoException;
 import decodes.util.DecodesSettings;
 
+import opendcs.dao.DaoHelper;
+import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * This handles the I/O of the PlatformList object and some of its
+ * PlatformListIO handles the I/O of the PlatformList object and some of its
  * kids to/from the SQL database.
  * When reading, this reads the Platform and TransportMedium tables.
  * Note, though, that only the MediumType and MediumId fields of the
@@ -38,7 +42,7 @@ import org.slf4j.LoggerFactory;
  */
 public class PlatformListIO extends SqlDbObjIo
 {
-    private static org.slf4j.Logger log = LoggerFactory.getLogger(PlatformListIO.class);
+    private final static Logger log = LoggerFactory.getLogger(PlatformListIO.class);
 
     /** Transient reference to the PlatformList that we're working on. */
     protected PlatformList _pList;
@@ -101,6 +105,12 @@ public class PlatformListIO extends SqlDbObjIo
 
         _configListIO = configListIO;
         _equipmentModelListIO = emlIO;
+    }
+
+    private DaoHelper getDaoHelper()
+    {
+        DaoHelper h = new DaoHelper(_dbio,"sql.platformListIO",connection());
+        return h;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/decodes/sql/PlatformListIO.java
```diff
@@ -421,129 +417,102 @@ private TransportMedium rs2tm(ResultSet rs, Platform platform)
...
+									+ p.getKey() + " has invalid siteID " + siteId);
+							p.setSite(null);
 						}
 					}
+				} else {
+					try {
+						siteDAO.readSite(p.getSite());
+					} catch (Exception e) {
+						// TODO Auto-generated catch block
+						e.printStackTrace();
```
MikeNeilson, line 458: This should log to the logger, not be printed to stdout.

Also, for this block, bracket placement.

src/main/java/decodes/sql/PlatformListIO.java
```diff
@@ -421,129 +417,102 @@ private TransportMedium rs2tm(ResultSet rs, Platform platform)
...
+					}
+				}
+			} finally {
+				siteDAO.close();
```
MikeNeilson, line 462: `SiteDAI` and `SiteDAO` implements `AutoClosable` you can migrate this to a try-with-resources.

src/main/java/decodes/sql/PlatformListIO.java
```diff
@@ -421,129 +417,102 @@ private TransportMedium rs2tm(ResultSet rs, Platform platform)
...
-						{
-							throw new DatabaseException(e.getMessage());
-						}
-						finally
-						{
-							propsDao.close();
-						}
-					}
-				}
+		if (getDatabaseVersion() >= DecodesDatabaseVersion.DECODES_DB_6) {
```
MikeNeilson, line 504: bracket styling.

src/main/java/opendcs/util/functional/ResultSetConsumer.java
```diff
@@ -13,5 +15,5 @@ public interface ResultSetConsumer extends ThrowingConsumer<ResultSet,SQLExcepti
     /**
      * @param rs a valid resultset, user function should *NOT* call next
      */
-    public void accept(ResultSet rs) throws SQLException;
+    public void accept(ResultSet rs) throws SQLException, DatabaseException;
```
MikeNeilson, line 18: Hmm, that does simplify some things, but not not entirely convinced that that's right approach... then again it might be the best approach until we get things better sorted out.

src/main/java/opendcs/util/functional/ThrowingConsumer.java
```diff
@@ -1,10 +1,12 @@
 package opendcs.util.functional;
 
+import decodes.db.DatabaseException;
+
 /**
  * Function interval to allow for checked exceptions.
  */
 @FunctionalInterface
 public interface ThrowingConsumer<T,E extends Exception>
 {
```
MikeNeilson, line 10: This is not correct here. This may be for other than just Database operations. Requiring everything to check for Database exception when there may be no database operations doesn't make sense.

While the situation certainly isn't perfect I've bee just catching the Non-SQL exceptions and rethrowing them with `new SQLException("some extra detail", ex)` where `ex` is that other exception.

MikeNeilson (changes requested):

src/main/java/decodes/sql/PlatformListIO.java
```diff
@@ -288,21 +233,20 @@ public void read(PlatformList platformList)
 	protected void readAllTransportMedia(PlatformList platformList)
 		throws SQLException, DatabaseException
 	{
-		try (Statement stmt = createStatement();)
+		try (Statement stmt = createStatement())
```
MikeNeilson, line 236: you can use doQuery here, it will loop through the result set list and call the provided function for each returned row.

MikeNeilson (commented):
Looks like it's tarting to come together.

MikeNeilson (changes requested):
Queries look good. Some minor formatting issues.

ktarbet:
do we have helpers for inserting data?

MikeNeilson:
Yes, they just aren't called "insert", there's `DaoBase::doModify` and `DaoBase::doModifyBatch`

Also the doQuery/getSingleResult are also suitable if you want to trying using "returning into" type statements on insert. But I don't think we'll be doing that for a while.

PR #530, opened 2024-03-21
Add ScalarAdder30.java

msweier:
## Problem Description
Limited ability to add more than 10 timeseries.

Fixes #. 
Add a java algo that allows 30 timeseries to be added.

## Solution

Describe your solution.

## how you tested the change

## Where the following done:

### DIFF
--- a/src/main/java/decodes/tsdb/algo/ScalerAdder30.java
+++ b/src/main/java/decodes/tsdb/algo/ScalerAdder30.java
@@ -64,7 +64,14 @@ public class ScalerAdder30
 	public double input28;	//AW:TYPECODE=i
 	public double input29;	//AW:TYPECODE=i
 	public double input30;	//AW:TYPECODE=i
-	String _inputNames[] = { "input1", "input2", "input3", "input4", "input5", "input6", "input7", "input8", "input9", "input10", "input11", "input12", "input13", "input14", "input15", "input16", "input17", "input18", "input19", "input20", "input21", "input22", "input23", "input24", "input25", "input26", "input27", "input28", "input29", "input30" };
+	String _inputNames[] = { 
+		"input1", "input2", "input3", "input4", "input5", 
+		"input6", "input7", "input8", "input9", "input10", 
+		"input11", "input12", "input13", "input14", "input15", 
+		"input16", "input17", "input18", "input19", "input20", 
+		"input21", "input22", "input23", "input24", "input25", 
+		"input26", "input27", "input28", "input29", "input30" 
+	};
 //AW:INPUTS_END
 
 //AW:LOCALVARS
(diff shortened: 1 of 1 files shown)


src/main/java/decodes/tsdb/algo/ScalerAdder30.java
```diff
@@ -0,0 +1,290 @@
...
+	public String input3_MISSING = "ignore";
+    public String input30_MISSING = "ignore";
+	public String input4_MISSING = "ignore";
+	public String input5_MISSING = "ignore";
+	public String input6_MISSING = "ignore";
+	public String input7_MISSING = "ignore";
+	public String input8_MISSING = "ignore";
+	public String input9_MISSING = "ignore";
+	String _propertyNames[] = { "coeff1", 
+    "coeff10", "coeff11", "coeff12", "coeff13", "coeff14", "coeff15", "coeff16", "coeff17", "coeff18", "coeff19", "coeff2",
```
MikeNeilson, line 141: trying grouping these by 5s. And starting the data on the next line

```java
String _propertyNames[] =
{ 
    "coeff1", "coeff2", "coeff3", "coeff4", "coeff5",
    "coeff6", "coeff7", "coeff8", "coeff9", "coeef10"
// and so on
}
```

This will make it easier to verify everything is presented and accounted for.

src/main/java/decodes/tsdb/algo/ScalerAdder30.java
```diff
@@ -0,0 +1,290 @@
...
+	 *
+	 * @throws DbCompException (or subclass thereof) if execution of this
+	 *        algorithm is to be aborted.
+	 */
+	protected void doAWTimeSlice()
+		throws DbCompException
+	{
+//AW:TIMESLICE
+		double tally = 0.0;
+		if (!isMissing(input1))
```
MikeNeilson, line 190: This comment is for our future selves not work to do right now... there has got to be a better way to handle this.... so much repeated logic, but using java reflection is probably too slow. Might be worth an investigation in the future

MikeNeilson (changes requested):

ktarbet:
Could this be made more generic, where we are not limited to 30, but a dynamic sizing?   Perhaps during some initializing, or '@Setup' method like Junit, where the number of items is declared at that point.

I'm new to OpenDCS algorithms, so I'm not sure what effort this would be.

MikeNeilson:
> Could this be made more generic, where we are not limited to 30, but a dynamic sizing? Perhaps during some initializing, or '@setup' method like Junit, where the number of items is declared at that point.
> 
> I'm new to OpenDCS algorithms, so I'm not sure what effort this would be.

I don't think so in the current design structure. But that desire is where my "must be a better way" comment came from.

MikeNeilson (approved):

Issue #526, opened 2024-03-20
CompRun, with large groups timesout before execution.


MikeNeilson:
**Describe the bug**
When manually running a computation that uses a large group, it 

**To Reproduce**
1. Create a large group (You'll need a lot of timeseries)
2. Create comp that uses said large group
3. Open CompEdit, go to comp and attempt to execute
4. Wait too long for list, execution may not happen

**Expected behavior**
Reasonable performance, Proper status indication, and at the least not actually failing.

**Screenshots**

**Desktop/ServerSide (please complete the following information):**

**Additional context**
This needs a combination of SwingWorker status notifications, and improving the TsGroupDAO queries.
There's also a really nasty loop that calls `TimeSeriesDb::transformTsidByCompParm` which in CWMS no uses bind vars but tight loops that make database calls like that aren't good.

PR #533, opened 2024-03-21
Bugfix/526 timeseries list building speed

MikeNeilson:
## Problem Description

Partial Fix of #526. 

## Solution

Allow the system to report the current state so user at least know where a long running time series expansion is in the process.
To do this I've rearranged the the runComp handler into 3 functions.
- check if request valid
- build group comp list (if required)
- run computations

The building of computations for groups is now in a SwingerWorker that will call yet another SwingWorker when the time comes to actually run the selected computations.

Addtionally, since the reason for this issue included a group that has 14000+ timeseries assosciated to it, I've altered the timeseries transformation slightly to at least report the current progress.

There will need to be slightly more major changes to that particular loop to better increase performance.

## how you tested the change

Manually:
- Running actual computations
- Cancelling run from the start (e.g. not selecting any time series)
- Stopping in middle of run.

## Where the following done:

### DIFF
--- a/src/main/java/decodes/tsdb/comprungui/CompRunGuiFrame.java
+++ b/src/main/java/decodes/tsdb/comprungui/CompRunGuiFrame.java
@@ -51,6 +51,7 @@ import java.util.Vector;
 import java.util.concurrent.ExecutionException;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
+import java.util.function.Supplier;
 
 import javax.swing.BorderFactory;
 import javax.swing.BoxLayout;
@@ -63,7 +64,6 @@ import javax.swing.JPanel;
 import javax.swing.JProgressBar;
 import javax.swing.JScrollPane;
 import javax.swing.JTable;
-import javax.swing.SwingUtilities;
 import javax.swing.SwingWorker;
 import javax.swing.WindowConstants;
 import javax.swing.border.Border;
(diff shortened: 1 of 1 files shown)


src/main/java/decodes/tsdb/comprungui/CompRunGuiFrame.java
```diff
@@ -760,6 +780,69 @@ private void runButtonPressed()
 			return;
 		}
 
+		TsGroup compGroup = isSelectionValid(compVector);
+
+		if (compGroup == null)
+		{
+			Logger.instance().debug3("None of the selected computations uses a group.");
```
ktarbet, line 787:
```suggestion
			Logger.instance().debug3("None of the selected computations uses a group.");
			return;
```
compGroup gets passed as argument below, not sure if null is supported.

src/main/java/decodes/tsdb/comprungui/CompRunGuiFrame.java
```diff
@@ -760,6 +780,69 @@ private void runButtonPressed()
...
+					   .log("Unable to execute computations.");
+					showError(ex.getLocalizedMessage());
+				}
+			}
+		};
+		runCompWorker.addPropertyChangeListener(event -> updateProgress(event));
+		runCompWorker.execute();
+	}
+
+	private TsGroup isSelectionValid(Collection<DbComputation> compVector)
```
ktarbet, line 844: isSelectionValid as a method name really 'feels' like it should return a Boolean.

MikeNeilson: Hmm. yeah. That was my original intent but than couldn't return all the correct information.

ktarbet (approved):

src/main/java/decodes/tsdb/comprungui/CompRunGuiFrame.java
MikeNeilson: Completely changed the way this works.

ktarbet (approved):

PR #525, opened 2024-03-20
Feature/333 lrgs tests

MikeNeilson:
## Problem Description

Fixes #333. 

## Solution

Slowly setting up a test harness and initial tests. This PR will be a balance of ease vs need.

e.g. DAMS-NT (DRGS), super easy to fake with reasonable data, HRIT files/HRIT, kinda need the actual data.
So  after having tests that excercise the archive itself, I'll probably start with DRGS + Network DCPs. 
If the code isn't too much I'll throw a DDS client test in there but that may be a different PR.

This PR is going follow the principal of "some directly useful test of a system is superior to no useful tests."

## how you tested the change

Describe what was done to test the change. This section can be left blank 
if automated tests demonstrating usage are provided in the PR.

## Where the following done:

- [ ] Tests. Check all that apply:
   - [ ] Unit tests created or modified that run during ant test.
   - [x] Integration tests created or modified that run during integration testing
         (Formerly called regression tests.)
   - [ ] Test procedure descriptions for manual testing
- [ ] Was relevant documentation updated?
- [ ] Were relevant config element (e.g. XML data) updated as appropriate

If you aren't sure leave unchecked and we will help guide you to want needs changing where.

PR #534, opened 2024-03-25
Refactor/reduce build boilerplate

MikeNeilson:
## Problem Description

After getting #525 working I realized we'd be running the LRGS multiple times in the automated build (as it's part of the current integration test suite but doesn't depend on any of that infrastructure.)

After deciding it would be best to create yet-another-test-target the amount of duplicate XML that would be required bothered me. Thus I've reduced the boilerplate to make it easier to add new compile and test targets.

## Solution

- Created a `compile-java` macrodef to handle compiling java and dealing with when things are up-to-date
- Created a `test` macrodef to handle running tests.

Callers of the task are required to set the src files for compile, classes for test, and the classpath for both.
Additional jvmargs may be provided for the test tests.
Additional filessets to determine if the output is up-to-date may also be provided.

## how you tested the change

Since the tests pass, the build works. (NOTE: at least for those things under test, but that's now varied enough I feel confident things are in a good state.)

## Where the following done:

### DIFF
--- a/src/main/java/lrgs/gui/LrgsBuild.java
+++ b/src/main/java/lrgs/gui/LrgsBuild.java
@@ -19,8 +19,8 @@ package is built.
 */
 public class LrgsBuild
 {
-	public static final String buildDate = "@DATE@";
-	public static final String rcNum = "@RCNUM@";
-	public static final String version = "@VERSION@";
+	public static final String buildDate = "7.0.13";
+	public static final String rcNum = "13";
+	public static final String version = "7.0";
 }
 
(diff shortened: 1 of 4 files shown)


src/main/java/lrgs/gui/LrgsBuild.java
```diff
@@ -19,8 +19,8 @@
 */
 public class LrgsBuild
 {
-	public static final String buildDate = "@DATE@";
-	public static final String rcNum = "@RCNUM@";
-	public static final String version = "@VERSION@";
+	public static final String buildDate = "7.0.13";
```
ktarbet, line 22: Is this, setting specific values, intentional?

MikeNeilson: Yeah. Though certainly not ideal.

ktarbet (approved):

PR #535, opened 2024-03-25
Update logging and some minor formatting changes in PlatformListIO.java

ktarbet:
Small step toward issue #523


src/main/java/decodes/sql/PlatformListIO.java
```diff
@@ -100,1194 +38,1195 @@
...
+                    {
+                        try (ResultSet rs = stmt.executeQuery(q);)
+                        {
+                            DbKey pid = Constants.undefinedId;
+                            if (rs != null && rs.next())
+                            {
+                                log.info("Imported platform clashes with old platform ID={}"
+                                     +" -- deleting old platform.",pid);
+                                pid = DbKey.createDbKey(rs, 1);
+                                rs.close();
```
MikeNeilson, line 682: these aren't actually necessary. both the statement and result set are already within a try-with.

MikeNeilson (changes requested):

MikeNeilson (approved):

PR #538, opened 2024-03-26
Refactor/update compile dialog

MikeNeilson:
## Problem Description

Fixes #. 

Developers will often get a warning about the direct use of the `com.sun.tools.javac.Main` package. It's also deprecated in 9+ up (at least as accessible that way.)

The logic in the doCompile step tried to do too much on behalf of the user.

## Solution

Use the javax.tools.ToolProvider.getSystemJavaCompiler() to get an instance of `JavaCompiler`. This is the standard interface to get the compiler.

Additionally modified addToJar tasks to use JarFile and JarOutputStream to handle creating and updating the Jar; thus removing another direct `com.sun...` dependency.

On a JRE an error is returned and nothing else is done. As the `.getSystemJavaCompiler` will return null.

For the curious, the returned compiler *is* `com.sun.tools.javac.api.JavacTool` with OpenJDK 17. But this avoids that direct dependency that causes the compiler warning message and would also allow for alternate implementations (e.g. in other JDKs)

NOTE: this also means our build is now compatible with up-to java 17. The tests pass using that JDK. The output is still targeted at 1.8. I know a JDK 21 won't work, there's runtime issues due some APIs being deprecated but I'll add another PR after this one to expand the build and test actions to include LTS JDKs up to 17 for runtime verification.

## how you tested the change

- Open algoedit
- Create algorithm (file -> new) doesn't need to do much
- Compile & and Jar
- Verify class file in correct location in and jar is valid (used the JD-GUI java decompiler)
- Create new algorithm with different name and slightly altered contents
- Compile
- Add to existing jar
- Using JD-GUI again verify that both files are in the jar and JD-GUI is able to decompile them.

NOTE: tested with Java JDK 17, set to generate source/target 1.8 files.

## Where the following done:

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -254,8 +254,8 @@ more information about each.
             <fileset dir="${opendcs.resources}"/>
             <fileset dir="${opendcs.classes}"/>
             <!-- AW_AlgorithmTemplate.java needed by Algorithm Editor. -->
-            <fileset dir="${src.dir}/decodes/tsdb/algo">
-                <include name="AW_AlgorithmTemplate.java"/>
+            <fileset dir="${src.dir}" >
+                <include name="**/AW_AlgorithmTemplate.java"/>
             </fileset>
         </jar>
     </target>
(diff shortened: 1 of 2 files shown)


ktarbet (approved):
This is great!   I no longer need to comment out specific lines when compiling with Java 11.

PR #542, opened 2024-03-29
Refactor/launcher stream usage

MikeNeilson:
## Problem Description

With recent logging changes and tweaks to the launcher profile system users have been reporting certain interfaces are slower than usual.

## Solution

Modify the profile threads that wait on application data to use BufferedReaders so that there are more waits.
Lowered the priority to not hang up other operations.

As an extra measure, implemented simple async logging within FileLogger and TraceLogger to improve application speed.
GUI apps do appear to load faster/more consistently but, for example, the NWP comp that uses a coup with near 14000 time series continues to be slow processing only 7 timeseries/second. See #526 for more details on that.

If, after additional measurement, the async logging proves valuable I'll reorganize the code to avoid so much duplication.

## how you tested the change

Running applications directly and from the launcher to gauge the speed of when thing first rendered.
It did feel faster to me.

I suspect the local test on my VM is not a fair test. The VM itself is hosted on an independant SSD from the OS and writers are generally very fast. The async logger change *should* help in situations of NFS, HDDs, or SD cards (e.g. a raspberry pi, which I'll test later when I have access.)

I plan to also test this change on the HEC T7 (SPARC/Solaris, uses NFS for just about everything) but likely no earlier than Monday.

## Where the following done:

### DIFF
--- a/src/main/java/decodes/dbeditor/TraceLogger.java
+++ b/src/main/java/decodes/dbeditor/TraceLogger.java
@@ -1,6 +1,11 @@
 
 package decodes.dbeditor;
 
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
 import ilex.util.Logger;
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/java/decodes/dbeditor/TraceLogger.java
```diff
@@ -11,13 +16,39 @@ public class TraceLogger extends Logger
...
 		super(procName);
 		this.dlg = null;
+		writerThread = new Thread(() ->
+		{
+			while (closeOperations.get() == false)
+			{
+				try
+				{
+					String line = queue.poll(1, TimeUnit.SECONDS);
+					if (dlg != null)
```
ktarbet, line 37:
```suggestion
					if (dlg != null && line != null)
```

ktarbet:
I have a PR into this PR for javaDoc improvement and minor housekeeping

https://github.com/opendcs/opendcs/pull/551

ktarbet (approved):

PR #543, opened 2024-03-30
Move LRGS properties to provider.

MikeNeilson:
This is to reduce log output that would be confusing.

## Problem Description

Kept seeing messages in the log for variables we shouldn't.

## Solution

Created lrgs property provider to organize and fixed typo,

## how you tested the change

Ran LRGS manually to verify operations and logging.

## Where the following done:


-
agilmore2, line 1: ?

lrgs.nohup
agilmore2, line 1: Likewise

src/main/java/org/opendcs/utils/properties/DcsVariablesProvider.java
agilmore2, line 1: So is there a default properties file as well, or is it generated on install?

lrgs.nohup
MikeNeilson: Oops. Didn't realize those were even checked it. That's what i get for running startLRGS in the priject root.

src/main/java/org/opendcs/utils/properties/DcsVariablesProvider.java
MikeNeilson: No, these are setup on program start (startLRGS sets the LRGS home, and decj sets the other variables.) 
LRGS included some extras beyond LRGSHOME that are set at runtime so I split them out.
But all this does is pull from either `System.getProperty` or, in the case of that lrgs report, a custom properties object that gets created.

The `JavaPropertiesProvide` does the same thing but requires using `${java.<Some name>}` and that would've require changing a lot of in-place configs so instead Lrgs and DcsVariablesProvider simply check for those fixed names and extracts the value.

The need arose out of needing to be able to inform the user that they weren't using a provider (e.g. in the case of secrets especially) but choosing to fallback and not warn left some difficult edge cases that were basically impossible to sort out.

ktarbet:
I've compiled this branch and started LRGS -- The warning message doesn't give me specific steps to take.  Could the message indicate what file to edit (if that is what is needed) ?

```
C:\test\OPENDCS.lrgs\bin>set LRGSHOME=c:\test\opendcs.lrgs

C:\test\OPENDCS.lrgs\bin>startLRGS.bat
WARNING 04/12/24 16:57:55 property value source was not set. Assuming original behavior and retrieving 'LRGSHOME' from provided properties object.This will be removed in future releases and you should update values to ${java.LRGSHOME} at your earliest convenience.
```

MikeNeilson:
Hmm, not for that one. Odd though, that should be specifically covered in this change.

MikeNeilson:
is it possible that startLRGS.bat is doing something that unsets the variable instead of using it?

ktarbet:
Opps.  I got versions confused.  It is working with out the error message.

ktarbet (approved):
I was able to startup LRGS, config a remote LRGS for backup data.  Also connected with message browser, and rtstat.

MikeNeilson:
That'll do it. thanks!

PR #544, opened 2024-04-01
Fix which column name to use

MikeNeilson:
## Problem Description

Discovered while testing a branch that I got a bit of login incorrect causing and an invalid sql statement.

## Solution

Fix order

## how you tested the change

Tested manually.... bit surprised one of the integration tests didn't pick it up though.

## Where the following done:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,7 +75,7 @@ jobs:
           name: test-results-${{matrix.platform}}-jdk-${{matrix.jdk}}
           path: |
             build/reports/**
-            build/test-integration/tmp/configs**
+            build/test/test-integration*/tmp/configs**
             build/test-gui.webm
 #docker-images:
 #  runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


MikeNeilson:
@agilmore2 can you sanity check the logic in the second instance for me.
For now I'm going ignore that fact the this is being done twice in two different ways and will fix that later.

MikeNeilson:
I'm also not entirely sure why the tests didn't capture this mistake.

MikeNeilson:
Oddly this is working locally. suspect timing issue and the actions runner not having quite enough power to match my laptop.

MikeNeilson:
Reviewing all the SQL it looks like no one actually uses the site_datatype_id in the column anymore. even HDB shows ts_id and you have a mapping table.

MikeNeilson:
Nevermind, Since the logic was in two places, when I cleaned up that code the or condition never got called. I just fixed  #520 to have that OR in the now single setup of that column name.

PR #569, opened 2024-05-16
Bugfix/519 476 user directory usage

### LINKED ISSUE #519: Python Algorithms on windows don't have DCSTOOL_HOME environment
**Describe the bug**
When trying to run python algorithms on windows the PyAlgoEnv.py file on line ten uses os.environ instead of java->System.getProperty to retreieve DCSTOOL_HOME

If this happens a "None Type cannot be added to str" happens, which is confusing.

**To Reproduce**
1. Fresh install,  DCSTOOL_HOME user or system environment does not exist.
2. Create or otherwise run a python algorithm.
3. See error

**Expected behavior**
System should just work. Either the installer should create the variable, or like the other components of opendcs System.getProperty, or preferably `EnvExpander.expand` to get the value.

**Screenshots**

 - OS: Windows
 - Version 7.0.x

**Additional context**

MikeNeilson:
## Problem Description

Fixes #519 and #476. 

## Solution

#519 - Use environment to allow user override, fall back to properties if not set, decj or decj.bat will always set the Java Property

#476 - Scripts now determine there location to find the appropriate decj instance in unix like they do in windows.
Additionally added logic to decj and decj.bat to create an appropriate default instance.

$HOME/.opendcs in unix land
%APPDATA%\opendcs in windows. (This maps to the User's "AppData\Roaming" directory.

I also made the LRGS scripts take a hard stance on LRGSHOME being set and preferably to a location outside the install directory However, if set it doesn't care, the logic only triggers if an LRGSHOME environment variable is not set.

## how you tested the change

Manually exercising the scripts on windows and Linux to very operations.

## Where the following done:


install/bin/decj
```diff
@@ -25,16 +35,10 @@ fi
...
-  elif [ -d $HOME/opendcs_user ]
-  then
-    DCSTOOL_USERDIR=$HOME/opendcs_user
-  else
-    DCSTOOL_USERDIR=$HOME
-  fi
+  DCSTOOL_USERDIR=$HOME/.opendcs
+  echo "Creating Local User Directory and initial properties in ${DCSTOOL_USERDIR}"
+  mkdir -p $DCSTOOL_USERDIR
+  cp $DH/decodes.properties $DCSTOOL_USERDIR/user.properties
```
ktarbet, line 41: This should probably be part of install ;  If this runs every time that may clobber data.
May need to update documentation.

ktarbet: OH, I see this only runs if DCSTOOL_USERDIR doesn't exist.

install/bin/decj.bat
```diff
@@ -21,4 +21,18 @@ for /R "%APP_PATH%/dep" %%a in (*.jar) do (
 
 set "CLASSPATH=!CLASSPATH!"
 
-java -Xmx240m -cp "!CLASSPATH!" -DDCSTOOL_HOME="%APP_PATH%" -DDECODES_INSTALL_DIR="%APP_PATH%" -DDCSTOOL_USERDIR="%APP_PATH%" %*%
+if not defined DCSTOOL_USERDIR (
+  set DCSTOOL_USERDIR="%APPDATA%\opendcs"
```
ktarbet, line 25: consider making this consistent with Unix,   leading dot    '.opendcs'

install/bin/editPasswd
```diff
@@ -1,2 +1,6 @@
+#!/bin/bash
+SCRIPT=$(readlink -f "$0")
+# Absolute path this script is in, thus /home/user/bin
+APP_PATH=$(dirname "$SCRIPT")
 
-%INSTALL_PATH/bin/decj ilex.util.PasswordFileEditor -f $LRGSHOME/.lrgs.passwd
+${APP_PATH}/decj ilex.util.PasswordFileEditor -f $LRGSHOME/.lrgs.passwd
```
ktarbet, line 6: Should we handle checking for $LRGSHOME like done in ./elp below.

install/bin/postInstall
ktarbet, line 1: How about deleting this file?  (and corresponding .bat version)    jars should be managed with dependency tools

install/bin/decj
MikeNeilson: Though I do need to check that the folder doesn't exist already. I think I forgot that part.

install/bin/editPasswd
MikeNeilson: Yep, good catch.

install/bin/decj
MikeNeilson: Also didn't want to put it as part of the install as we may be getting away from the installer.

install/bin/postInstall
MikeNeilson: Yeah. I'm not even sure it's actually executed anymore. Certainly not as part of the build anyways.

install/bin/decj.bat
```diff
@@ -21,4 +21,18 @@ for /R "%APP_PATH%/dep" %%a in (*.jar) do (
...
 
-java -Xmx240m -cp "!CLASSPATH!" -DDCSTOOL_HOME="%APP_PATH%" -DDECODES_INSTALL_DIR="%APP_PATH%" -DDCSTOOL_USERDIR="%APP_PATH%" %*%
+if not defined DCSTOOL_USERDIR (
+  set DCSTOOL_USERDIR="%APPDATA%\.opendcs"
+)
+
+if not exist %DCSTOOL_USERDIR%\ (
+  echo "Creating Local User Directory and initial properties in %DCSTOOL_USERDIR%"
+  mkdir %DCSTOOL_USERDIR%
+  cp %APP_PATH%\decodes.properties %DCSTOOL_USERDIR%\user.properties
```
ktarbet, line 31:
```suggestion
  copy %APP_PATH%\decodes.properties %DCSTOOL_USERDIR%\user.properties
```

izpack/opendcs-install.xml
ktarbet, line 1: consider removing all lines that have:
```xml
<parsable .... />
```

 the *dynamic* batch and bash scripts don't need to be parsed during install.

install/bin/decj.bat
MikeNeilson: Good catch.

docs/source/start-software.rst
```diff
@@ -270,12 +272,31 @@ file into a browser).
...
+
+
+This allows new users to get started more quickly. Additionally if you are setting up a fresh SQL based database, the
+contents of `$DCSTOOL_USERDIR/edit-db` should be imported in, starting with the `loading-app` directory.
+
+.. NOTE::
+   
+   We are currently in the process of improving how this required standard data is handled. We've provided this basic
+   instruction to help users get started but would preferred not to write out detail documentation since we'll be changing it soon.
+   If you have further questions please use github.com/opendcs/opendcs/discourse for more clarity.
```
ktarbet, line 293:
```suggestion
   If you have further questions please use https://github.com/opendcs/opendcs/discussions for more clarity.
```

install/bin/decj
```diff
@@ -5,16 +5,26 @@
...
+DH=$(realpath "${APP_PATH}/..")
 
 # Make sure DCSTOOL_HOME is set
 if [ -z "$DCSTOOL_HOME" ]
 then
   DCSTOOL_HOME=$DH
   export DCSTOOL_HOME
+elif [ "$DH" != "$DCSTOOL_HOME" ]
+then
+  echo "Environment has set a diffent DCSTOOL_HOME than the location of the application start script."
```
ktarbet, line 21:
```suggestion
  echo "Environment has set a different DCSTOOL_HOME than the location of the application start script."
```

install/bin/elp-local
```diff
@@ -1,12 +1,14 @@
...
+# Absolute path this script is in, thus /home/user/bin
+APP_PATH=$(dirname "$SCRIPT")
 
-DH=%INSTALL_PATH
 
 if [ -z "$LRGSHOME" ] 
 then
-  LRGSHOME=$DH
-  export LRGSHOME
+  echo "the environment variable 'LRGSHOME' should be set to a path outside the installation directory."
```
ktarbet, line 10:
```suggestion
  echo "Error: The environment variable 'LRGSHOME' is not set.   Set to a path outside the installation directory."
```

install/bin/decj.bat
```diff
@@ -21,4 +21,20 @@ for /R "%APP_PATH%/dep" %%a in (*.jar) do (
...
+if not defined DCSTOOL_USERDIR (
+  set DCSTOOL_USERDIR="%APPDATA%\.opendcs"
+)
+
+if not exist %DCSTOOL_USERDIR%\ (
+  echo "Creating Local User Directory and initial properties in %DCSTOOL_USERDIR%"
+  echo "The default XML database has been copied to this directory."
+  mkdir %DCSTOOL_USERDIR%
+  copy %APP_PATH%\decodes.properties %DCSTOOL_USERDIR%\user.properties
+  copy /R %APP_PATH%\edit-db %DCSTOOL_USERDIR%\edit-db
```
ktarbet, line 33:
```suggestion
  xcopy %APP_PATH%\edit-db %DCSTOOL_USERDIR%\edit-db /E /I
```
unfortunately /R is not part of copy

xcopy /?

/E           Copies directories and subdirectories, including empty ones.
 /I           If destination does not exist and copying more than one file,
               assumes that destination must be a directory.

docs/source/dev-docs.rst
```diff
@@ -127,6 +127,40 @@ The following workflow can be used:
...
+
+    If this is a major concern you should set the DCSTOOL_USERDIR for the session ant runs in
+    to point to a directory that only contains profiles that connect to test systems.
+
+.. code-block:: bash
+    # to just run the launcher
+    ant run
+
+    # to run a specific app
+    ant run -Dopendcs.app=compedit
```
ktarbet, line 149: note: the run target seems to assume you have first done a build, such as 'ant opendcs'

ktarbet (approved):

docs/source/dev-docs.rst
MikeNeilson: There's a dependency on "stage" that will force a build to happen. so this is safe to run before `ant jar`.

MikeNeilson:
@ktarbet Since we're creating a default profile with the XML database in the new user directory, should we just go ahead and do the initial baseline/metadata import?

It's not trivial to start setting everything up manually and you have to know what to import. Seems like doing it here would:
1. Simplify the initial setup process more
2. Provide at least some maintained guidance on what to do.

Future work to better organize the metadata setup would just be used here as well so it's not like it's going to be even more to maintain.

ktarbet:
@MikeNeilson 

During the install process, yes, it would be good to install the XML database.  That would be cleaner that having the scripts in ./bin do a just-in-time install.

I think separation of user data location from the application install location is worth the mental model change that will be required for users.  

Users that have an existing XML database may need to manually adapt, and another installed XML database could be confusing.


I'm concerned we may be doing too much in a single PR if we keep adding changes.

MikeNeilson:
I agree with your point about this PR already getting somewhat large.

Though not with during the installer, I think.

You've suggested just a zip, so then when would it happen?
That would be additional work to add to the installer, though it wouldn't be super difficult, basically just the same logic we currently have in the decj files + dbimport.

Issue #241, opened 2023-06-15
Replace LoboBrowser components


MikeNeilson:
**Component to be refactored**
LoboBrowser code

**Purpose of refactoring**
I just spent about 3 hours trying to figure out why a method didn't exist during compile and determine that LoboBrowser embedded the class files of one of it's dependencies instead of just depending on them. And since java is first class loaded wins, I couldn't use the newer version of the library I desired.

This is the 2nd library that has bit us with Lobo. the first was JOOQ and this time it was slf4j. 

**Proposed method**
There are certainly other HTML renders available which is primarily what it is used for. There is some usage in the WebDirectoryDataSource however that can likely be replaced simple some java 8 built-ins.

**Additional context**
yes I'm mostly frustrated, but the packaging is extremely sloppy so why would we want to depend on such.

ktarbet:
Does anyone have an example using WebDirectoryDataSource; or even an example web page/url to test with?

WebDirectoryDataSource.java mentions dd.weatheroffice.gc.ca   -- I can't get a response.

ktarbet:
https://github.com/jhy/jsoup/ is an option for parsing html.

ktarbet:
link for reference:

https://dd.weather.gc.ca/bulletins/

MikeNeilson:
Reviewing the code again, lobobrowser/cobra are only used in WebDirectory datasource. 

Additionally looking at the functionality, the HTML doesn't really need to be "parsed" just scanned for `<a href="...` tags that make sense and used to get the direct directory or a file.

So just using the java regular expressions should work. The structure of the page itself is immaterial to the task at hand.

PR #580, opened 2024-05-24
drop RoutMonFrame

ktarbet:
investigation on issue #241

This would drop one usage of org.lobobrowser.gui

The Routing Monitor UI is in RoutingMonitor.java

Describe your solution.

delete RoutMonFrame, since a search didn't reveal any usages.


MikeNeilson (approved):
Yep, no usages.

PR #587, opened 2024-05-31
moving non-UI code out of ScreeningEditTab

ktarbet:
suggested changes into PR https://github.com/opendcs/opendcs/pull/583


MikeNeilson (approved):
Thanks, that was bugging me the moment I wrote it.

PR #592, opened 2024-06-03
User color chooser for color setting properties.

### LINKED ISSUE #548: Color selection properties should use the
Hex values are not friendly even to the advance user.

image

Properties that are colors should use a color picker

**Additional context**
Python Colors and the DecodesScript colors use this.
NOTE: we may be moving both systems to RSyntaxTextArea so this may be short lived, but I still think useful.

MikeNeilson:
## Problem Description

Fixes #548. 

## Solution

Implement a new PropertySpec type and Change all instances of Color (in the user.properties) to use it.
The properties file still uses the standard 0x prefixed HEX value, but I'm not opposed to the work of allow additional formats)

## how you tested the change

Manually. Open the properties editor for the database, see the colors selected in the properties list, on clicking edit a JColorChooser pops up and allows selection by method of users choosing.

## Where the following done:

### DIFF
--- a/src/main/java/decodes/gui/PropertiesEditPanel.java
+++ b/src/main/java/decodes/gui/PropertiesEditPanel.java
@@ -9,6 +9,9 @@ import java.awt.*;
 import javax.swing.*;
 import javax.swing.table.*;
 
+import org.cobraparser.html.style.TableCellRenderState;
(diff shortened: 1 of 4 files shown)


ktarbet (approved):
I tested and was able to change a color, then restart , and the color saved.

PR #597, opened 2024-06-05
Initial import of new algorithms.

### LINKED ISSUE #380: Mechanism to automatically load, or inform user about new algorithms.
Currently users have to know about new algorithms *before* then can be imported in. And then have to manually load an xml file.
As we update OpenDCS and add new algorithms it would be good if these could be picked up automatically.

Anything that can do the above. But preferably with an ear to the future.

In order of preference:
- SPI mechanism that algorithms implement
- Embedding XML config data for new computations into the jar
- Putting a list on the wiki and in the docs (actually we should probably do this anyways.)

In the far future SPI alone makes sense, however that combined with system that takes those and creates the appropriate entries would likely be reasonable for now.

**Additional context**
We do get asked about algorithms a lot.

### LINKED ISSUE #503: Move USACE custom algorithms into project
The algorithms here (sans pineflat inflow) are useful to anyone. 

https://github.com/HydrologicEngineeringCenter/OpenDCS-extras/tree/700d2ce26555583f15f14af0348d783e234de441/src/main/java/spk/algo/inflow

Update as required, rename package, otherwise have in distribution.

**Additional context**

MikeNeilson:
## Problem Description

Fixes #503
Fixes #380 

## Solution

Have simply imported in some custom applications and renamed and fixed formatting. Still need to type up docs and tests.
Additionally may use this to start work on some annotations or improved algo import mechanism.

## how you tested the change

## Where the following done:


ktarbet (approved):

MikeNeilson:
I'm going to add the new annotations to the algorithms I just brought in. I'll start adding them to the existing algorithms in a separate PR as that would get rather large.

I think we also want to add an addition filter in the `ExecClassSelectDialog` to exclude those already imported.

Also as a separate PR, take the work from `ExecClassSelectDialog` and extract it out and make it more generic for the database migration tool to pick things up.

src/annotations/java/org/opendcs/annotations/PropertySpec.java
ktarbet, line 1: There is another class called PropertySpec,   Will this annotation version will replace decodes.util.PropertySpec?   

If this doesn't replace consider including Annotation or something in the name of the class, such as PropertyAnnotation.

MikeNeilson: I had annotation on it and it looked... weird. And looking at the existing PropertySpec, I think this will replace it. The annotation just needs to be applied to a lot of classes before that happens. Hence the current usage is calling out the full class name when it's what's actually used.

MikeNeilson: The decodes.util.PropertySpec "TYPE" should probably just be a Class<?> entry, and I'm thinking I'll set that in the annotation and we can slowly migrate things to that mechanism.

Actually now that I say that the type is know at all entries as the Annotation is associated to the Field and it can be looked up at time of usage. I'll do that. It'll be easier to replace that way I think.

src/main/java/decodes/tsdb/algo/AW_AlgorithmBase.java
```diff
@@ -1547,7 +1552,16 @@ protected abstract void doAWTimeSlice()
...
+	public String[] getPropertyNames()
+	{
+		org.opendcs.annotations.PropertySpec[] annos = this.getClass().getAnnotationsByType(org.opendcs.annotations.PropertySpec.class);
+		ArrayList<String> ret = new ArrayList<>();
+		for (org.opendcs.annotations.PropertySpec a: annos)
+		{
+			ret.add(a.name());
+		}
+		return ret.toArray(new String[0]);
+	}
```
ktarbet, line 1564: consider putting this logic into AnnotationHelpers.

src/main/java/decodes/tsdb/compedit/AlgorithmsListPanel.java
```diff
@@ -58,45 +63,68 @@ JPanel getFieldPanel() {
...
+	protected JPanel getJContentPane()
 	{
-		if (jContentPane == null) 
+		if (jContentPane == null)
 		{
 			jContentPane = new JPanel();
 			jContentPane.setLayout(new BorderLayout());
-			jContentPane.add(getButtonPanel(), java.awt.BorderLayout.SOUTH);
+			JPanel buttonPanel = getButtonPanel();
+			JButton checkForNew = new JButton(ResourceBundle.getBundle("ilex/resources/gui").getString("EditorMenuSet.checkForNew"));
```
ktarbet, line 91: Suggestion on the naming, method checkForNew() and label 'Check for new'

How about: 'Refresh' ;   I admit I'm confused on what is being synced here.

src/main/java/decodes/tsdb/algo/AW_AlgorithmBase.java
MikeNeilson: Thanks, new I missed a spot.

src/main/java/decodes/tsdb/compedit/AlgorithmsListPanel.java
MikeNeilson: It's definitely not a "refresh", that's used in other GUIs to reload from the database. It's specifically checking for new possible algorithms from various sources.

src/main/resources/ilex/resources/gui_en_US.properties
```diff
@@ -24,6 +24,7 @@ EditorMenuSet.cut = Cut
 EditorMenuSet.copy = Copy
 EditorMenuSet.paste = Paste
 EditorMenuSet.delete = Delete
+EditorMenuSet.checkForNew = Check for New
```
ktarbet, line 27: need Spanish translation for  this.  If we use 'Update' or 'Refresh'   Spanish could be 'Actualizar'

ktarbet (approved):
This PR is getting large,  and has too many goals beyond original of new algorithms.

MikeNeilson:
Well, to be fair my goal was import new comps AND explore a better mechanism of informing the user.

MikeNeilson:
@ktarbet Just made another change to better handle the return of Fields with the annotations required.

PR #607, opened 2024-06-13
Algorithm DAO now uses bind variables and transactions.

### LINKED ISSUE #602: Bind variables for AlgorithmsDAO
`opendcs.dao.AlgorithmsDAO`

**Purpose of refactoring**
Peformance, security.

**Proposed method**
Use the new handlers

**Additional context**
Was testing importing new algorithms and got bit by a string column limit and then realize there is no bind variable usage at all.

MikeNeilson:
## Problem Description

Fixes #602 . 

## Solution

Use new handlers, including transactions where appropriate.
Additional retrieved comment column length from data to allow for future expansion.

## how you tested the change

Covered under existing integration tests.

## Where the following done:

### DIFF
--- a/src/main/java/opendcs/dao/AlgorithmDAO.java
+++ b/src/main/java/opendcs/dao/AlgorithmDAO.java
@@ -1,48 +1,17 @@
-/*
- * $Id$
- * 
- * $Log$
- * Revision 1.6  2016/11/19 15:55:51  mmaloney
- * Join PARM records with parent records to bring in implicit VPD filter in CWMS.
- *
- * Revision 1.5  2016/03/24 19:21:07  mmaloney
- * Support for algo scripts.
- *
- * Revision 1.4  2016/01/27 22:12:10  mmaloney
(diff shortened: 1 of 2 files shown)


src/main/java/opendcs/dao/AlgorithmDAO.java
```diff
@@ -54,432 +23,493 @@
...
+    {
+        super(tsdb, "AlgorithmDao");
+        if (COMMENT_LENGTH == null)
+        {
+            COMMENT_LENGTH = 1000;
+            try(DaoBase dao = new DaoBase(tsdb, module))
+            {
+                dao.doQuery("select cmmnt from cp_algorithm", rs ->
+                {
+                    ResultSetMetaData rsmd = rs.getMetaData();
```
ktarbet, line 54: Will this code re-read the meta-data for each row?    

consider   dao.getFirstResult() instead.

MikeNeilson: oh that's a good point.

PR #612, opened 2024-06-14
Devops/algorithm integration tests

MikeNeilson:
## Problem Description

Integration/Regression tests of algorithms require a rather large amount of setup to get running. 

## Solution

Currently not, initial test is to help layout components required. Goal is to be able to setup additional tests rather trivially given (algorithm, computation parameters, inputs, correct output for comparison) and just loop over them without much fuss.

Time series database is currently required to very the inputs. May be practical to bring this to the unit tests, but since some algorithms do require extra database use there's a varying level of difficulty. With the change to TsImport to have the "TsImporter" though definitely a bit more practical.

## how you tested the change

This is a test.

## Where the following done:

### DIFF
--- a/src/main/java/opendcs/dao/AlgorithmDAO.java
+++ b/src/main/java/opendcs/dao/AlgorithmDAO.java
@@ -1,48 +1,17 @@
-/*
- * $Id$
- * 
- * $Log$
- * Revision 1.6  2016/11/19 15:55:51  mmaloney
- * Join PARM records with parent records to bring in implicit VPD filter in CWMS.
- *
- * Revision 1.5  2016/03/24 19:21:07  mmaloney
- * Support for algo scripts.
- *
- * Revision 1.4  2016/01/27 22:12:10  mmaloney
- * When reading the complete list, get the params and props in one go rather than
- * individually.
- *
- * Revision 1.3  2015/12/30 20:39:23  mmaloney
- * Bugfix: listAlgorithms was using the same Statement in nested queries, causing SQL exception.
- *
- * Revision 1.2  2014/07/03 12:53:41  mmaloney
- * debug improvements.
- *
- * Revision 1.1.1.1  2014/05/19 15:28:59  mmaloney
- * OPENDCS 6.0 Initial Checkin
- *
- * This software was written by Cove Software, LLC ("COVE") under contract 
- * to the United States Government. 
- * 
- * No warranty is provided or implied other than specific contractual terms
- * between COVE and the U.S. Government
- * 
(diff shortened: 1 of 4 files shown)


src/main/java/opendcs/dao/AlgorithmDAO.java
```diff
@@ -1,48 +1,17 @@
...
- * Revision 1.1.1.1  2014/05/19 15:28:59  mmaloney
- * OPENDCS 6.0 Initial Checkin
- *
- * This software was written by Cove Software, LLC ("COVE") under contract 
- * to the United States Government. 
- * 
- * No warranty is provided or implied other than specific contractual terms
- * between COVE and the U.S. Government
- * 
- * Copyright 2014 U.S. Army Corps of Engineers, Hydrologic Engineering Center.
```
agilmore2, line 30: While this copyright is incorrect, feels strange removing these and not replacing them with anything. As a rule, should I be putting copyright and Apache license statements on the files I'm altering?

MikeNeilson: Now that you mention it, yeah. And I should be too.

src/main/java/opendcs/dao/AlgorithmDAO.java
```diff
@@ -54,432 +23,493 @@
...
+            });
+            q = "select a.algorithm_id, count(1) as CompsUsingAlgo "
+              + "from cp_algorithm a, cp_computation b "
+              + "where a.algorithm_id = b.algorithm_id "
+              + "group by a.algorithm_id";
+            doQuery(q, rs ->
+            {
+                DbKey algoId = DbKey.createDbKey(rs, 1);
+                int numCompsUsing = rs.getInt(2);
+                for(AlgorithmInList ail : ret)
```
agilmore2, line 497: Isn't this O(N^2)? Seems like these could be populated in one query.

agilmore2 (dismissed):

agilmore2:
Any reason why the ugly but passing test failed on all the Ubuntu test rigs?

MikeNeilson:
I forgot to put the annotation on it that would prevent it from running the test under the XML database. Thus it's always running and probably throwing a null pointer exception.

Also not done, i wasn't planning to merge this for a bit. This was to demo an idea to a new HEC dev and give a space to try out some ideas. Eventually I want it to behave much like the Decodes tests we have, e.g. parameterized and a consistent flow and the actual config outside this code.

Just need to get bit with all the details of computation execution setup.

ktarbet (approved):

ktarbet:
from a brainstorming session , here is an idea to organize test data.

image

Issue #601, opened 2024-06-10
Feature: Basic ability to plot time-series


ktarbet:
**Is your feature request related to a problem? Please describe.**

When debugging Computations it is awkward/difficult to use other software.

**Describe the solution you'd like**
Ability to plot from the time-series list.
![image](https://github.com/opendcs/opendcs/assets/4818531/c6ca9184-1b4e-432f-a238-bba70504850f)

The plot should work with different database backends.

A simple time-window will be estimated, which can be fine-tuned in the window the plot appears.

**Describe alternatives you've considered**
Using separate software for plotting would work, and specific solutions exist; something built-in would improve quality of life.

PR #620, opened 2024-06-17
progress on plotting time series

ktarbet:
## Problem Description

initial work on #601

## Solution

Created some classes to wrap jfreechart. supports multiple lines.

TODO:  next pull request will have date/time pickers as done in the computation tester.  Currently hardcoded for last 7 days.


src/main/java/decodes/tsdb/groupedit/TsListPanel.java
```diff
@@ -215,7 +230,24 @@ public void refresh()
...
-	
+
+	@Override
+	public void plot()
+	{
+		try
+		{
+			TimeSeriesDAI dao = theDb.makeTimeSeriesDAO();
+			final TimeSeriesIdentifier[] tsids = tsListSelectPanel.getSelectedTSIDs();
+			TimeSeriesChartFrame f = new TimeSeriesChartFrame(dao,tsids);
```
MikeNeilson, line 241: I suspect this should take the TimeseriesDb and not the TimeseriesDao, this would make it easier to keep the chart frame in memory and reuse it by just passing in new tsids. 

Or the chart shouldn't be responsible for loading the the timeseries and that work should happen in... but in a swingworker.

src/main/java/decodes/gui/TimeSeriesChartFrame.java
```diff
@@ -0,0 +1,51 @@
...
+    private final static org.slf4j.Logger log = LoggerFactory.getLogger(TimeSeriesChartFrame.class);
+    private final TimeSeriesDAI timeSeriesDAO;
+    private final TimeSeriesIdentifier[] tsids;
+    public TimeSeriesChartFrame(TimeSeriesDAI timeSeriesDAO, TimeSeriesIdentifier[] tsids )
+    {
+        this.timeSeriesDAO = timeSeriesDAO;
+        this.tsids = tsids;
+    }
+
+    public void plot(){
```
MikeNeilson, line 22: After looking at the below usage of plot, would `::plot(The data)` be better. Only one instance of the frame would need to be created and could then be reused.

MikeNeilson (commented):
A few questions, but I'll approve after some minor discussion.

src/main/java/decodes/gui/TimeSeriesChartFrame.java
ktarbet: The next step in my mind was to add a panel with a time-window selection and a 'refresh button, to the TimeSeriesChartFrame, so the initial time window of 7 days could be changed by the user.

I think you may be suggesting another object TimeSeriesSeriesChartData that gets passed in, so the time-window selection would be elsewhere?

MikeNeilson: No, you're passing into the constructor a DAO and a list. I'm suggestion that this function just takes "the data" already processed for rendering. 

You're refresh button would then be responsible for actually getting the data.

E.g. why does the ChartFrame need to know how to access the database? Seems like it could take only it's constructions, `TimeSeriesLine`.

src/main/java/decodes/tsdb/groupedit/TsListPanel.java
MikeNeilson: Yeah, I'm suggesting it be given list of TimeSEriesLine, and TsListController/Panel would handle actually retrieving the data.

src/main/java/decodes/gui/TimeSeriesChartFrame.java
ktarbet: Because, when expanded later the Frame  (hopefully refactored later to have tsPlotController) will need to be able to update based on different time window inputs ( startDate:   ,  endDate) not yet implemented.    That logic isn't TsListPanel specific. 

The 'Refresh' button, start and end time inputs will also need a home.

MikeNeilson: Okay, so your intention is to move that logic out later. Just when there's a more appropriate place?

MikeNeilson (approved):

PR #626, opened 2024-06-20
Refactor/tsdb caches

MikeNeilson:
## Problem Description

 

Existing caching implementation was causing performance issues as the DAO objects we're constantly getting reset so each would need calculate that it needed to fill the cache.

While the cache was static, and thus should've been fine, this work moves the actual Cache up a level, and create an interface to allow alternative cache implementations in the future.

After discussion with @agilmore2 reviewed current design and implemented a possible solution that will be useful going forward.

## Solution

Create interface for DbObjectCache.
Move Site cache to TimeseriesDb/SqlDatabaseIO (NOTE: this does mean there will be two caches, not available easily, if at all in the current design.)

TODOs:
- [ ] Add a Cached "null" return for cases where the same name was recently search and not found (this helps with compedit rendering the group for a computation... and probably compdepends as well.)
- [x] Rename the original implementation.
- [ ] MXBeans for tracking cache usage from JMX.
- [ ] Move the cache somewhere that can be shared between SqlDatabaseIO and TimeseriesDatabase. Like something related to DecodesSettings (if made immutable can be used as a hashkey).

## how you tested the change

Test of the interface created.

Integration tests already cover object cache usage. (All SiteDAO implementations use the cache in some way.)

## Where the following done:

### DIFF
--- a/src/main/java/decodes/cwms/CwmsTimeSeriesDb.java
+++ b/src/main/java/decodes/cwms/CwmsTimeSeriesDb.java
@@ -698,8 +698,10 @@ import opendcs.dai.LoadingAppDAI;
 import opendcs.dai.ScheduleEntryDAI;
 import opendcs.dai.SiteDAI;
 import opendcs.dai.TimeSeriesDAI;
+import opendcs.dao.CachableDbObject;
 import opendcs.dao.DaoBase;
 import opendcs.dao.DaoHelper;
+import opendcs.dao.DbObjectCache;
 import opendcs.dao.LoadingAppDao;
 import opendcs.opentsdb.OpenTsdbSettings;
 import opendcs.util.sql.WrappedConnection;
@@ -1683,6 +1685,4 @@ public class CwmsTimeSeriesDb
 			Logger.instance().warning("Unable to close returned connection: " + ex.getLocalizedMessage());
 		}
 	}
-	
-	
 }
--- a/src/main/java/decodes/cwms/CwmsTimeSeriesDb.java
+++ b/src/main/java/decodes/cwms/CwmsTimeSeriesDb.java
@@ -701,7 +701,8 @@ import opendcs.dai.TimeSeriesDAI;
 import opendcs.dao.CachableDbObject;
 import opendcs.dao.DaoBase;
 import opendcs.dao.DaoHelper;
-import opendcs.dao.DbObjectCache;
+import opendcs.dao.ScheduledReloadDbObjectCache;
+import opendcs.dao.SiteDAO;
 import opendcs.dao.LoadingAppDao;
 import opendcs.opentsdb.OpenTsdbSettings;
 import opendcs.util.sql.WrappedConnection;
@@ -723,6 +724,7 @@ import ilex.util.TextUtil;
 import ilex.var.NamedVariable;
 import ilex.var.Variable;
 import decodes.cwms.rating.CwmsRatingDao;
+import decodes.cwms.validation.Screening;
 import decodes.cwms.validation.dao.ScreeningDAI;
 import decodes.cwms.validation.dao.ScreeningDAO;
 import decodes.db.Constants;
@@ -784,7 +786,20 @@ public class CwmsTimeSeriesDb
 		curTimeName = "sysdate";
 		maxCompRetryTimeFrmt = "%d*1/24";
 		module = "CwmsTimeSeriesDb";
-		
+		cacheMap.put(Screening.class,
+        new ScheduledReloadDbObjectCache<Screening>(SiteDAO.CACHE_MAX_AGE, false, (cache) ->
+        {
+        try (ScreeningDAI dao = this.makeScreeningDAO())
+        {
+            dao.reloadCache(cache);
+        }
+        catch (DbIoException ex)
+        {
+            log.atError()
+            .setCause(ex)
+            .log("Unable to refresh Site Cache.");
+        }
+        }, cacheExecutor));
 	}
 
 	public static Connection getDbConnection(final CwmsConnectionInfo info) throws BadConnectException
--- a/src/main/java/decodes/cwms/CwmsTimeSeriesDb.java
+++ b/src/main/java/decodes/cwms/CwmsTimeSeriesDb.java
@@ -786,7 +786,8 @@ public class CwmsTimeSeriesDb
 		curTimeName = "sysdate";
 		maxCompRetryTimeFrmt = "%d*1/24";
 		module = "CwmsTimeSeriesDb";
-		cacheMap.put(Screening.class,
+                // add Cwms specific cache
+		cacheMap.put(Screening.class,     
         new ScheduledReloadDbObjectCache<Screening>(SiteDAO.CACHE_MAX_AGE, false, (cache) ->
         {
         try (ScreeningDAI dao = this.makeScreeningDAO())
@@ -797,7 +798,7 @@ public class CwmsTimeSeriesDb
         {
             log.atError()
             .setCause(ex)
-            .log("Unable to refresh Site Cache.");
+            .log("Unable to refresh Screening Cache.");
         }
         }, cacheExecutor));
 	}
--- a/src/main/java/decodes/tsdb/TimeSeriesDb.java
+++ b/src/main/java/decodes/tsdb/TimeSeriesDb.java
@@ -650,7 +650,7 @@ public abstract class TimeSeriesDb
                         {
                             log.atError()
                             .setCause(ex)
-                            .log("Unable to refresh Site Cache.");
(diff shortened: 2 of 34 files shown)


src/main/java/opendcs/opentsdb/OpenTsdb.java
```diff
@@ -30,6 +35,7 @@
 import decodes.db.DataType;
 import decodes.db.Database;
 import decodes.db.PresentationGroup;
+import decodes.db.Site;
```
ktarbet, line 38:
```suggestion
```

src/main/java/org/opendcs/database/CacheKey.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+        else
+        {
+            this.key = key;
+        }
+        this.uniqueName = uniqueName;
+    }
+
+    /**
+     * Loose equality. If only one set needs to match of the other set has a null value.
```
ktarbet, line 48:
```suggestion
     * Loose equality. If only one set needs to match or the other set has a null value.
```

ktarbet: do we need any null checks to make the above comment work without NPE?

ktarbet (approved):
I think these are the 'same names'  you mentioned  you many update.

org.opendcs.database.DbObjectCache  (Interface)
opendcs.dao.DbObjectCache

MikeNeilson:
Okay, so it's a little large, but seems fairly well constrained. At a minimum it didn't make the integration tests take any longer... and it's entirely plausible the CWMS integration may have run a bit faster, sadly no hard measurements available.

I'm going to run this on one of the HEC test systems to see how it does. My homelab is already running the "auth/datasource" PR so it's going to take me a bit to rebase and merge there, plus I only have 3 sites and 1 computation so performance just isn't an issue.

src/main/java/decodes/tsdb/TimeSeriesDb.java
```diff
@@ -485,11 +495,26 @@
...
+    private static final org.slf4j.Logger log = LoggerFactory.getLogger(TimeSeriesDb.class);
     public static String module = "tsdb";
 
     /** The application ID of the connected program */
     protected DbKey appId = Constants.undefinedId;
 
+    /**
+     * Thread pool used for running cache reload tasks.
+     */
+    protected static final ScheduledExecutorService cacheExecutor = Executors.newScheduledThreadPool(1, (r) -> {
```
ktarbet, line 507: There is also a public cacheExecutor in SqlDatabaseIO.  Do these two cacheExecutors work together?

src/main/java/decodes/tsdb/TimeSeriesDb.java
```diff
@@ -593,9 +618,94 @@ public TimeSeriesDb()
 
         cpCompDepends_col1 = isHdb() ? "TS_ID" : "SITE_DATATYPE_ID";
         getLogDateFormat().setTimeZone(TimeZone.getTimeZone("UTC"));
+        setupCaches();
+    }
+
+    private void setupCaches()
```
ktarbet, line 624: I don't understand the difference in this setupCaches() and the one in SqlDatabaseIO.

src/main/java/decodes/tsdb/TimeSeriesDb.java
```diff
@@ -593,9 +618,94 @@ public TimeSeriesDb()
...
+                     {
+                        try (TimeSeriesDAI tsDai = this.makeTimeSeriesDAO())
+                        {
+                            tsDai.reloadTsIdCache(cache);
+                        }
+                        catch (DbIoException ex)
+                        {
+                            log.atError()
+                            .setCause(ex)
+                            .log("Unable to refresh Site Cache.");
```
ktarbet, line 653:
```suggestion
                            .log("Unable to refresh TimeSeriesIdentifier Cache.");
```

src/main/java/decodes/tsdb/TimeSeriesDb.java
```diff
@@ -593,9 +618,94 @@ public TimeSeriesDb()
...
+                    {
+                        try (ComputationDAI dao = this.makeComputationDAO())
+                        {
+                            dao.fillCache(cache);
+                        }
+                        catch (DbIoException ex)
+                        {
+                            log.atError()
+                            .setCause(ex)
+                            .log("Unable to refresh Site Cache.");
```
ktarbet, line 667:
```suggestion
                            .log("Unable to refresh DbComputation Cache.");
```

src/main/java/decodes/tsdb/TimeSeriesDb.java
```diff
@@ -593,9 +618,94 @@ public TimeSeriesDb()
...
+            {
+                try (AlarmDAI dao = this.makeAlarmDAO())
+                {
+                    dao.reloadCache(cache);
+                }
+                catch (DbIoException ex)
+                {
+                    log.atError()
+                    .setCause(ex)
+                    .log("Unable to refresh Site Cache.");
```
ktarbet, line 681:
```suggestion
                    .log("Unable to refresh AlarmScreening Cache.");
```

src/main/java/decodes/tsdb/TimeSeriesDb.java
```diff
@@ -593,9 +618,94 @@ public TimeSeriesDb()
...
+                    {
+                        try (TsGroupDAI dao = this.makeTsGroupDAO())
+                        {
+                            dao.fillCache(cache);
+                        }
+                        catch (DbIoException ex)
+                        {
+                            log.atError()
+                            .setCause(ex)
+                            .log("Unable to refresh Site Cache.");
```
ktarbet, line 702:
```suggestion
                            .log("Unable to refresh TsGroup Cache.");
```

ktarbet (commented):
Do we have a strategy to test , including HDB?

src/main/java/decodes/cwms/CwmsTimeSeriesDb.java
```diff
@@ -782,7 +786,20 @@ public CwmsTimeSeriesDb()
 		curTimeName = "sysdate";
 		maxCompRetryTimeFrmt = "%d*1/24";
 		module = "CwmsTimeSeriesDb";
-		
+		cacheMap.put(Screening.class,
```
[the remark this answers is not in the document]
MikeNeilson: For this particular one I know this screening only works in CWMS at the moment and the test is simply to open the screening editor and work, plus have a screening comp.

I'll be setting up the HEC T7 to use this branch for one of the datasets, probably SPK since no one is using it.

Otherwise the other cache operations, except unfortunately the reload, are covered under the integration tests.

src/main/java/decodes/tsdb/TimeSeriesDb.java
MikeNeilson: We *could* make the executors part of the settings object, but I'd want to think about how that would affect work like #520 and such. My gut says in a good way, DecodesSetttings could just be the application context vs just settings. but again, definitely larger task than what this PR already is.

src/main/java/decodes/tsdb/TimeSeriesDb.java
MikeNeilson: Well, there really isn't actually, we could make this a static function that takes a DatabaseConnectionOwner and CacheMap?

SqlDatabaseIO vs TimeSeriesDb is one of the unfortunate parts of OpenDCS history - it'll slowly get phased out, but changes like this, even though some work is duplicated are I think necessary to make that happen. E.g. this adds a little complexity but overall simplifies some operations *and* future changes.

MikeNeilson:
> Do we have a strategy to test , including HDB?

I'll be setting up the HEC T7 to use this branch for one of the datasets, probably SPK since no one is using it.

Otherwise the other cache operations, except unfortunately the reload, are covered under the integration tests. So I know the basic workflow of the caches works, it just remains to be seen if performance is improved.

@agilmore2 will have to test HDB, though I have enough infra at home to setup a simple test oracle database to help when I'm making these changes, it's just the the instructions at https://github.com/usbr/hdb are thus far as clear as mud (to me - probably fine if I did it once.)

However, this came about due to conversation about HDB performance being so bad 7.0 was unusable due to the caches constantly being recreated.

MikeNeilson:
Okay, definitely needs a bit more testing. discovered that comproc will get connection closed errors which means something in this design is holding a connection open.

Plus I need to rebase from master given the find/get addition that was just merged in.

MikeNeilson:
I'm closing this, been doing some research on disk backed (e.g. disk and memory) as well as caching Optional<Obj> and other strategies. will likely revisit after #681 is finished.

PR #628, opened 2024-06-25
Added ToIrregularUsingPattern Algorithm

### LINKED ISSUE #314: Request: A CCP Algorithm similar to "To Irregular using Pattern" function in DSS-VUE
Currently there doesn't appear to be an algorithm that interpolates a dataset using the time pattern of another dataset.

I was wondering if I could request an algorithm that's similar to the "To Irregular using Pattern" function in DSS-VUE (Please see screenshot below)

https://www.hec.usace.army.mil/confluence/dssdocs/dssvueum/math-and-statistical-functions/time-functions
image

Please see below.

**Additional context**
We have several dams on event reporting, so it’s timeseries will have ~15minutes. At some of those dams, we use the downstream gage’s flow values. Whenever a pool elevation data point comes in, it will use a DSSMATH function (TTSI, first screenshot) to interpolate the downstream gage’s flow values based on the timestamps of the pool elevations, so that we can get pool elevation and flow values at the same timestamps (second screenshot). Subsequent DSSMath functions then generate storage and then inflow values at ~15minutes and we finally create a 15minute regular interval inflow dataset.

We use ~15minute because the data is typically regular interval 15minute data when it’s not raining. The DCPs send out a set of 15minute data via GOES every hour. However, when certain thresholds are exceeded as measured by the datalogger at the DCP, it will start sending random transmission as often as every 5 minutes. The point of this is so that we don’t have to wait up to an hour to know the latest data. 

Interpolating to 15 minutes would simplify things, but that would mean a noticeable delay in the 15 minute datasets. Alternatively, we could interpolate to regular 5minute interval, but that would create a lot of filler data points, which seems wasteful in terms of taking up storage space in the CWMS DB. It also means the 5-minute regular interval data will take 3 times longer to process or work with (ie, comes, displaying tables, plots, etc). vs 15-minute regular interval data or even ~15minute data.

image
image

oskarhurst:
## Problem Description

Fixes #314 

## Solution

Created a ToIrregularUsingPattern Algorithm for interpolating data to fit a pattern.

## how you tested the change

Ran tests by running Algorithm on Sample test data sets and achieved expected results.

## Where the following done:


MikeNeilson (commented):
Looks reasonable, can you provide sample input/output data for a few timeseries?

Similar to how the existing integration tests are structured, but you don't need to actually write the test. We can take care of that later, but having the sample data would be helpful for review.

src/main/java/decodes/tsdb/algo/ToIrregularUsingPattern.java
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	/**Interpolation helper function
+	 * Date Pattern data - data time which you want to interpolate the value for.
+	 * TimedVariable prev - TimedVariable of the nearest in time previous input variable
+	 * TimedVariable next - TimedVariable of the nearest in time next input variable
+	 * Checks the type of data stored in the input and preforms appropriated interpolation between two points.
+	**/
+	void interpolate(Date patternDate, TimedVariable prev, TimedVariable next){
+		CwmsTsId tsId = (CwmsTsId) inputTS.getTimeSeriesIdentifier();
```
MikeNeilson, line 127: Remind me why this is currently CWMS specific. Did the TimeSeriesIdentifier not provide sufficient information?

oskarhurst: only CwmsTsId have the variable ParamType (Inst, Total, Ave) which is needed to check what type of interpolation we need to preform. Do you recommend I add a check to the algorithm to see if the input is from Cwms and if it is not use default linear Interpolation.

oskarhurst: Ok I will make sure to do that.

MikeNeilson: @agilmore2 does HDB provide equivalent information in some way?
e.g.  Is the data already an average, instantatenous, a total between interverals, etc, but just using different names in the code?

MikeNeilson (approved):

MikeNeilson (approved):

PR #631, opened 2024-06-28
fixes issue #338 using Regular expressions

### LINKED ISSUE #338: Ability to Scrape HTML pages and parse data through OpenDCS
The abstract web pull in OpenDCS allows you to read web pages and parse data for multiple gages if the web page is consistent and readable (i.e., CSV).  It may be able to read HTML and parse but there's not enough documentation/examples out there for me to figure that out.  We have one web page that we have to scrape and look for keywords to parse data, and to my knowledge there's not a way to do that through OpenDCS.  If there is please let me know!

A way to scrape an HTML page and look for data, storing it into CWMS time series.

We currently do this with a java program written by @hoffmanjs that runs on the cron on our T7 server.  Our understanding from what we've been told about the cloud migration is that these jobs cannot be there in CWBI, and we need to look at alternative ways to get the data.

**Additional context**
The program visits the URL (https://cnppid.com/wp-content/uploads/2016/06/lakeRiverData.html) and scrapes Lake McConaughy data, Keystone diversion data, and North Platte RIver at North Platte data.  This website is the only automated way to ingest this data - the other data on the site we are pulling from its source (NE DNR).

ktarbet:
Fixes #338 

## Solution

added new regex decodes function


src/main/java/decodes/decoder/RegexFunction.java
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	@Override
+	public void execute(DataOperations dd, DecodedMessage decmsg) throws DecoderException
+	{
+		log.trace("Executing with args '" + argString + "'");
+		this.sensorNumber = getSensorNumber();
+
+		Pattern pattern = Pattern.compile(this.argString);
+		Matcher matcher = pattern.matcher(dd.getRawMessage().toString());
```
MikeNeilson, line 45: this should probably track with where it is in the message vs always using the whole thing but that will make the sample location below more difficult to determine and we can probably handle it as part of future work.

Unless the user has setup a command to do so, the general idea is that the functions "consume" the text they've dealt with.

MikeNeilson (approved):

MikeNeilson (approved):

PR #634, opened 2024-07-02
explicitly include opendcs-annotations.jar in class path in batch/bash files

ktarbet:
fixes error some developers are getting:

```
  java.lang.NoClassDefFoundError: org/opendcs/annotations/PropertySpec
```

### DIFF
--- a/install/bin/decj
+++ b/install/bin/decj
@@ -48,7 +48,7 @@ if [ ! -d "$DCSTOOL_USERDIR" ]; then
 fi
 
 # Build classpath
-CP=$DH/bin/opendcs.jar:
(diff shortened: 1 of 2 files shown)


MikeNeilson (changes requested):
Something else is wrong, this file should be present in the DCSTOOL_HOME/dep directory.

See https://github.com/opendcs/opendcs/blob/28673064894655cb319105321cad3f0d6a61bdc8/build.xml#L375

ktarbet:
Thanks @MikeNeilson ,  I thought that extra Jar went into bin , so my primitives developer environment wasn't getting updated properly.

PR #636, opened 2024-07-07
Refactor/513 switch to jaxrs

### LINKED ISSUE #513: Move LRGS status interface to JAX-RS to align with opendcs/rest_api
`org.opendcs.lrgs.http./LrgsHttpInterface.java`

**Purpose of refactoring**
There was a desire to establish a quick initial LRGS "LrgsInputInterface" that could handle a simple HTTP status request.
This one done before the creation and release of source of the https://github.com/opendcs/rest_api which uses JAX-RS.

Before adding additional functionality to this interface we should align with the REST API to keep the number of dependencies down.

**Proposed method**
Move to few existing handlers to JAX-RS and establish the appropriate setup.

**Additional context**

MikeNeilson:
## Problem Description

Fixes #513 . 

## Solution

Removed Javalin, added in Jetty directly and the jersey implementation of JAX-RS.
This aligns the main project with the choice of API that the https://github.com/opendcs/rest_api made.

## how you tested the change

Manually. LRGS setup with no data /health correctly returns "inactive" and HTTP 503, with data returns "active" and HTTP 200 OK.
/status returns the appropriate "lrgs status" page correctly in either case.

Jetty starts and stops correctly.

## Where the following done:

### DIFF
--- a/ivy.xml
+++ b/ivy.xml
@@ -79,7 +79,18 @@
         </dependency>
         <dependency org="mil.army.usace.hec" name="hec-cwms-ratings-core" rev="1.1.0"/>
         <dependency org="org.slf4j" name="slf4j-api" rev="${slf4j.version}"/>
-        <dependency org="io.javalin" name="javalin" rev="4.6.8" />
+
+        <dependency org="org.eclipse.jetty" name="jetty-jsp" rev="${jetty.jsp.version}"/>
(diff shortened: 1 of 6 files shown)


ktarbet (approved):

MikeNeilson:
@adamkorynta For reference. Stole a few things from the rest_api but you might want to consider what I did with the ServerContext attributes. Notice the "DbInterface" is a global singleton which would be good to avoid whenever possible.

PR #639, opened 2024-07-08
Refactor/property edit panel to mvc

MikeNeilson:
## Problem Description

Our GUIs do not provide a proper seperation of concerns. This hasn't generally been an issue but now that we've introduced the SPI system for properties tools like the eclipse WindowBuilder can't reload the design which makes tweaking gui elements difficult (personally I'd prefer not to do that by hand).

## Solution

I took the PropertiesEditPanel and converted it to more of a control that uses a model-view-controller architecture (maybe model-view-presenter... frankly I've decided not to get too hung up on details so long as there's a reasonable separation of concerns.)
This particular Panel wasn't affected by the SPI issue and was chosen due to a reasonable balance of complexity and simplicity to demonstrate the conversion.

Doing this separation of concerns also made it easier to know when and where to use SwingWorkerThreads. Again not nearly as critical for this panel, but the purpose was to have a reference available for more complex interfaces to be converted.

## how you tested the change

Manually. Interacted with the various panels that used PropertiesEditPanel to verify operations remain unchanged.

## Where the following done:

### DIFF
--- a/src/main/java/decodes/dbeditor/ConfigSensorEditDialog.java
+++ b/src/main/java/decodes/dbeditor/ConfigSensorEditDialog.java
@@ -35,6 +35,7 @@ import decodes.db.Database;
 import decodes.util.TimeOfDay;
 import decodes.db.Constants;
 import decodes.gui.GuiDialog;
+import decodes.gui.properties.PropertiesEditPanelController;
 
 /**
 Dialog for editing a configuration sensor.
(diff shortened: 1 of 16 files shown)


ktarbet:
I'm testing with this branch, and found a property I can't edit.    I see the same behavior on the default branch. Perhaps some properties can't be edited?

image

src/main/java/decodes/dbeditor/ConfigSensorEditDialog.java
```diff
@@ -154,9 +155,9 @@ public ConfigSensorEditDialog(ConfigSensor cs, boolean csNewlyAdded)
 			 && theSensor.getUsgsStatCode() == null)
 				theSensor.setUsgsStatCode(sc); 
 
-		    propertiesEditPanel = new PropertiesEditPanel(theProperties);
+		    propertiesEditPanel = PropertiesEditPanel.from(theProperties);
```
agilmore2, line 158: Can we make a global fix so our indentations are standardized?

MikeNeilson: Wow, yeah. didn't realize how out of whack that file was.

agilmore2 (approved):
Haven't taken time to test, yay formatting!

MikeNeilson:
Well, this panel is used *everywhere* so you'll definitely test it the next time you run the main branch.

PR #640, opened 2024-07-08
using Annotations for BigAdder 

ktarbet:


src/main/java/decodes/tsdb/algo/BigAdder.java
```diff
@@ -1,123 +1,154 @@
...
+	public double input17;	
+	@Input
+	public double input18;	
+	@Input
+	public double input19;	
+	@Input
+	public double input20;	
 
-//AW:LOCALVARS
 	String alg_ver = "1.0.01";
```
agilmore2, line 64: Probably should bump this, if it's even used?

src/main/java/decodes/tsdb/algo/BigAdder.java
```diff
@@ -1,123 +1,154 @@
...
 
-//AW:OUTPUTS
+	@Output
 	public NamedVariable output = new NamedVariable("output", 0);
-	String _outputNames[] = { "output" };
-//AW:OUTPUTS_END
 
-//AW:PROPERTIES
-        public double input_constant = 0.0;
+	@org.opendcs.annotations.PropertySpec(value="0.0")
```
agilmore2, line 69: Import this annotation?

src/main/java/decodes/tsdb/algo/BigAdder.java
```diff
@@ -1,123 +1,154 @@
...
+		description ="Takes up to 20 input values labeled input1 ... input20. Multiplies\n" +
+				"them by coefficients supplied in properties coeff1 ... coeff20.\n" +
+				"Adds them together and produces a single output labeled 'output'.\n" +
+				"The property input_constant is added to the ouput result.\n" +
+				"input_constant defaults to 0 if not supplied\n" +
+				"Values not assigned by computation are ignored.\n" +
+				"All coefficients default to 1.0 if not supplied." )
 public class BigAdder extends decodes.tsdb.algo.AW_AlgorithmBase
 {
-//AW:INPUTS
```
agilmore2, line 31: Does the AlgoEdit tool support annotations now, and handle code concatenations without these comments?

agilmore2 (changes requested):

src/main/java/decodes/tsdb/algo/BigAdder.java
MikeNeilson: One can, but since I used the PropertySpec name again over, it's better to use the fully qualified name for now. Eventually all that info and usage will go this this class definition, but it was too much work to do now, so it stays looking a bit funky.

src/main/java/decodes/tsdb/algo/BigAdder.java
MikeNeilson: No, nor am I inclined to do so. What these do do is make it easier to deal with the algorithms in a normal IDE.

If people really push for it, we can make improvements, but for creating new algorithms, VSCode and maybe a simple extension to help is likely to be a lot better of an option.

src/main/java/decodes/tsdb/algo/BigAdder.java
ktarbet: done

src/main/java/decodes/tsdb/algo/BigAdder.java
agilmore2: I guess I missed the deprecation of the algoedit tool. It's not the first time that a federal agency wanted a simplified interface for coding water resource algorithms then never used it.

agilmore2 (approved):

PR #641, opened 2024-07-08
more algorithm annotations

ktarbet:
## Problem Description

Adding more Annotations to Algorithms

One case I could not drop the  _inputNames[] and String _outputNames[] because that are used in the actual algo.  So I kept the associated get____ methods;

### DIFF
--- a/src/main/java/decodes/tsdb/algo/ExpressionParserAlgorithm.java
+++ b/src/main/java/decodes/tsdb/algo/ExpressionParserAlgorithm.java
@@ -9,7 +9,6 @@ import ilex.util.StringPair;
 import ilex.var.IFlags;
 import ilex.var.NamedVariable;
 import org.nfunk.jep.SymbolTable;
-
 import org.opendcs.annotations.algorithm.Algorithm;
 import org.opendcs.annotations.algorithm.Input;
 import org.opendcs.annotations.algorithm.Output;
--- a/src/main/java/decodes/tsdb/algo/ExpressionParserAlgorithm.java
+++ b/src/main/java/decodes/tsdb/algo/ExpressionParserAlgorithm.java
@@ -1,61 +1,47 @@
 package decodes.tsdb.algo;
 
+import decodes.tsdb.DbCompException;
+import decodes.tsdb.ParmRef;
+import decodes.tsdb.TimeSeriesIdentifier;
+import decodes.tsdb.VarFlags;
+import decodes.tsdb.algo.jep.JepContext;
+import ilex.util.StringPair;
+import ilex.var.IFlags;
+import ilex.var.NamedVariable;
+import jnr.ffi.annotations.In;
+import org.nfunk.jep.SymbolTable;
+
+import org.opendcs.annotations.algorithm.Algorithm;
+import org.opendcs.annotations.algorithm.Input;
+import org.opendcs.annotations.algorithm.Output;
+
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
-import java.util.Date;
-import java.util.Observable;
-import java.util.Observer;
 import java.util.Properties;
 
-import org.nfunk.jep.SymbolTable;
-
-import ilex.var.IFlags;
-import ilex.var.NamedVariableList;
-import ilex.var.NamedVariable;
-import decodes.db.Constants;
-import decodes.db.SiteName;
-import decodes.tsdb.DbAlgorithmExecutive;
-import decodes.tsdb.DbCompException;
-import decodes.tsdb.DbIoException;
-import decodes.tsdb.VarFlags;
-import decodes.tsdb.algo.AWAlgoType;
-import decodes.tsdb.algo.jep.JepContext;
-import decodes.tsdb.CTimeSeries;
-import decodes.tsdb.ParmRef;
-import ilex.var.TimedVariable;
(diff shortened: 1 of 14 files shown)


src/main/java/decodes/tsdb/algo/ExpressionParserAlgorithm.java
```diff
@@ -429,12 +404,8 @@ else if (out1value instanceof String)
 			else if (out2value instanceof String)
 				setOutput(out2, (String)out2value);
 		}
-//AW:AFTER_TIMESLICES_END
 	}
 
-	/**
-	 * Required method returns a list of all input time series names.
-	 */
 	public String[] getInputNames()
```
MikeNeilson, line 409: This method is defined in the super class, you can remove this and the array and just call this.getInputNames when needed, or cache it during beforeTimeSlices as appropriate.

src/main/java/decodes/tsdb/algo/ExpressionParserAlgorithm.java
```diff
@@ -1,61 +1,47 @@
...
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.Date;
-import java.util.Observable;
-import java.util.Observer;
-import java.util.Properties;
-
-import org.nfunk.jep.SymbolTable;
```
MikeNeilson, line 11: and this still compiles?

ktarbet: I used my IDE to cleanup the imports, import org.nfunk.jep.SymbolTable; is still included.

src/main/java/decodes/tsdb/algo/ExpressionParserAlgorithm.java
ktarbet: What do you think about caching in the super class;  the inputs, and outputs, should not be changing should they?

src/main/java/decodes/tsdb/algo/HdbEvaporation.java
```diff
@@ -1,61 +1,38 @@
...
+		"Primary value is \"area\".\n" +
+		"Really need average area, so two choices for second input:\n" +
+		"prevArea = previous EOP area or diffArea = deltaArea\n" +
+		"For now, chose prevArea.\n" +
+		"Third input is evapCoeff. This is a SDI pointer to either\n" +
+		"a coefficient timeseries, or to lookup a coefficient from the stat tables\n" +
+		"Output is evaporation as a volume.\n" +
+		"Properties include: \n" +
+		"\n" +
+		"ignoreTimeSeries - completely ignore changes to any timeseries value from evapCoeff, and always lookup from database.\n" +
```
ktarbet, line 19: I removed the Html markup previously used for Javadoc's, but not inside annotation;  I don't think the text box will render Html?

ktarbet: We could keep javadocs also, but that feels like duplication.

MikeNeilson: We could also make that box render HTML; Swing has some support for the basics. But we'd need some way identify that it is HTML.

MikeNeilson (approved):

src/main/java/decodes/tsdb/algo/HdbEvaporation.java
ktarbet: I'll keep the HTML for future rendering if HTML is detected, I don't think Html will hurt in text rendering.  Many algos only have text in the javadocs

ktarbet:
I investigated adding Annotations to PythonAlgorithm.java, but It seems the Inputs and outputs are defined elsewhere for a using or creating a custom Python Algorithm.

MikeNeilson (approved):

ktarbet:
A few Algorithms that I've seen seem to have two layers of Properties.  Do we need both?

* the String _propertyNames[]  (that can be replaced with @org.opendcs.annotations.PropertySpec(value="?")
* protected PropertySpec[] getAlgoPropertySpecs() -- that is defined separately but similar to the annotations. but is explicit on the data type for example: PropertySpec.BOOLEAN

Specific Example  :  PeriodToDate.java

```java
private PropertySpec ratingPropertySpecs[] = 
	{
		new PropertySpec("goodQualityOnly", PropertySpec.BOOLEAN,
			"(default=false) Only include good quality values in the calculation.")
	};
	@Override
	protected PropertySpec[] getAlgoPropertySpecs()
	{
		return ratingPropertySpecs;
	}

	@org.opendcs.annotations.PropertySpec(value="false")
	public boolean goodQualityOnly = false;
```

src/main/java/decodes/tsdb/algo/ExpressionParserAlgorithm.java
MikeNeilson: We'll have to think about that. so long as it's an instance variable that should work without issue, I think.

MikeNeilson:
> I investigated adding Annotations to PythonAlgorithm.java, but It seems the Inputs and outputs are defined elsewhere for a using or creating a custom Python Algorithm.

Yeah, that's not super trivial. You *can* add Java annotations to Jython code, but we don't structure our Python Algorithms in a way that would work.

PR #652, opened 2024-07-17
Refactor/649 tsdao find

### LINKED ISSUE #496: Improve performance of Group processing. and Bind Vars
GroupHelper, TsDbGroupDao, etc

**Purpose of refactoring**
- Code does not properly use bind variables
- Performance on large groups is slow
- Can likely benefit from parallelism.

**Proposed method**
- Review code and at a minimum move all queries to using bind vars and the new handlers.
- Investigate structure and improve performance if possible
- at a minimum identify future areas of parallelism. Each group is independent so processing can be done separately
  - If general restructuring results in sufficient performance this would not be required.

**Additional context**
This is an area where JOINs are likely the only solution to improve performance to avoid nested queries. If that is the case JDBI will be used directly to assist with that as we are already planning it's introduction into the code base.

### LINKED ISSUE #649: TimeSeries DAO needs a findTimeseriesByUnique that doesn't throw exception
TimeSeriesDAI and implementations.

**Purpose of refactoring**

The transformTsId loop can call for a lot of timeseries that don't exist, most likely because they need some sort of minor purmutation.
getTimeseriesIdentifiery(String Unique) throws NoSuchObjectException, in the case of USACE we have at least one group that is 14,000 timeseries long, dealing with this many exceptions simply takes a long time and in the case of the function is not "exceptional."

**Proposed method**
Create a findTimeSeries(UniqueString) that returns an Optional<TimeSeriesIdentifier>, the origin getTimeSeriesIdentifier can then use to throw the exception, but points of code for which the Identifier not existing is normal can use the find.

**Additional context**
I waited ~20 minutes and the full list still hadn't been populated even after reducing the logging.
I think I will also hide some of the Cache missing log statements behind a feature flag, or maybe just print the rate, since that's also fairly normal behavior for a cache.

MikeNeilson:
## Problem Description

Fixes #649.
Fixes #496  

## Solution

- A combination of allowing less drastic error handling
  - returning a (result,failure) pair so that users of the result can determine if it's a normal condition or something more critical.
- Fixing some SQL usage
  - Using a transaction to use only one connection.
  - Fixing the transaction DAO connection closing logic so it's correctly closed
  - Fixing withTransaction to use the correct java.sql.Connection (original design was always new, however that breaks using the same connection if used in a transaction (note: I think I'm going to need some additional logic here as that behavior is otherwise desired)
- Tweaking the behavior of a time series identifier "cache" miss to no go to the database. The group in question or #496 starts with 14,000+ time series, but only ~71 are needed for the list, so there's 14,000 calls to the database to decide the time series identifier actually doesn't exist... about 2 seconds after we just pulled every time series identifier into a cache. Took me a while to realize that, but once I saw it it seemed, crazy.

I'm not super happy with the primary logic of just "cache updated within" but I think the cache design needs to be expanded more before that's addressed in a better way. I'll have to rebase #626 on to this if merged, but I think worth the hassle and some of the improvements I made to lend themselves to providing say, the last time this was searched for and missed and actually having the cache, through callback, do the search to track that. But definitely beyond scope of this PR, which is already a bit unwieldy.

It's entirely plausible the cache behavior tweak at the end is the only thing required, it would thusly bypass a lot of exception, but I think this work overall is worth the effort and a few hidden bugs need to be fixed.

NOTE: I haven't implemented that "cache miss is actually because data missing" tweak in the OpenDCS reference impl or HDB. But easy to do so if @agilmore2 thinks good for HDB.
Will likely do so for OpenDCS, but don't have a sufficiently large DB at hand to test and measure improvements.

I do want to do some additional testing, but I think the logic is reasonable and it's certainly time for some review and sanity checks.

## how you tested the change

Integration tests cover existing behavior. Manually to time results.

- Original performance was 40+ minutes.
- Removing excess exception throwing brought it 30 minutes.
- Reducing db connection opens, and fixing transaction connection handling, brought it to 10 minutes.
- Telling CwmsTsDao to assume the "miss" is correct when it polls the cache if the last load is recent, brought it to 11 seconds.

VisualVM's sampler and profiler were used to determine hotspots and where in the loops things were taking too long.

## Where the following done:

### DIFF
--- a/src/main/java/decodes/cwms/CwmsConnectionPool.java
+++ b/src/main/java/decodes/cwms/CwmsConnectionPool.java
@@ -20,6 +20,8 @@ import java.util.logging.Logger;
 import javax.management.JMException;
 import javax.management.ObjectName;
 import javax.management.openmbean.OpenDataException;
+
+import org.eclipse.jetty.server.Authentication.Wrapped;
 import org.opendcs.jmx.ConnectionPoolMXBean;
 import org.opendcs.jmx.WrappedConnectionMBean;
 import org.opendcs.utils.sql.SqlSettings;
@@ -305,7 +307,7 @@ public final class CwmsConnectionPool implements ConnectionPoolMXBean
      */
     public synchronized void returnConnection(Connection conn) throws SQLException
     {
-        if(connectionsOut.contains(conn))
+        if (connectionsOut.contains(conn))
         {
             // the lambda handler above handles the count
             connectionsFreed++;
@@ -317,10 +319,16 @@ public final class CwmsConnectionPool implements ConnectionPoolMXBean
         {
             log.warning("Unknown connection returned to my pool.");
             unknownConnReturned++;
-            if(SqlSettings.TRACE_CONNECTIONS)
+            if (SqlSettings.TRACE_CONNECTIONS && !(conn instanceof WrappedConnection))
             {            
                 log.warning("Connection is from");
                 logStackTrace(log,Level.WARNING,Thread.currentThread().getStackTrace(),BEFORE_CUR_THREAD_STACK_CALL+1);
+            }
+            else if (SqlSettings.TRACE_CONNECTIONS)
+            {
+                final WrappedConnection wc = (WrappedConnection)conn;
+                log.warning(() -> "Connection opened from: " + String.join(System.lineSeparator(), wc.getOpenStackTrace()));
+                log.warning(() -> "Connection closed at: " + String.join(System.lineSeparator(), wc.getClosedStackTrace()));
             }
         }
     }
--- a/src/main/java/decodes/cwms/CwmsConnectionPool.java
+++ b/src/main/java/decodes/cwms/CwmsConnectionPool.java
@@ -21,7 +21,6 @@ import javax.management.JMException;
 import javax.management.ObjectName;
 import javax.management.openmbean.OpenDataException;
 
-import org.eclipse.jetty.server.Authentication.Wrapped;
 import org.opendcs.jmx.ConnectionPoolMXBean;
 import org.opendcs.jmx.WrappedConnectionMBean;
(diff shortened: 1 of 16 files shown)


ktarbet:
I have a manual test of Loading the Computation Editor. Here are the results before/after using this branch.

|                    | Before   | After   |
|--------------------|----------|---------|
| connectionsRequested | 240      | 23      |
| Time elapsed        | 17 seconds | 8 seconds |

src/main/java/decodes/cwms/CwmsConnectionPool.java
```diff
@@ -20,6 +20,8 @@
 import javax.management.JMException;
 import javax.management.ObjectName;
 import javax.management.openmbean.OpenDataException;
+
+import org.eclipse.jetty.server.Authentication.Wrapped;
```
ktarbet, line 24:
```suggestion
```

src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
```diff
@@ -16,9 +16,12 @@
 import java.util.Date;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
```
ktarbet, line 19:
```suggestion
```

src/main/java/opendcs/util/sql/WrappedConnection.java
```diff
@@ -58,6 +59,14 @@ public class WrappedConnection implements Connection, WrappedConnectionMBean
     private final Thread openingThread;
 
 
+    private Predicate<String> stackTraceFilter = (l) ->
```
ktarbet, line 62:
```suggestion
    // limits stack trace to OpenDCS and java code (filters out other dependencies)
    private Predicate<String> stackTraceFilter = (l) ->
```

ktarbet (approved):

PR #659, opened 2024-07-23
Bugfix/325 sched entry dao

### LINKED ISSUE #325: [JDBC] Schedule Entry status update and write
`opendcs.dao.ScheduleEntryDAO::writeScheduleStatus`

**Purpose of refactoring**
Query is indentified as the 10th most used affecting overall performance.

**Proposed method**
Migrate to new handlers for SQL or use prepared statements and bind variables directly.

**Additional context**
There are a lot of possible nulls in this data set.

MikeNeilson:
## Problem Description

Fixes #325. 

## Solution

Migrate SQL to new usage.
Delete operations are within transactions.

## how you tested the change

Create integration test for DAO delete operations. Other operations are covered by existing tests.

## Where the following done:

### DIFF
--- a/src/main/java/opendcs/dao/ScheduleEntryDAO.java
+++ b/src/main/java/opendcs/dao/ScheduleEntryDAO.java
@@ -1,6 +1,6 @@
 /**
  * $Id$
- * 
+ *
  * $Log$
  * Revision 1.13  2017/06/16 15:33:18  mmaloney
  * To handle import from XML, when writing, if loading app ID is null but the name is not,
@@ -39,9 +39,9 @@
  * Revision 1.2  2014/07/03 12:53:41  mmaloney
  * debug improvements.
  *
- * 
+ *
  * This software was written by Cove Software, LLC ("COVE") under contract
- * to the United States Government. No warranty is provided or implied other 
+ * to the United States Government. No warranty is provided or implied other
  * than specific contractual terms between COVE and the U.S. Government.
  *
  * Copyright 2014 U.S. Army Corps of Engineers, Hydrologic Engineering Center.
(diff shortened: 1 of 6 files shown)


src/main/java/opendcs/dao/ScheduleEntryDAO.java
```diff
@@ -56,489 +56,596 @@
...
+            }
+        }
+
+
+        scheduleEntry.setLastModified(new Date());
+        // It might be an import from an xml file. If no key, try to lookup from name.
+        if (scheduleEntry.getKey().isNull())
+        {
+            String q = "select schedule_entry_id from schedule_entry where "
+                + " upper(name) = upper(?)";// + sqlString(scheduleEntry.getName().toUpperCase());
```
ktarbet, line 302:
```suggestion
                + " upper(name) = upper(?)";
```

src/main/java/opendcs/dao/ScheduleEntryDAO.java
```diff
@@ -56,489 +56,596 @@
...
+			parameters.add(scheduleEntry.getRoutingSpecId());
+			parameters.add(scheduleEntry.getStartTime());
+			parameters.add(scheduleEntry.getTimezone());
+			parameters.add(scheduleEntry.getRunInterval());
+			parameters.add(scheduleEntry.isEnabled());
+			parameters.add(scheduleEntry.getLastModified());
+        }
+        else // do an update
+        {
+            q.append("update schedule_entry set ")
```
ktarbet, line 335:
```suggestion
            q.setLength(0);
            q.append("update schedule_entry set ")
```

src/main/java/opendcs/dao/ScheduleEntryDAO.java
```diff
@@ -56,489 +56,596 @@
...
+
+        scheduleEntry.setLastModified(new Date());
+        // It might be an import from an xml file. If no key, try to lookup from name.
+        if (scheduleEntry.getKey().isNull())
+        {
+            String q = "select schedule_entry_id from schedule_entry where "
+                + " upper(name) = upper(?)";// + sqlString(scheduleEntry.getName().toUpperCase());
+            try
+            {
+				doQuery(q, rs -> scheduleEntry.forceSetId(DbKey.createDbKey(rs, 1)), scheduleEntry.getName());
```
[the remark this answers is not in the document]
MikeNeilson: The `ScheduledEntry` provided to this function should have a name already but it may be from a reference that wasn't loaded from the database so this looks up the database id for that name. and then assigns the id.

This is then used to determine if the entry is an update or new entry below. Might be a good idea to put check for a valid name at the top of the method though just to make that clear.

src/main/java/opendcs/dao/ScheduleEntryDAO.java
```diff
@@ -56,489 +56,596 @@
...
+				q.setLength(0);
+				q.append("delete from platform_status where last_schedule_entry_status_id in "
+					+ "(select schedule_entry_status_id from schedule_entry_status "
+					+ "where schedule_entry_id = ?)");
+				doModify(q.toString(), scheduleEntry.getKey());
+
+				q.setLength(0);
+				q.append("delete from schedule_entry_status where schedule_entry_id = ?");
+				doModify(q.toString(), scheduleEntry.getKey());
+			});
```
ktarbet, line 606: scheduleEntry.getId() and scheduleEntry.getKey() are used consistently with before changes. 

However,  If I'm reading correctly they may be the same, perhaps one should be deprecated/preferred.

MikeNeilson: They're identical. I like getId better, but I'll add deprecated in a different PR. Pretty sure they're identical but doesn't hurt to do a scan of the code first to verify.

src/main/java/opendcs/util/sql/ConnectionInTransaction.java
```diff
@@ -13,8 +13,18 @@ public ConnectionInTransaction(Connection realConnection)
         super(realConnection);
     }
 
+    /**
+     * So long as the actual state isn't changing, this can be a noop.
```
ktarbet, line 17:
```suggestion
     * Throws an Exception if the state would change, because we are in a transaction.
     * So long as the actual state isn't changing, this can be a noop.
```

ktarbet (dismissed):

ktarbet (approved):

PR #661, opened 2024-07-23
Feature/129 csv set delimiter

### LINKED ISSUE #129: Adding Capability to Define Delimiter in csv function
The current methodology of the csv function assumes a comma delimiter, and the function allows you to easily move through columns and decode or skip them.  

Many of the web pulls are using tab or space delimiters, and the csv function would allow for much more efficient and easier decoding if it allowed us to define the delimiter.

Current decoding methodologies require me to check at the beginning of each column for missing data before moving forward, either with a new format statement if the column is blank, or with decoding that piece of data.  It results in many more format statements.

MikeNeilson:
## Problem Description

Fixes #129 . 

## Solution

Allow user to set a delimiter. Default is ",".

## how you tested the change

Additional decodes tests were added with the script and input set to the delimiter.

## Where the following done:


MikeNeilson:
@hoffmanjs, FYI.

docs/source/start-decoding.rst
```diff
@@ -87,11 +87,17 @@ The following table includes the csv parser operation.  This command is
...
-+======================+==========================================================+
-| csv(sens#,...)       | Parse a series of comma separated values                 |
-+----------------------+----------------------------------------------------------+
++-----------------------------+------------------------------------------------------------+
+| **Command**                 | **Description**                                            |
++=============================+============================================================+
+| csv(sens#,...)              | Parse a series of comma separated values                   |
++-----------------------------+------------------------------------------------------------+
+| csv(delimiter=\|,sens#,...) | Parse a series of values separated by a specific delimiter |
+|                             | In this example the pipe, \|, character is used.           |
```
agilmore2, line 96: remove commas and add either single or double quotes around the sample delimiter in the text like you do for the whitespace characters below.

src/main/java/decodes/decoder/CsvFunction.java
```diff
@@ -53,12 +54,12 @@ public void execute(DataOperations dd, DecodedMessage decmsg) throws DecoderExce
...
 					break;
 				}
-				if (c == ',')
+				if (dd.checkString(delimiter))
 				{
 					processColumn(col++, sb.toString().trim(), decmsg, dd.getCurrentLine(),
 						fieldStart, dd);
 					sb.setLength(0);
-					fieldStart = dd.getBytePos()+1;
+					fieldStart = dd.getBytePos()+delimiter.length();
```
agilmore2, line 62: how does a "delimiter can be more than one character" work here? Are the multiple characters in the delimiter string treated as set of options the delimiter could be, or each field is delimited by all the listed characters?

agilmore2 (dismissed):

src/main/java/decodes/decoder/CsvFunction.java
MikeNeilson: Basically you can require that the delimiter be a sequence of characters, or a word. E.g. if you use "bob" as the delimiter all the items are separated by the word "bob". I'll update the docs to make that more clear.

it's not done a lot but I've done it in the past for certain sets of data.

KEYS
ktarbet, line 1: I'm not sure if adding this KEY is intentional.?

src/main/java/decodes/decoder/CsvFunction.java
```diff
@@ -137,14 +138,30 @@ public void setArguments(String argString, DecodesScript script) throws ScriptFo
 		{
 			String t = st.nextToken();
 			t = t.trim();
-			try
+			if (t.startsWith("delimiter"))
 			{
-				int sensorNum = Integer.parseInt(t);
-				sensorNumbers.add(sensorNum);
+				String[] parts = t.split("=");
+				this.delimiter = parts[1];
```
ktarbet, line 144:
```suggestion
				if( parts.length ==2 )
				{
					this.delimiter = parts[1];
				}
```

src/test/resources/decodes/db/CsvPipeDelimiterDocSample.input
```diff
@@ -0,0 +1,4 @@
+ignored header line
+03/01/2014 12:00:00 23.95| 17.2| 8252| 0| 0| 98.1
```
ktarbet, line 2: why not a pipe delimiter after the time-stamp; before the value?

KEYS
MikeNeilson: Well, it was, but shouldn't have been in this branch. I've gone and rebased

src/test/resources/decodes/db/CsvPipeDelimiterDocSample.input
MikeNeilson: it appears I've simply forgotten it... though now I need to check the csv I copied that from.

MikeNeilson: Okay, just checked the decodes script processed the date first and then passes the remaining bits to the CSV, so the CSV function only starts after the date.

docs/source/start-decoding.rst
MikeNeilson: Took me a minute to figure out what was actually wrong but good call.

ktarbet (approved):

PR #667, opened 2024-08-02
Create ServiceProvider Interface for DecodesFunctions.

### LINKED ISSUE #218: DecodesFunctions moved to SPI
Custom functions within DecodesScripts

**Purpose of refactoring**
Easier initialization, Future expansion.

**Proposed method**
SPI mechanism. This allows several benefits:
- Current manual factory no longer required
- Users can add Functions without altering base code. (for example the Nos6min and NosHourly could be maintained on a separate development cycle in another repo.)

**Additional context**
There's going to be more decodes functions in the future. Let's just make it easier for ourselves.

MikeNeilson:
## Problem Description

Fixes #218 . 
Seeing as we're starting to create some more decodes functions seems like we should get this in sooner rather than later.

## Solution

Create a function provider, and alter FunctionsList class to load available on first lookup. Implemented Provider class for all current DecodesFunctions.

## how you tested the change

Behavior is covered by existing tests.

## Where the following done:

### DIFF
--- a/docs/source/dev-docs.rst
+++ b/docs/source/dev-docs.rst
@@ -259,7 +259,7 @@ PropertyProvider
 The PropertyProvider system as added to support EnvExpander retrieving values from sources other than the java `System.properties`.
 The mechanism uses the java ServiceProvider mechanism so downstream users can implement any custom sources they need.
 
-To implement a custom implement the following class `org.opendcs.spi.properties.PropertyValueProvider`.
+To implement a custom property provider the following class `org.opendcs.spi.properties.PropertyValueProvider`.
 
 .. code-block:: java
     :linenos:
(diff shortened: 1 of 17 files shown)


docs/source/dev-docs.rst
```diff
@@ -350,6 +350,46 @@ The following prefixes are reserved:
 |file      |Values from files on the file system. |
 +----------+--------------------------------------+
 
+Custom Decodes Functions
+========================
+
+To implement a custom implement the following class `org.opendcs.spi.decodes.DecodesFunctionProvider`. And Derived
```
ktarbet, line 356:
```suggestion
To create a custom function, implement the following interface `org.opendcs.spi.decodes.DecodesFunctionProvider`, and derive
```

src/main/resources/META-INF/services/org.opendcs.spi.decodes.DecodesFunctionProvider
ktarbet, line 1: I don't understand all the benefits of the Service Provider.

If we need to edit the resource file, to add a function, this doesn't seem to meet the original goal:

"Users can add Functions without altering base code"  https://github.com/opendcs/opendcs/issues/218

Adding functions and also adding provider to the list feels like more work than updating the FunctionsList.

MikeNeilson: There is that minor conceit and there are annotation processors that can take care of it, google's AutoService for example, but the benefit here is that an the `META-INF/services/` file is *per jar*, so a downstream user that needs a custom decodes function can just make the jar and it gets picked up. The way jar works you're always going to need something listed out somewhere, this as the benefit of easier user expansion and some type safety.

With the current design, you *must* register in that FunctionList, which a user provided jar would have no ability to do.

MikeNeilson: But basically it's just Inversion of control, the FunctionList is now only responsible for knowing DecodesFunctions exist, not what they all are.

ktarbet (approved):

PR #669, opened 2024-08-05
Fix hdbsite

### LINKED ISSUE #668: HDB siteDAO sitecode string init skipped
HDBSiteDAO only queries sitecode if it's null, and it has a default which only works for one office.

Fix underway.

agilmore2:
## Problem Description

Fixes #668 

## Solution

Added a boolean fence to synchronized string. Please review for best practice.

## how you tested the change

Observed that sitecode is queried now, when before it was not.

## Where the following done:

   - [X] Test procedure descriptions for manual testing

### DIFF
--- a/src/main/java/decodes/hdb/HdbSiteDAO.java
+++ b/src/main/java/decodes/hdb/HdbSiteDAO.java
@@ -31,6 +31,7 @@ public class HdbSiteDAO extends SiteDAO
 	private static HashMap<String, DbKey> stateName2Id = new HashMap<String, DbKey>();
 	private static HashMap<String, DbKey> stateAbbr2Id = new HashMap<String, DbKey>();
 	private static String myDbSiteCode = "ECO";
+	private static boolean siteCodeInit = false;
 	
 	public HdbSiteDAO(DatabaseConnectionOwner tsdb)
 	{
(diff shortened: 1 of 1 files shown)


src/main/java/decodes/hdb/HdbSiteDAO.java
```diff
@@ -102,7 +103,7 @@ public HdbSiteDAO(DatabaseConnectionOwner tsdb)
 			}
 		}
 		
-		if (myDbSiteCode == null)
+		if (!siteCodeInit)
```
MikeNeilson, line 106: AtomicBoolean *might* be better here, but given the synchronized block it's probably fine like this.

Though I'd move the synchronized to either siteCodeInit or at least before siteCodeInit; there's a small chance of a race condition or sequencing issue *if* this gets called again.

MikeNeilson (approved):

src/main/java/decodes/hdb/HdbSiteDAO.java
agilmore2: If you're good as is, I'll merge. I would like to know more about how a race condition could be an issue here. Perhaps a deep discussion for September.

MikeNeilson: Hard to say really, it's just that your checking the state of a variable and then synchronizing on a different object after which you set the value of the first. So not really a race condition but if two threads were calling that the both would check and start, though presumably it would get stopped at the synchronization block so maybe not an issue right now.

PR #676, opened 2024-08-16
Fix issue with classpath length in windows.

MikeNeilson:
## Problem Description

Depending on installation location the CLASSPATH variable becomes too long leaving off important dependencies and causing the application start to fail.

## Solution

Use the short DOS 8.3 of each file to keep the CLASSPATH variable smaller.

## how you tested the change

Manually, was able to start applications that were previously unable to start due to install location.
Verified run task still works.

## Where the following done:


DanielTOsborne:
The DOS 8.3 names aren't required and can be disabled on a system.
Wildcards might be a better option: https://docs.oracle.com/javase/8/docs/technotes/tools/windows/classpath.html

MikeNeilson:
I thought of that, but at the moment there's an ordering problem. once we kill the need for lobo and cobra we can do that... but thanks for letting me know, that may be my afternoon project given this information.

Java 9+ also allow a file for parameter for this situation but we've currently committed to supporting java 8. But that was a while ago so I think I'm going to ask the mailing list how they feel about it now.

agilmore2:
Wildcards sound like they would definitely help. Could specify the jars that need ordering before the wildcard for that directory.

MikeNeilson:
That's the problem, the jars need to be order so LoboBrowser and Cobra are last, but with the wildcare there is zero garauntee that it will be stable. So that doesn't work at the moment.

The good news is, those dependencies are only used to process some HTML, but the way that HTML is processed doesn't actually require an HTML parser, it's just getting directory names that match a pattern from any "a href" tags, so fairly trivial to replace thankfully. I may work on that this after noon if I finish my other projects.

PR #678, opened 2024-08-16
Refactor/241 remove lobobrowser AND fix classpath issues

### LINKED ISSUE #241: Replace LoboBrowser components

MikeNeilson:
## Problem Description

Fixes #241. 
Fixes issue with large classpaths by using wild card class path for the two directories.
Fixes issue with determined vs set DCSTOOL_HOME on POSIX system reporting an a different when there really isn't one.

## Solution

Moved block in WebDirectoryDataSource to use simple regex match to find link, verify that at least for my understanding the behavior is correct (I think the original DataSource folder structure has changed.)
Used wildcard classpaths.
Compared determined "realpath" to determine if environment setup is questionable.

Removed LoboBrowser and Cobra dependencies and references thereto.

## how you tested the change
On windows/linux

Verified applications could run from ant run task, shell/cmd, or double clicking the .bat file in windows.

## Where the following done:

### DIFF
--- a/install/bin/decj
+++ b/install/bin/decj
@@ -40,8 +40,8 @@ fi
 export DCSTOOL_USERDIR
 
 if [ ! -d "$DCSTOOL_USERDIR" ]; then
-    echo "Creating Local User Directory and initial properties in ${DCSTOOL_USERDIR}"
-    echo "The default XML database has been copied to this directory."
+    echo "Creating Local User Directory and initial properties in ${DCSTOOL_USERDIR}" 1>&2
+    echo "The default XML database has been copied to this directory." 1>&2
     mkdir -p $DCSTOOL_USERDIR
     cp $DH/decodes.properties $DCSTOOL_USERDIR/user.properties
     cp -r $DH/edit-db $DCSTOOL_USERDIR/edit-db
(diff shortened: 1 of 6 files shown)


MikeNeilson:
Okay, this time I think I've got it.

ktarbet:
I did some basic testing in windows and Linux.   I verified there are two less dependencies in the ./dep directory in a install.:

These jars are gone from the install in this branch, but exist in the master branch.

```bash
./OPENDCS.master/dep/zzzCobra-1.0.2.jar
./OPENDCS.master/dep/zzzLoboBrowser-1.0.0.jar
```

install/bin/decj
```diff
@@ -50,43 +50,9 @@ fi
...
-# then
-#   CP=$CP:$CWMS_JAR_DIR/hec.jar:$CWMS_JAR_DIR/hecData.jar:$CWMS_JAR_DIR/heclib.jar:$CWMS_JAR_DIR/rma.jar:$CWMS_JAR_DIR/lookup.jar
-# fi
-# if [ -n "$CWMS_SYSJAR_DIR" -a -d "$CWMS_SYSJAR_DIR" ]
-# then
-#   CP=$CP:$CWMS_SYSJAR_DIR/jdom.jar:$CWMS_SYSJAR_DIR/jdom-1.0.jar
-# fi
-
-
-if [ ! -z "$CP_SHARED_JAR_DIR" -a -d "$CP_SHARED_JAR_DIR" ]
```
MikeNeilson, line 64: @agilmore2 where these for HDB, the ones above that says CWMS are rather obvious and we don't need them anymore but I'm not sure if I went to far with these.

ktarbet (approved):
The website referenced in the WebDirectoryDataSource dd.weatheroffice.gc.ca - doesn't seem to be active so testing/issue-feedback is being left to anyone using WebDirectoryDataSource.

MikeNeilson:
I'll update the source with the link I used to at least somewhat test it, but I think the general structure changed and this should operate recursively, but I'm going to wait for someone to ask before doing the work.

agilmore2:
I don't see usage of this $CP_SHARED_JAR_DIR in any of the HDB instances,
but it is a documented feature:

https://opendcs-env.readthedocs.io/en/latest/legacy-cp-userguide.html#computation-meta-data

I do like the ability to drop a library of algorithms in a directory and
have them work.

MikeNeilson:
Hmm. Thanks i'll go read the docs. But that behavior is already covered by
the userdir, wouldnt be surprised if this was for those that didnt use it.
But that at least meams it's not the right time to yank it out without
warning.

ktarbet (approved):

PR #681, opened 2024-08-23
Initial work on having one path to create database instances.

### LINKED ISSUE #221: SPI for Database Implementation

### LINKED ISSUE #345: Get all implementations (including XML) properly using Connection Pools/java.sql.DataSource
All JDBC usages.

**Purpose of refactoring**
Holding onto an open connection is an anti-pattern for database usage. Additionally java.sql.Connection implementations are not guaranteed to be thread-safe and should not be shared between threads, which we'd like to do more of.

**Proposed method**
Short term:

- Wrap all usages of a Connection in try-with-resources block
- Tweak DAOs that save instances of other DAOs to open those in a try-with-resource block at time of usage

Mid/Long term:
- Classes that implement `DatabaseConnectionOwner` should provide connections from a pool and expect them back.
  - Additionally there should only be one per implementation.
- Combine the Decodes (`Database`) and Time Series Datasebase (`TimeseriesDb`) into a single `Database` class
- Implement SPI system for database creation (see #221)

**Additional context**
Trying to work around this for just CWMS which uses the pool is causing issues for the other implementations, which need to go about this as well anyways so no time like the present.

Some initial work show it was fairly trivial to create a simple XML/JDBC datasource/connection implementation that just ferried the database directory location to were it was needed.

MikeNeilson:
## Problem Description

Fixes #221. 
Fixes #345.

## Solution

Implement SPI for Database provider and make all database types use a javax.sql.DataSource for the root source of where to go.

## how you tested the change

Existing tests cover behavior (well mostly.)

TODO:
- [ ] test apps manually
- [x] test against CWMS
- [ ] test against HDB

## Where the following done:


MikeNeilson:
Unfortunately I don't think I can make this smaller, though I do think I've made it a bit simpler than another similar PR.

Performance is reasonable against a blank database, need to test against larger, will probably put the "initDecodes" back in play just restructure it a bit, like being instance members instead of an external function.

MikeNeilson:
Okay, so unfortunately this is going to expand to more files. The good news is, instead of all of those files kinda doing their own thing to initialize a database it'll be somewhat centralized. The bad news is that this is done in a rather large number of places.

So I ask that the current work be reviewed as-is (FYI, the apps won't work due to how DecodesInterface was being used). And once primary concerns are addressed I'll continue on.

DecodesInterface is the next thing to tweak, I may move the functions to Database as members, or I may keep that as-is and make the various function take a Database as an argument. But either way may goal is to remove the use of Database::getDb at that level. It's still used too nested in several places but these need to be taken care of at this time for this PR.

No rush though, not going to try and merge this in for 7.0.13.

MikeNeilson:
Okay, so this is definitely working, but it's going to take a lot, it doesn't help the way the Launcher is setup and I plan to refactor those internals during the next release. As much as I want to just barrel through this effort I'm going to force myself to stop for a bit to focus on other things.

It would still be could for everyone to look through and comment on anything you know will, or suspect might, break things in odd ways.

src/main/java/decodes/tsdb/TimeSeriesDb.java
```diff
@@ -576,64 +571,65 @@ public abstract class TimeSeriesDb
     /**
      * No args constructor required.
      */
-    public TimeSeriesDb()
+    public TimeSeriesDb(String appName, javax.sql.DataSource dataSource, DecodesSettings settings) throws DatabaseException
```
ktarbet, line 574: appName:    It feels odd to require TimeSeriesDb, OpenTsDb, CwmsTimeSeriesDb, and HdbTimeSeriesDb to need to be told the name of the application connecting.   I can see that may be needed for computation processing to keep track of different processes.

Could there be a OpenDCSDataSource instead of javax.sql.DataSource has that optional appName/appID/key if required?

MikeNeilson: It's not actually used everywhere, but it's pervasive enough we're stuck with it.

As for renaming, that doesn't make sense to me. the javax.sql.DataSource *is* the connection or connection pool. And just that. The app name really has nothing to do with it, I would make the appname part of the `OpenDcsDatabase` class and tweak how it's used. (Mostly in logging really, which is all going away, but also to lookup properties which should be at that higher level.)

ktarbet (commented):
It is interesting to see the similar methods in Database  aka : decodesDatabase 

initializeForEditing();
initializeForDecoding()
initMinimal(DecodesSettings)

MikeNeilson:
> It is interesting to see the similar methods in Database aka : decodesDatabase
> 
> initializeForEditing(); initializeForDecoding() initMinimal(DecodesSettings)

I'm moving them from DecodesInterface. Haven't been deleted yet due to how they were used, but their removal from DecodesInterface is part of the effort of this PR.

src/main/java/decodes/tsdb/TimeSeriesDb.java
MikeNeilson: So the AppName is used by TsImport, and at least the Postgres database, but I believe HDB as well to mark were a timeseries value was brought in from.

I do think it should go away from this level, but going to take some time to properly move around.

MikeNeilson:
NOTE: I think I'm going to rebase/merge with with the gradle switch. While it will involve a bit of "sillyness" it will be easier to create these new interfaces as a separate project and thus jar, which will help enforce separation going forward.

java/opendcs/src/main/java/decodes/tsdb/ComputationApp.java
```diff
@@ -336,9 +336,9 @@ public void runApp( )
 			while(!shutdownFlag)
 			{
 				Logger.instance().debug3("ComputationApp start of main loop.");
-				try(
-					TimeSeriesDAI timeSeriesDAO = theDb.makeTimeSeriesDAO();
-					LoadingAppDAI loadingAppDAO = theDb.makeLoadingAppDAO();
+				try(// TODO: This is a good place to hook a transaction. Still need to be implemented though before we can demo it
```
MikeNeilson, line 339: @ktarbet here is some of the initial usage you were interested from the "access-key" PR.

Implementation is in `SimpleOpenDcsDatabaseWrapper`.

Still need to create the transaction, but I think that might come in a separate PR. We've gotten good feedback on the general design and the idea was always to merge into main as soon a practical so all of the finalizing changes are in smaller PRs.

MikeNeilson:
@opendcs/opendcs-core-devs 
@ktarbet @adamkorynta @zack-rma @agilmore2 

This, https://github.com/opendcs/opendcs/pull/681/files/d553babf70169f83c83c04046247774e70370330..67e9bc6eb3b190fd26b03a07bad930a6bc5eacf6, no shows an example implementation and usage of the new OpenDcsDatabase/OpenDcsDao structure with both SQL and XML.

The actual implementation is not complete as the goal was to do just enough for the new EnumDaoTestIT to pass and show an actual usage; not a complete implementation of file locking and such in the XML (though it's needed.)

Here's what I'd like to do
1. Gather commentary on the design so far; I think everyone is on-board and now we have something in the new scheme.
2. Adjust things to commentary
3. Fully implement the EnumDAO in SQL and XML (but for XML still using the IO and engineering list)
4. Fix as many lingering "I need to create a database and run something" issues as possible in short order... knowing we'll find more later.
5. Finally merge in to main
6. Begin work in earnest of adapting downstream systems, especially RoutingScheduler, to the new scheme.

This is already 100 files and difficult to review. Unfortunately I do not think there is a good discrete way to to this that is practical to maintain in parallel. 

The 7.0 branch will continue to avoid these particular Database DAO changes until they're better vetted, but we'll still be able to handle moving bug fixes from main as needed. May just be a bit more work for some of them.

The goal is to move them to 7.0 as the finalization of the connection pool introduction.

NOTE: To make this moving of work from main to the 7.0 branch we will be squash merging PRs into main; this is a permanent change in behavior for the project.
NOTE2: Unless someone has done too much work in a PR and it really is separate discrete work, but we will likely push back on that.

java/opendcs/src/main/java/org/opendcs/database/DatabaseService.java
```diff
@@ -0,0 +1,115 @@
...
+     * @param dataSource
+     * @return
+     * @throws DatabaseException
+     */
+    private static DecodesSettings decodesSettingsFromJdbc(javax.sql.DataSource dataSource) throws DatabaseException
+    {
+        DecodesSettings settings = new DecodesSettings();
+        // TODO: Need a special case for the XML database.
+        try (Connection conn = dataSource.getConnection();
+             PreparedStatement stmt = conn.prepareStatement("select name,value from database_properties");
```
adamkorynta, line 97: I don't see any reference to database scripts creating this table

java/opendcs/src/main/java/org/opendcs/database/DatabaseService.java
```diff
@@ -0,0 +1,115 @@
...
+             ResultSet rs = stmt.executeQuery())
+        {
+            //
+            while (rs.next())
+            {
+                final String name = rs.getString("name");
+                final String value = rs.getString("value");
+                //TODO: map database name/values to appropriate field using reflection.
```
adamkorynta, line 105: TODO

java/opendcs/src/main/java/org/opendcs/database/DatabaseService.java
MikeNeilson: Yep, doesn't exist yet. This was a strawman, at least for the moment.

Can either add a specific table though, or create a "database" app in the existing tsdb_properties table.

MikeNeilson:
@adamkorynta warning, rebasing from main. 

Attempted merge but it lost several important changes in key files.

MikeNeilson:
Alright, I'm quite sure it's going to break some thing, no way to get to literally everything in a single PR, but the bones are there and there's a lot of downstream work waiting on this getting merged in so it can all either start or continue.

I did not bother with updating the newest editions to EnumDAI to the new datatransaction object, I will do so in a separate PR (though I wouldn't be sad if someone else beat me to it.).

The integration tests pass (or at least did locally, waiting on actions) so the basic structure is correctly. A good follow up would likely be to remove Database::makeDatabaseIo and see what gets exposed as needing it, but I'll leave to do another day to decide.

@adamkorynta has a PR to this PR, but instead of merging it in first I'm more inclined to merge this and than that. Planning to do the usual squash commit on those and I feel that the change it's making, a new release artifact, should really be in it's own commit.

adamkorynta (approved):

PR #684, opened 2024-08-29
Bugfix/database access behavior

MikeNeilson:
## Problem Description

A user testing 7.0.13-RC05 was having their routing scheduler fail within a few minutes. While trying to cleanup my test environment I was able to delete more than 4-5 Routing Specs, the primary cause was in the listScheduledEntries code were it did not correctly put the dao in a transaction before a secondary query to the task.

## Solution

- made it possible to enable the trace setting using the run task
- Cleaned up the delete method for rating spec to use transactions and bind variables
- Moved the inTransctionOf line above the usage of a dao in ScheduledEntryDao

## how you tested the change

Local environment for delete, was able to delete an infinite number and verify with the ConnectionPoolMxBean that connection were not left open.

Waiting on feedback from user who's Routing Scheduler failed quickly, likely tomorrow morning.

## Where the following done:

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -822,7 +822,13 @@ more information about each.
             <isset property="opendcs.lang"/>
         </condition>
 
-        <property name="decj_opts" value="${debugOpts} ${jfrOpts} ${lang}"/>
+        <condition property="extra_opts"
+                  value="${opendcs.app.decj}"
+                  else="">
+            <isset property="opendcs.app.decj"/>
+        </condition>
+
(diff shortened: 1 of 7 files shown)


MikeNeilson:
User reported back that their setup has been running more than an hour so far today. They will continue the test for a few more hours and provide some specific logs for more insight.

ktarbet (dismissed):

MikeNeilson:
Unfortunately not a complete fix, but it did help and looks like the test provided enough information for some additional followup.

ktarbet (approved):

PR #695, opened 2024-09-10
Feature/add annotations add to previous

dvdhoz:
test...

### DIFF
--- a/src/main/java/decodes/tsdb/algo/AddToPrevious.java
+++ b/src/main/java/decodes/tsdb/algo/AddToPrevious.java
@@ -1,82 +1,50 @@
 package decodes.tsdb.algo;
 
-import java.util.Date;
-
 import opendcs.dai.TimeSeriesDAI;
 
-import ilex.var.NamedVariableList;
 import ilex.var.NamedVariable;
 import ilex.var.TimedVariable;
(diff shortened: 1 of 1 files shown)


src/main/java/decodes/tsdb/algo/AddToPrevious.java
```diff
@@ -1,82 +1,51 @@
...
-
-//AW:USERINIT
-		// Code here will be run once, after the algorithm object is created.
-//AW:USERINIT_END
 	}
 	
 	/**
 	 * This method is called once before iterating all time slices.
 	 */
 	protected void beforeTimeSlices()
```
MikeNeilson, line 45: If you feel you have the time, beforeTimeSlices, doAWTimeSlice, and afterTimeSlice should also have the `@Override` annotation.

Not huge deal, but it can save some trouble at compile time.

MikeNeilson (commented):
Changes as is look good.

MikeNeilson (approved):

dvdhoz:
@MikeNeilson - I don't have merge privileges. If you don't mind merging the branch for me.

MikeNeilson:
Yep, we basically just merge when we're ready to accept the consequences, not that's really any with this PR.

Issue #683, opened 2024-08-28
DECODES Database Editor Table Sorting


jbatterman:
**Describe the bug**
I have not fully tested all tables, but noticed that the sorting seems to be broken again in this release.  On the Routing Tab, I sorted by "Name" to alphabetize the routing specs and start editing some properties.  The table sorting looked like it worked, but when I try opening a row it is opening the wrong row.

**To Reproduce**
1. Open DECODES Database Editor Routing Tab
2. Click on "Name" to sort by that column
3. Click on "CODWR-WebService-Flow1_CWMS" and choose Open (in NWDM database)
4. The "Metar_Testing" routing spec is opened

**Expected behavior**
The table sorting should follow through so opening rows to edit should match what is shown on the table.

**Screenshots**
Sorted Routing tab by Name:
![image](https://github.com/user-attachments/assets/7f30b827-0d16-4b53-8234-0d26546aeac4)

Opening CODWR-WebService-Flow1_CWMS:
![image](https://github.com/user-attachments/assets/f95ca7ce-dd51-43bb-a43a-f5ef00326097)

**Desktop/ServerSide (please complete the following information):**
 - OS: Windows 10 OpenDCS 7.0.13-RC05
 - Version (server) 7.0.13-RC05

**Additional context**

MikeNeilson:
Thanks. I thought I was going insane a few days ago seeing the same problem.

MikeNeilson:
Additional note, after delete, either the wrong row was used (quite likely) or there's also an issue with the table re-rendering correctly.

MikeNeilson:
Yep, it's the way the TableModels interact. It seems to behave okay if you don't sort again (reverse) on the same column. I'll start updating the models to use the correct methods. Should be able to add some better behavior display wise as well.

PR #696, opened 2024-09-10
Migrate Routing Spec list to standard JTable sorting.

MikeNeilson:
## Problem Description

Partial fix of #683. 

## Solution

Migrated to standard JTable sorting.

## how you tested the change

Manually sorted the table by different means. Verified correct entries opened or deleted.

## Where the following done:


MikeNeilson:
This was smaller than expected, the Sources table is doing the same thing and I will add similar changes to this PR.

MikeNeilson:
Okay, that's enough for this PR. The equipment table is next.

We're also going to need a follow up of PRs to just correct behavior. In reviewing various GUI patterns and just looking at the code. The TableModels should either be where all the database operations go, or they should be a wrapper to a generic form of of such model (preferred any future GUI toolkit changes). Currently the Panels, and even some upstream panels, are performing the same database i/o (deletes and inserts) that should be handled by the model and communicated back when done if another component cares

src/main/java/decodes/dbeditor/NetlistListPanel.java
```diff
@@ -78,51 +54,54 @@ public class NetlistListPanel extends JPanel
...
-	/** Constructor. */
-	public NetlistListPanel()
-	{
+    /** Constructor. */
+    public NetlistListPanel()
+    {
         try
-		{
-			tableModel = new NetlistListTableModel();
-			netlistListTable = new SortingListTable(tableModel,
```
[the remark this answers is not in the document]
MikeNeilson: Good idea.

src/main/java/decodes/dbeditor/presentation/PresentationGroupTableModel.java
```diff
@@ -0,0 +1,144 @@
...
+	public void add(PresentationGroup ob)
+	{
+		pgList.add(ob);
+		fireTableDataChanged();
+	}
+
+	public void replace(PresentationGroup oldOb, PresentationGroup newOb)
+	{
+		pgList.remove(oldOb);
+		pgList.add(newOb);
```
ktarbet, line 72: should the table refresh here?

ktarbet (approved):

src/main/java/decodes/dbeditor/presentation/PresentationGroupTableModel.java
MikeNeilson: Yep, good catch.

ktarbet (approved):

PR #699, opened 2024-09-12
Feature/add more annotations

dvdhoz:
Add annotations for algorithms

### DIFF
--- a/src/main/java/decodes/tsdb/algo/AverageAlgorithm.java
+++ b/src/main/java/decodes/tsdb/algo/AverageAlgorithm.java
@@ -2,7 +2,7 @@ package decodes.tsdb.algo;
 
 import ilex.var.NamedVariable;
 import decodes.tsdb.DbCompException;
-import decodes.util.PropertySpec;
+import org.opendcs.annotations.PropertySpec;
 import org.opendcs.annotations.algorithm.Algorithm;
 import org.opendcs.annotations.algorithm.Input;
 import org.opendcs.annotations.algorithm.Output;
@@ -21,24 +21,14 @@ public class AverageAlgorithm extends decodes.tsdb.algo.AW_AlgorithmBase
 
 	double tally;
 	int count;
-	PropertySpec specs[] =
-	{
-		new PropertySpec("negativeReplacement", PropertySpec.NUMBER, 
-			"(no default) If set, and output would be negative, then replace with the number supplied.")
-	};
-	@Override
-	protected PropertySpec[] getAlgoPropertySpecs()
-	{
-		return specs;
-	}
-
-	public double negativeReplacement = Double.NEGATIVE_INFINITY;
 
 	@Output
 	NamedVariable average = new NamedVariable(AVERAGESTRING, 0);
 
 	@org.opendcs.annotations.PropertySpec(value = "1")
 	public long minSamplesNeeded = 1;
+	@PropertySpec(value = "-1.0 / 0.0")
+	public double negativeReplacement = Double.NEGATIVE_INFINITY;
 
 	// Allow javac to generate a no-args constructor.
 
--- a/src/main/java/decodes/tsdb/algo/AverageAlgorithm.java
+++ b/src/main/java/decodes/tsdb/algo/AverageAlgorithm.java
@@ -27,7 +27,7 @@ public class AverageAlgorithm extends decodes.tsdb.algo.AW_AlgorithmBase
 
 	@org.opendcs.annotations.PropertySpec(value = "1")
 	public long minSamplesNeeded = 1;
-	@PropertySpec(value = "-1.0 / 0.0")
+	@PropertySpec(value = "Double.NEGATIVE_INFINITY")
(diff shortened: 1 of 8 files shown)


src/main/java/decodes/tsdb/algo/AverageAlgorithm.java
```diff
@@ -1,84 +1,71 @@
...
-//AW:INPUTS_END
+	private static final String AVERAGESTRING = "average";
+	@Input
+	double input;
 
-//AW:LOCALVARS
 	double tally;
 	int count;
-	PropertySpec specs[] = 
+	PropertySpec specs[] =
```
MikeNeilson, line 24: you can delete this list and the associated method. The base class now takes care of it given the annotation below.

src/main/java/decodes/tsdb/algo/AverageAlgorithm.java
```diff
@@ -1,84 +1,71 @@
...
 			"(no default) If set, and output would be negative, then replace with the number supplied.")
 	};
+	@Override
+	protected PropertySpec[] getAlgoPropertySpecs()
+	{
+		return specs;
+	}
 
-//AW:LOCALVARS_END
+	public double negativeReplacement = Double.NEGATIVE_INFINITY;
```
MikeNeilson, line 35: Put the PropertySpec annotation on this one.

src/main/java/decodes/tsdb/algo/CopyAlgorithm.java
```diff
@@ -12,46 +12,36 @@
...
-//AW:LOCALVARS
 	private UnitConverter converter = null;
 	private boolean enhancedDebug = false;
 	private String inUnits = null;
 	private String outUnits = null;
-	private String debugTsid = "DET.Elev-Forebay.Inst.0.0.MIXED-RAW";
-	
-	private PropertySpec copyPropertySpecs[] = 
+
+	private final PropertySpec copyPropertySpecs[] =
```
MikeNeilson, line 44: This and the associated function can be deleted.

MikeNeilson (changes requested):
Looks good. Few things that can also be deleted though.

src/main/java/decodes/tsdb/algo/AverageAlgorithm.java
dvdhoz: @MikeNeilson - how should I handle the String value? Set it as "Double.NEGATIVE_INFINITY" or "-1 / 0" ?

MikeNeilson: Without any value would work, I think, but probably better to just copy the actual value in. You can get the actual value from the java source or something.

Going to have to think about that though, long term should probably be the text "missing" in case some implementation doesn't want to use NEGATIVE_INFINITY to show that, but we don't need to decided that now.

src/main/java/decodes/tsdb/algo/AverageAlgorithm.java
```diff
@@ -1,84 +1,61 @@
...
 
-//AW:OUTPUTS
-	NamedVariable average = new NamedVariable("average", 0);
-	String _outputNames[] = { "average" };
-//AW:OUTPUTS_END
-
-//AW:PROPERTIES
+	@org.opendcs.annotations.PropertySpec(value = "1")
 	public long minSamplesNeeded = 1;
+	@PropertySpec(value = "-1.0 / 0.0")
```
MikeNeilson, line 30: I don't believe our system will do this math and thus this may break usages.

Don't worry about it for now though, I'll look into that a bit later.

src/main/java/decodes/tsdb/algo/CopyAlgorithm.java
```diff
@@ -12,110 +12,79 @@
...
-//AW:OUTPUTS
-	// A single output named 'output':
-	public NamedVariable output = new NamedVariable("output", 0);
-	String _outputNames[] = { "output" };
-//AW:OUTPUTS_END
+	@Output
+	public NamedVariable output = new NamedVariable(OUTPUTSTRING, 0);
 
-//AW:PROPERTIES
+	@org.opendcs.annotations.PropertySpec(value = "1")
```
MikeNeilson, line 47: Should use 1.0 and 0.0 to better indicate it's not an integer that's required.

MikeNeilson (changes requested):
One change to make now.

Please wait on the negativeReplacement until I've had time to investigate behavior.

Otherwise looking good.

NOTE: please to do force push unless asked. We can, and likely will squash on merge so there is no need to clean up commits during PR review; makes it difficult to review as additional changes are made.

src/main/java/decodes/tsdb/algo/AverageAlgorithm.java
MikeNeilson: I just looked, the properties system will handled the text `Double.NEGATIVE_INIFINITY` correct, use that exact text as the default value.

MikeNeilson (changes requested):

PR #704, opened 2024-09-18
Change windows min memory setting to actually be min memory.

MikeNeilson:
## Problem Description

Discovered that we were setting the max memory instead of the min memory in the windows batch file.

## Solution

change parameter to the correct -Xms

## how you tested the change

Currently on a linux machine (was already correct in batch script) and cannot test directly.

## Where the following done:


adamkorynta:
oops

adamkorynta (approved):

adamkorynta:
should the shell script equivalent get updated to 240m?

MikeNeilson:
> should the shell script equivalent get updated to 240m?

I thought about it, but it's been that way for so long I decided to just leave it.
I think that 240m may have intentionally been a max given it was always that way. 

arguably i probably makes more sense to remove both.

PR #711, opened 2024-09-26
Update generic_es_US.properties  (consistance key  case with English)

ktarbet:
github editor made this change look bigger than simple edit on line 70.


MikeNeilson:
Weird. wonder if the browser tried to change the encoding or something. I'm going to download the branch first to see what's going on.

MikeNeilson (approved):
Looks like the original values were not encoded correctly. Github did us a favor and just updated everything to correct UTF8.

PR #728, opened 2024-10-10
Preload TsId cache before loading groups. Move group cache fill to tr…

MikeNeilson:
…ansaction and avoid exception handling in tight loop.

## Problem Description

User reported regression in loading CompEdit between .12 and .13. Discovered Group processing was calling into the database in a tight loop for every TimeSeriesID the group would need.

## Solution

Have compedit load the TsId Cache before getting the group list.

## how you tested the change

Manually, loading of a database with large groups went from minutes to near instantaneous.

## Where the following done:


DanielTOsborne (approved):
I ran this and confirmed these changes fix the slow startup.

PR #746, opened 2024-10-18
Bugfix/745 routing spec updates

### LINKED ISSUE #745: Issue saving/updating routing spec
**Describe the bug**
DbEdit hangs, and then reports failure, while updating a routing spec.

**To Reproduce**
1. Open Decodes Database Editor
2. Go to routings tab
3. copy a routing
4. attempt to update
5. see errors about parent key missing

**Expected behavior**
Updates save correctly.

**Screenshots**

 - OS: Ubuntu
 - Version 7.0.13, latest commit

**Additional context**
After a minor tweak to the code the hang happens.

MikeNeilson:
## Problem Description

Fixes #745. 

## Solution

1. Moved all operations within the appropriate try-catch block
2. Modified function calls in RoutingSpecListIO so ensure the same connection/transaction was used for all operations
3. Added an addOrReplace to TableModel to prevent needless deletion from the database.
4. Migrated propertiesDAO to use bind variables
5. Migrated more of RoutingSpecListIO operations to use bind variables (new handlers to properly wrap connections are at least used.)
6. Changed GUI behavior on commit. A TraceDialog will open, if successful will close automatically, if not it will remain open with the contents of the error.
7. Changed GUI behavior on copy. No longer immediately saves the copy.
8. Changed GUI behavior on save. Database operations now off GUI thread.

## how you tested the change

Manually
1. copied routing spec, saved, closed/repoen dbedit, verified it remained.
2. opened routing spec, modified, saved, closed/reopen dbedit, verified it remained and had changes.
3. opened routing spec, modified with bad data (properties too long), saved, saw errors, closed/repopen dbeidt, verified it remained with original data.

## Where the following done:

### DIFF
--- a/src/main/java/decodes/dbeditor/routing/RSListTableModel.java
+++ b/src/main/java/decodes/dbeditor/routing/RSListTableModel.java
@@ -117,6 +117,9 @@ public class RSListTableModel extends AbstractTableModel
         RoutingSpec found = theList.find(ob.getName());
         if (found != null)
         {
+            // we very specifically do not call DbIo.deleteRoutingSpec here as we just saved
+            // the updates and we don't need to bother writing them to the database again.
+            // we just need the RoutingSpecList to have the new spec.
             theList.remove(found);
         }
         theList.add(ob);
--- a/src/main/java/decodes/dbeditor/routing/RSListTableModel.java
+++ b/src/main/java/decodes/dbeditor/routing/RSListTableModel.java
@@ -120,6 +120,7 @@ public class RSListTableModel extends AbstractTableModel
             theList.remove(found);
         }
         theList.add(ob);
+        fireTableDataChanged();
     }
 
     public void deleteObject(RoutingSpec ob)
--- a/src/main/java/decodes/dbeditor/routing/RSListTableModel.java
+++ b/src/main/java/decodes/dbeditor/routing/RSListTableModel.java
@@ -112,6 +112,16 @@ public class RSListTableModel extends AbstractTableModel
         fireTableDataChanged();
     }
 
+    public void addOrReplace(RoutingSpec ob)
+    {
+        RoutingSpec found = theList.find(ob.getName());
+        if (found != null)
+        {
+            theList.remove(found);
+        }
+        theList.add(ob);
+    }
+
     public void deleteObject(RoutingSpec ob)
     {
         try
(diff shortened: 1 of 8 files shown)


src/main/java/decodes/dbeditor/routing/RSListTableModel.java
```diff
@@ -112,6 +112,16 @@ public void addObject(RoutingSpec ob)
         fireTableDataChanged();
     }
 
+    public void addOrReplace(RoutingSpec ob)
+    {
+        RoutingSpec found = theList.find(ob.getName());
+        if (found != null)
+        {
+            theList.remove(found);
```
ktarbet, line 120: does the database version also need to be deleted here? (call deleteObject(RoutingSpec ob)

src/main/java/opendcs/dao/PropertiesSqlDao.java
```diff
@@ -182,77 +194,83 @@ public void deleteProperties(String tableName, String idColumn,
...
                         if (value == null)
+                        {
                             value = "";
+                        }
 
                         chp.setProperty(name, value);
-                        n++;
-                        continue nextProp;
+                        n.getAndIncrement();
+                        return;
```
ktarbet, line 264: This return 'feels' like we will only get one item?

ktarbet (commented):
the build was failing, I've added some comments.

src/main/java/decodes/dbeditor/routing/RSListTableModel.java
MikeNeilson: This is specifically *not* calling the database delete, as "add" doesn't call the database write, and thus the Spec will magically disappear from the database.

There's a lot of things to fix, or just make a firm decision about, in the table models.

src/main/java/opendcs/dao/PropertiesSqlDao.java
MikeNeilson: no, that return move to the next ResultSet so it could increment from that. Honestly that whole block confuses me in design, had a jump to label originally.

DanielTOsborne (approved):
Looks good to me. I tested this branch locally and only ran into an issue with the updated item changing order in the list, but commit cc5d6be4e95a511708be26dcfca2fc2dbd25ed8c fixes that.

PR #768, opened 2024-10-29
Devops/refactor integration fixtures

adamkorynta:
## Problem Description

Need to be able to publish artifacts to mavenLocal() for downstream usage. 
Need to be able to use the database Configurations for integration testing in the REST API.

## Solution

Add maven-publish to gradle files.
Refactor test fixtures from the tests subproject into the fixtures subproject.

## how you tested the change

Testing through the REST API.

## Where the following done:

   - [X] Integration tests created or modified that run during integration testing
- [X] Were relevant config element (e.g. XML data) updated as appropriate


MikeNeilson:
The changes overall look good. I'm currently working on adding the publications and signing in main at the moment. I'd prefer to get that done there and merge into this, but the changes to the fixtures themselves look reasonable so and the publications will be roughly the same so you can continue from your branch for any testing without major distruption.

integrationtesting/fixtures/src/main/java/org/opendcs/fixtures/configuration/Configuration.java
```diff
@@ -1,4 +1,4 @@
-package org.opendcs.spi.configuration;
+package org.opendcs.fixtures.configuration;
```
MikeNeilson, line 1: I don't see an issue with moving this interface, but I've been keeping "spi" in the package name intentionally as a hint to the reader.

Does make sense to keep under fixtures but perhaps "...fixtures.spi..."?

ktarbet (commented):
Looks OK to me

java/opendcs/build.gradle
```diff
@@ -2,7 +2,7 @@ plugins
 {
     id 'opendcs.deps-conventions'
     id 'opendcs.java-conventions'
-    id 'java-library'
```
MikeNeilson, line 5: Are you sure this doesn't need to key java-library?

java/opendcs/src/main/java/opendcs/dai/DaiBase.java
```diff
@@ -11,8 +11,9 @@
 import opendcs.dao.DaoBase;
 import opendcs.dao.DaoHelper;
 import opendcs.util.functional.DaoConsumer;
+import org.opendcs.database.api.OpenDcsDao;
 
-public interface DaiBase extends AutoCloseable
+public interface DaiBase extends AutoCloseable, OpenDcsDao
```
MikeNeilson, line 16: FYI, I've been intentionally not doing this; however, I see no need to change here for you're current testing.

I think the main goal should be to not have the DAI/DAO chain of inheritance but it's going to take time to break.

Good news is, on the rest_api end you shouldn't need to do anything. Just not going to merge in in to the target so it's clear which DAOs have actually been migrated.

java/opendcs/src/main/java/opendcs/dai/EnumDAI.java
```diff
@@ -21,7 +21,7 @@
  * Data Access Interface for database-resident enumerations
  * @author mmaloney Mike Maloney
  */
-public interface EnumDAI extends DaiBase, OpenDcsDao
+public interface EnumDAI extends DaiBase
```
MikeNeilson, line 24: Same comment as DaiBase; was intentionally done but I see no reason to change it in the current refactor branch as the long term affect is intended to be the same.

MikeNeilson:
@adamkorynta the target branch is actually updated with the CWMS Flyway work to install the CCP schema.

PR #769, opened 2024-10-29
Devops/setup artifact signing

MikeNeilson:
## Problem Description

Setup artifact signing, using sigstore, and tag based versioning. Additionally setup initial publications

## Solution

Add sigstore plugin to build. This using OpenID connect workflows to verify identity and avoids needing to create gpg keys.
The plugin can use the project identity to sign on behalf the project, or when testing manually you use your own github account.

## how you tested the change

Manually running `publishToMavenLocal` task to trigger signing.
for tag versioning ran `./gradlew -q properties` to see appropriate version is set.

## Where the following done:

### DIFF
--- a/buildSrc/src/main/groovy/opendcs.publishing-conventions.gradle
+++ b/buildSrc/src/main/groovy/opendcs.publishing-conventions.gradle
@@ -21,11 +21,5 @@ if (JavaVersion.current() != JavaVersion.VERSION_1_8) {
                 issuer.set("https://oauth2.sigstore.dev/auth")
             }
         }
-}
-}
-/*
-tasks.withType(SigstoreSignFilesTask).configureEach {
-    javaLauncher = javaToolchains.launcherFor {
-        languageVersion = JavaLanguageVersion.of(11)
     }
-}*/
+}
--- a/buildSrc/src/main/groovy/opendcs.publishing-conventions.gradle
+++ b/buildSrc/src/main/groovy/opendcs.publishing-conventions.gradle
@@ -1,0 +1,11 @@
+plugins {
+    id 'signing'
+    id 'maven-publish'
+}
+
+
+
+
+signing {
+    useGpgCmd()    
+}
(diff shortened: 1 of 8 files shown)


MikeNeilson:
There are additional changes to support the distribution zip and tar as well as the containers to work into the build.yaml.

I'm planning to do those in a followup PR but Adam is working on something that could use this particular bit sooner.

buildSrc/src/main/groovy/opendcs.publishing-conventions.gradle
```diff
@@ -0,0 +1,31 @@
...
+            gitHub {
+                audience.set("sigstore")
+            }
+
+            web {
+                clientId.set("sigstore")
+                issuer.set("https://oauth2.sigstore.dev/auth")
+            }
+        }
+}
```
ktarbet, line 24:
```suggestion
              }
```

buildSrc/src/main/groovy/opendcs.publishing-conventions.gradle
```diff
@@ -0,0 +1,31 @@
...
+}
+/*
+tasks.withType(SigstoreSignFilesTask).configureEach {
```
ktarbet, line 27: should this commented part be a else?

ktarbet (approved):
The overall direction looks good to me.  I expect more adjustments will be needed, especially for Java 11.

buildSrc/src/main/groovy/opendcs.publishing-conventions.gradle
MikeNeilson: No, I should delete it, it was part of the orginal attempt to allow mixed java versioning in the build.

MikeNeilson:
> The overall direction looks good to me. I expect more adjustments will be needed, especially for Java 11.

Yeah, I imagine we'll end up with a few of these java version check blocks. SonarQube requires 17, but the tool is too useful not to deal with it.... and apparently our project is too big for automated analysis.

ktarbet (approved):

PR #778, opened 2024-10-30
Chore/7.0 updates from main

MikeNeilson:
Bring in latest build fixes, except sonarqube, into 7.0

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,15 @@
 name: OpenDCS Build and unit tests
 on:
   pull_request:
-    branches: ["master"]
+    branches:
+      - "main"
+      - "7.0"
+      - "8.0"
   push:
-    branches: ["master"]
+    branches:
+      - "main"
+      - "7.0"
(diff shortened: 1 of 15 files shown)


.github/workflows/build.yml
```diff
@@ -1,9 +1,15 @@
 name: OpenDCS Build and unit tests
 on:
   pull_request:
-    branches: ["master"]
+    branches:
+      - "main"
+      - "7.0"
+      - "8.0"
```
ktarbet, line 7: If this is the 7.0 branch, should we remove main and 8.0 PR's?   same question for push.

ktarbet (approved):
see my question about building 8.0 and main branch from the 7.0 branch

.github/workflows/build.yml
MikeNeilson: Honestly, I'm not entirely sure. We'll need to do a little research, but for know I at least know it's not breaking anything to do it this way.

ktarbet (approved):

PR #789, opened 2024-11-08
Feature/dds over http

### LINKED ISSUE #220: LRGS web interface
The LRGS main interface is already an HTML page. Though well embedded into Java. A pure HTTP(s) interface would simplify a lot of operations and setup going forward.

JAX-RS has been used rest-api API work.
Will likely create a client side single page application. USACE staff are learning React, we should review other frameworks among @opendcs/opendcs-core-devs to decide what will work best for us.

Javalin was previously used; however the https://github.com/opendcs_rest API settled on JAX-RS and after review that is arguably easier to work with.

**Additional context**
I will take an initial stab at a DDS-over-HTTP protocol. Documentation will be available at the https://github.com/opendcs/dcs_standards repo and an OpenAPI definition an LRGS instance can publish.

Add any additional requests/requirements here. No promises on more than making sure they can happen eventually.

MikeNeilson:
## Problem Description

Partial #220 . 

The DDS Protocol is often difficult to deal with, especially be custom and on a plain TCP Port.

This implements https://github.com/opendcs/dcs_standards/pull/3, which is incomplete, and begins adding appropriate HTTP/HTML/etc interfaces to an LRGS

## Solution

Using the existing embedded jetty used for the Health end point, added additional endpoints for retrieval of data from the archive

Implements basic dds endpoints for data retrieval. "dataSummary" was initially left off this initial work.

## how you tested the change

Additional integration tests.

## Where the following done:


java/opendcs/src/main/java/org/opendcs/lrgs/http/dto/DcpMsg.java
ktarbet, line 1: lrgs.common.DcpMsg is also competing for this name.

ktarbet (commented):
great start! 
Will this be able to run in parallel with existing LRGS data access?

java/opendcs/src/main/java/org/opendcs/lrgs/http/dto/DataSource.java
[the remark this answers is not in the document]
MikeNeilson: Those are intentional. They shadow those classes for the specific purpose of facilitating all of the JSON/XML configuration markers.

Though that should be LrgsDataSource, not the Generic DataSource.

MikeNeilson:
> great start! Will this be able to run in parallel with existing LRGS data access?

Yeah. I'm going to follow @DanielTOsborne 's lead and add STARTTLS support on the default port and will likely just use 16004, vs 16003, for the TLS only port.

install/src/main/dist/scripts/lrgs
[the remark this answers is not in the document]
MikeNeilson: this is a new script used for testing and a windows version hasn't been made yet. But that should be done before final merge.

integrationtesting/lrgs/src/test/java/org/opendcs/lrgs/dds/DdsHttpTest.java
```diff
@@ -0,0 +1,156 @@
...
+    """)
+final class DdsHttpTest
+{
+    static int port;
+    static MsgArchive archive;
+
+    @BeforeAll
+    static void setup(LrgsTestInstance lrgs)
+    {
+        var lrgsInput = (LrgsHttpInterface)lrgs.getLrgsInputs()
```
ktarbet: Is it necessary to have both LrgsInputInterfaceand LrgsHttpInterface., as opposed to just LrgsInterface?   

Perhaps some comments  on the purpose of LrgsInputInterface would help; it feels like more that just inputs/config.

ktarbet (commented):
checkpoint

integrationtesting/lrgs/src/test/java/org/opendcs/lrgs/dds/DdsHttpTest.java
MikeNeilson: LrgsInputIterface is the existing interface for LRGS inputs.

LrgsHttpInterfacce isn't an interface and should probably just be renamed to LrgsHttpInput.

MikeNeilson: I rename the implementation. The original name was left in as a derived class with a comment (we're actively using that name in the public lrgs) explaining why.

MikeNeilson:
sonar will report duplication, that is correct, will work on cleanups in follow ups; should be able to share some logic wit the new netty dds system.

ktarbet (approved):

PR #795, opened 2024-11-19
Bring the log rotation back to the file logger.

### LINKED ISSUE #794: OpenDCS 7.0.13-RC08 Server Side Log Size
We are running OpenDCS 7.0.13-RC08 on our T7 server.  Our comp,logs and routsched.logs are no longer rolling over at a certain size;  one of our comp.logs was 43 Gb before we manually rolled it.

MikeNeilson:
## Problem Description

Fixes #794. 

## Solution

Track bytes (characters?) written and rotate log when maximum is reached.

## how you tested the change

Created unit test to verify behavior.

## Where the following done:


java/opendcs/src/main/java/ilex/util/FileLogger.java
```diff
@@ -87,25 +88,33 @@ public FileLogger( String procName, String filename )
...
 						PrintStream ps = output.get();
 						if (msg != null && ps != null && !ps.checkError())
 						{
+							if (!isTerminal) // avoid any kind of buffer overflow errors since this would just keep accumulating.
+							{
+								bytesWritten.getAndAdd(msg.length());
+							}
 							ps.println(msg);
 						}
+						if (!isTerminal && (bytesWritten.get() >= this.maxLength))
```
agilmore2, line 114: Any chance we can get configurable log file lengths for OpenDCS programs, or is this whole file logging function being re-written elsewhere?

MikeNeilson:
We could. But the end goal is to elimanate this class in favor of using
standard loggers, such as java.util.logging where there already defined
ways to do so.

But if we reslly want it now it's more a matter of how easy it should be
for the user vs say a developer.

MikeNeilson:
But I would say that's a feature request. This is a break-fix, had one of
the district end up with a 45 gb log file.

java/opendcs/src/main/java/ilex/util/FileLogger.java
ktarbet: @agilmore2  - are you thinking about a feature where decodes.properties, or command line argument allows setting the max log file size?

ktarbet (commented):
There are some comments in FileLogger about ant that should probably be changed to gradle?  


If the FileLogger is started several times with the same filename you can get a file of any size because during the close method, or the constructor bytesWritten is set to zero.

Perhaps the use case always has a new file?

```java

@Test
    void make_large_log_file() throws Exception
    {
        Path filePath = Files.createFile(tempDir.resolve("test_large.log"));
        if(!Files.exists(filePath)){
            Files.createFile(filePath);
        }
        System.out.println(filePath.toString());

        for (int j = 0; j < 1000; j++) {
            FileLogger log = new FileLogger("test", filePath.toString(), 700);

            for (int i = 0; i < 6; i++) {
                log.doLog(1, ""+i);
            }
            Thread.sleep(50);
            log.close();
            long fileSizeInBytes = Files.size(filePath);
            System.out.println(j+" File size in bytes: " + fileSizeInBytes);
        }
    }
```
``` text
997 File size in bytes: 173652
998 File size in bytes: 173826
999 File size in bytes: 174000

MikeNeilson:
Yeah, there's not much safety in the file operations. a `FileChanel` was used to lock the file previously but it's rare than any user code would use this class so I think we're fairly safe in that regard.

Though one issue I can see is if a user accidentally starts the program again before it's finished. But in that case logging has always been rather iffy anyways.

ktarbet (approved):

PR #744, opened 2024-10-18
Feature/423 cwms flyway

MikeNeilson:
## Problem Description

Fixes #432 

Describe the problem you are trying to solve.

## Solution

Move CWMS CCP Schema elements to flyway style. Created CWMS-Oracle migration provider.
Modified integration-test code to use this.

TODO:
- [x] Allow appropriate bypass options for both CWMS Database and OpenDCS/CCP schema

## how you tested the change

Tested with the automated tests.

## Where the following done:

- [ ] Tests. Check all that apply:
   - [ ] Unit tests created or modified that run during ant test.
   - [ ] Integration tests created or modified that run during integration testing
         (Formerly called regression tests.)
   - [ ] Test procedure descriptions for manual testing
- [ ] Was relevant documentation updated?
- [ ] Were relevant config element (e.g. XML data) updated as appropriate

If you aren't sure leave unchecked and we will help guide you to want needs changing where.

PR #797, opened 2024-11-25
Bugfix/779 makeconcrete tsdao usage

### LINKED ISSUE #779: UCP-29: Failed to get a connection
**Describe the bug**
Tulsa district (SWT) was wanting to test modern versions of OpenDCS against our data/database. 

With 7.0.13 installed on our COOP we updated our user.properties and .decodes.auth to target our primary database. We ran our LRGS / CCP / RS / etc on the coop. 

Everything appeared to work with the exception of roughly 6 of our 9 algorithms 

**To Reproduce**
1. Stop OpenDCS suite on primary
2. Install 7.0.13 to COOP T7
3. Update link on COOP to point at new version
4. Update config files to point to primary (so that they can write)
5. Use `BootDCS` to start all opendcs applications including LRGS
6. Use `complocklist` to see if we are getting errors / review logs

**Expected behavior**
Comp processes to finish and write to database

**Screenshots**
image
 - OS: Solaris/T7
 - Version : 7.0.13

**Additional context**
Thought this might be related to #480 but the errors are different. 

Here is a github jist to the full log file
[Github Gist of log](https://gist.github.com/krowvin/9775423df4707987cde96ad21c68d624#file-swt-opendcs-7-0-13-ucp-error-log-L260)

MikeNeilson:
## Problem Description

Fixes #779 . 

## Solution

1. Change usages of DbCompResolver.makeConcrete to take a TimeSeriesDAI instance so that it can be set to a transaction to avoid opening to many connections in a tight loop.
2. Change from Oracle UCP in CWMS to HikariCP.

The change to makeConcrete usage *did* make the situtation better, but we still ran out of connections. However, due to the extra layers of abstraction from CwmsDbConnectionPool to Oracle UCP it is difficult to determine if the opendcs CwmsConnectionPool wrapper is the main problem or not. Since we're already planning to move to standard interfaces I've decided to just try HikariCP here. 

Initial results are good, and only affects CWMS.

I'm not sure this is the quite way to include a new connection pool concept in the the software, but this problem is frustrating enough I'm at least testing the "easy" way to verify were problems are.

## how you tested the change

Manually running compproc and seeing that it did fail in less than 10 minutes... and than 20 minutes. 
About to start a long running test on our infrastructure.

## Where the following done:


install/build.gradle
```diff
@@ -124,15 +124,15 @@ task runApp(type: Exec) {
     }
 
     if (appArgs != null) {
-        args += appArgs
+        args += appArgs//.split("\\s+")
```
ktarbet, line 127: I'm confused by the comment  //.split(...)

MikeNeilson: oh, when I originally made that change I thought I needed to split the provided string into an array, it turns out not. Forgot to delete the code completely.

ktarbet (commented):
Is the new dependency hikaricp worth maintenance effort of adding  another dependency?

MikeNeilson:
@ktarbet, @agilmore2, and @opendcs/opendcs-core-devs  So I am inclined towards using HikariCP for the main application distribution just to maintain consistency; however, both the oracle jdbc and postgres jdbc drivers include an appropriate Pooled Data Source in their respective packages. 

So I'd like to put it to others.

For this PR I'm confident the change to actually using a pool that we know closes the connections was the main issue so for CWMS just bypassing the hec-cwms-data-access provided extra layer on said pool is enough. If others think we should avoid additional dependencies I can change this to use the Oracle UCP directly and get the same affect.

MikeNeilson:
Okay, sadly it appears #744's minor change did not fix the connection pool issue corrected here.
However, in the interest of not going too overboard, I'm going to remove the hikariCP dependency from this PR in favor of the Oracle UCP directly. I'm in favor of HikariCP long term, but we can deal with that during the larger database initialization changes.

MikeNeilson:
Ran test in local VM against a computation process that was known to trigger the error. Not all data has processed yet but that appears to be upstream database issues (save call taking 2 seconds each) and not the changes themselves.

ktarbet (approved):

PR #798, opened 2024-11-26
Initial design of an opendcs 'data-access-key' for stateless DAOs.

MikeNeilson:
Strawman design for discussion of future "stateless" DAO design.

Frustration earlier in the week led me to think more about the future and how a stateless DAO would work, as well as accounting for the possiblity of "non-SQL" or mixed SQL/non-SQL within a database implementation.

I've come up with the idea of the "DataTransaction." An actual SQL transaction and sharing the name open connection between calls in a tight loop is how I've been fixing most recent performance or data consistency issues. So we clearly need to do that anyways, this would allow for cleaner code vs the existing "somehow I've hacked transactions into the existing DAOs" but also allow for proper transactions (e.g. file locks) for the the XML database.

This also doesn't force implementations into a particular implementation (SQL/HTTP/File/etc) while, I hope, making it clear what the intent is while reading and writing sections of code.


java/opendcs-api/src/main/java/org/opendcs/database/api/DataTransaction.java
```diff
@@ -0,0 +1,51 @@
...
+     * @param connectionType
+     * @return optional with the instance of the connection type, if available.
+     * @throws OpenDCSDatabaseException any issues with the connection, if implementations check validity.
+     */
+    <T> Optional<T> connection(Class<T> connectionType) throws OpenDcsDataException;
+
+    /**
+     * Finalize transaction state across all connections.
+     * 
+     * On successful commit implementations *MUST* provide a new valid transaction.
```
adamkorynta, line 36: nitpick, but because I read this incorrectly at first: missing comma after `On successful commit`

java/opendcs-api/src/main/java/org/opendcs/database/api/DataTransaction.java
```diff
@@ -0,0 +1,51 @@
...
+
+import javax.management.openmbean.OpenDataException;
+
+/**
+ * Instance of this class are used to hold appropriate connections
+ * to external systems. Most commonly will be SQL connections for the OpenDCS information.
+ * Implementations are free to provide any type required; however, generic implementations 
+ * will assume SQL based operations.
+ * 
+ * Implementations if this interface *MUST* hold and provide connections to datasources
```
ktarbet, line 13:
```suggestion
 * Implementations of this interface *MUST* hold and provide connections to datasources 
```

java/opendcs-api/src/main/java/org/opendcs/database/api/DataTransaction.java
```diff
@@ -0,0 +1,51 @@
...
+     * @throws OpenDcsDatabaseException any issues with finalizing the transaction.
+     */
+    void commit() throws OpenDcsDataException;
+
+    /**
+     * Default autoclose behavior is to call commit.
+     * @throws Exception
+     */
+    @Override
+    default void close() throws Exception
```
ktarbet, line 47: should there also be a rollback() method?

MikeNeilson: ... yes, I knew I was definitely missing something in that.

MikeNeilson: We did discuss the HTTP transaction thing on a CDA call once in relation to the Lock (of data) design, and the possibility for CDA. This design assumes someone may actually want to implement such a design in the future.

MikeNeilson: More I'm saying the *could* occur. Considering we don't have that now and it's "mostly" okay, it's more a "nice to have."

I'm certainly not going to stop anyone from doing so. And will likely at least add some locking in since we are trying to add more parallel operations... which means the "get a transaction" interface should probably have a timeout option instead of throwing an error because of a wait.

adamkorynta: I guess I'm concerned that the initial XML implementation would no-op the rollback. Definitely (?) wouldn't want to throw an exception, but also the contract of this interface would tell the caller that a successful rollback occurred.

MikeNeilson: True. But the challenge is how do we deal with that?

While writing the code for any sql operations I need to use the same connection, in a transaction, for all DAO operations. The current method is awkward.

At the same time, want to allow for the future where these things exist.

And the core code, needs to just not care where anything came from and leave the edge cases to the implementations.

That said, I don't think a rollback for the XML database would be too difficult to manage, so perhaps we put it on the required before out of RC phase.


The HTTP element is a whole other fiasco in that regard. It doesn't make sense to get a connection as a transaction for one thing and not another, the code, like the GUI that's using all of this should expect a single interface.

java/opendcs-api/src/main/java/org/opendcs/database/api/DataTransaction.java
```diff
@@ -0,0 +1,60 @@
...
+     * 
+     * @throws OpenDcsDatabaseException any issues with finalizing the transaction.
+     */
+    void commit() throws OpenDcsDataException;
+
+    /**
+     * Reset data sources to know state.
+     * Transaction *MUST* should be in the valid initial state after return.
+     */
+    void rollback() throws OpenDcsDataException;
```
ktarbet, line 49:
```suggestion
    /**
     * Reset data sources to known state.
     * Transaction *MUST* return to the valid initial state.
     */
    void rollback() throws OpenDcsDataException;
```

java/opendcs-api/src/main/java/org/opendcs/database/api/DataTransaction.java
adamkorynta: It could be delegated to an optional object `Optional<RollbackTransaction> getRollbackTransaction()`. This would open up the option for SavePoint capabilities as well.

MikeNeilson: I'm not sure that really solves the issue though.

Sure we don't really need a proper transaction for reads, just reusing the same connection helps a lot there, but for any writes, data could be across sources in our future.

It seems like all we can do is assert it *should* work a certain way, and leave it to downstream to decide how critical it is.

I mean, I'm in favor of saying "if you don't actually support a transaction in this way you haven't implemented things properly"... but them I'm also giving the 3 of us a rather difficult task going forward. Perhaps that's the right thing to do.


But one of the major issues with OpenDCS is core operations needing to check what the implementation is and adding an edge case. I'm not sure what you're suggesting avoids that.

adamkorynta: It changes it from asking what the implementation is to asking what the implementation supports.
```
DataTransaction dt = db.newTransaction();

....do all the things but we want to abort....

dt.getRollbackTransaction()
    .ifPresentOrElse(RollbackTransaction::rollback, () -> showDialog("Rollback is not supported, data left in an incomplete state"));
dt.close();
```

This could definitely be overdesign given the primary use-case is JDBC. I initially thought rollback() was intentionally left out due to rollbacks only being needed in the implementations which would already know that they are JDBC.

MikeNeilson: I was originally thinking of cross system transaction, hence naming the class `DataTransaction`

But I see, you aren't thinking a separate call within OpenDcsDatabase, but allowing a section of code to better assert what it wants out of a rollback.

Perhaps a callback handler for on error would be better?

```java

interface DataTransaction
{
  void onRollbackError() throws OpenDcsDataException;
  <T> boolean rollbackSupported(Class<T> connectionType);
}
```

This would allow for a simpler flow of logic

```
try(tx = db.newTransaction();)
{
   tx.setOnRollbackError((e) -> /*perform appropriate action*/)
   operation1
   operation2
}
catch (On Noo...)
{
   tx.rollback();
   other handling
}
```

I've see this in java.util.Stream operations.

Given the use of a callback that does imply perhaps an abstract class, or at least middle implementation, vs interface since all impls would need a handler.


But given your follow up and example, I think we're in agreement. We don't need to enforce a transaction, but we do need to make sure the user/system *know* if it was able to be rolled back or not.

adamkorynta: Yup, makes sense to me. Luckily, I don't see any SavePoints in the codebase. We'd want to avoid too many `boolean rollbackSupported(); boolean rollbackToSavePointSupported();....`

MikeNeilson: Yeah, i saw that in Connection, i think we're good with, "this just done" and "nope back it all off." At least for now. Interface wise, that's definitely easier to add later.

java/opendcs-api/src/main/java/org/opendcs/database/api/OpenDcsDatabase.java
```diff
@@ -20,4 +20,11 @@ public interface OpenDcsDatabase
      * @return A valid instance for this database, or an empty optional if the DAO is not supported
      */
     <T extends OpenDcsDao> Optional<T> getDao(Class<T> dao);
```
ktarbet, line 22: I"ll be watching for details inside OpenDcsDao

ktarbet (approved):

PR #807, opened 2024-12-05
correct token names.

MikeNeilson:
Used the wrong names for tokens.


MikeNeilson:
was able to manually run the workflow. still waiting for full results but it's gotten to the point of creating the distribution zip and tar at least.

MikeNeilson:
docker build was tested locally.

PR #820, opened 2024-12-09
Add Catch for Runtime Exception

### LINKED ISSUE #819: Runtime Exception error in Comp "run compuation" GUI uncaught
**Describe the bug**
When using the run computation GUI if algorithm contains a runtime error the user is never made aware of this error within a modal or dialog. This runtime error is never caught and the computation execute stops and displays data computed and saved at that point without informing the user.

**To Reproduce**
1. Go to the computations section of OpenDCS.
2. Click on the algorithms tab.
3. Select an algorithm which is known to have a runtime Error 
4. Click on the run Computation button
5. Click on the Execute Button 
6. The computation will run then stop without informing the user that a run time error has occurred.

**Expected behavior**
After Clicking on the Execute Button a user would expect a dialog to appear informing them that a runtime error has occurred

 - OS: Desktop Windows 10
 - Version: OpenDCS 7.014

**Additional context**
Under traditional circumstances we would assume that none of our algorithms contains runtime Exceptions, so this bug would never be caught or needed to be fix. However, when developing new algorithm this is extremely useful and important.

oskarhurst:
## Problem Description

Fixes #819 

## Solution

Describe your solution.

Added a catch within AW_AlgorithmBase class in the applyAlgorithm function. This catches runtimeExceptions and returns a DBCompException with an appropriate error message containing the type of error and where this error occurred.

## how you tested the change

I tested the fix by adding a RuntimeError to one of the algorithms and insuring that the appropriate dialog came up to inform the user of the exception and where it occurred.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/decodes/tsdb/algo/AW_AlgorithmBase.java
+++ b/java/opendcs/src/main/java/decodes/tsdb/algo/AW_AlgorithmBase.java
@@ -450,6 +450,9 @@ public abstract class AW_AlgorithmBase
 				}
 			}
 		}
+		catch(RuntimeException e){
+			throw new DbCompException("RunTime Error: "+e+"\n At: "+e.getStackTrace()[0].toString(), e);
+		}
 		finally
 		{
 			Logger.instance().setMinLogPriority(defLogPriority);


MikeNeilson (changes requested):
Generally excepted practice says this is a terrible idea in java but since a total "we can't fix this at all problem" would just trickle up again anyways I'm all for it. Plus we already do this in a few other places for "library level" runtime exceptions.

However, use `ex`, not e for the exception.

oskarhurst:
> Generally excepted practice says this is a terrible idea in java but since a total "we can't fix this at all problem" would just trickle up again anyways I'm all for it. Plus we already do this in a few other places for "library level" runtime exceptions.
> 
> However, use `ex`, not e for the exception.

NOTE For The Future: Exceptions within the GUI potion of openDCS are stored as "e" rather than "ex" https://github.com/opendcs/opendcs/blob/main/java/opendcs/src/main/java/decodes/tsdb/comprungui/CompRunGuiFrame.java#L1184-L1207 , perhaps in the future we can look at standardizing all exceptions to "ex" for consistency

MikeNeilson (approved):

PR #825, opened 2024-12-11
Computation run GUI bug fix and Run Comp standalone GUI Arguments

### LINKED ISSUE #821: "run computation" GUI, units of output values change when saving data
**Describe the bug**
When using the "run computation" GUI the units and labels for the output and input TimeSeries are set to those which are used within the computation algorithm which could either be the default of the units and labels which the user set to be input and output but could also be hardcoded defined units which are set within the algorithm. After the user executes their computation and clicks the "Save Output Data" button the units of only output time series which have labels hardcoded within the algorithm than that of which they are stored into the data base are then converted to the equivalent unit of that which is stored in the data base. While the units are correctly converted on save, the label of the units are never updated in the rendering, but the values of the data are.

**To Reproduce**
1. Go to the "run computation" GUI for your selected algorithm
2. Click execute
3. Click Save Output Data
4. if the units used within the algorithm are different that that saved to the database you will notice the values of the data to change but the label of the units will not change

**Expected behavior**
I would expect either the values of the data to never change and be re-rendered or for the label of the uniting to also change in the GUI

**Screenshots**
Before Saving Data-
runComputationGUI-beforeSave

After Saving Data-
runComputationGUI-afterSave

notice how the value of the data has changed for O:FTPK_Lower.EvapRate.Total.1hour but the label remains the same at "mm/hr"

 - OS: Desktop Windows 10
 - Version: OpenDCS 7.0.14

**Additional context**

oskarhurst:
## Problem Description

Fixes #821

TimeSeries table updates values of units but not labels of units on save button press in Comp Run GUI

## Solution

Describe your solution.
I set the table to completely re-render with new correct data and label on save button press.

secondary, added the feature where the user now has access to 3 new additional command line arguments -c "comp id or name" -S "since time where computation starts" -U "until time where computation ends". this match the structure of the compexec command. This is use full for faster testing so that the user does not have to reset parameters each time the GUI is opened.

## how you tested the change

Testing UI components manually by running computation which have a data unit mismatch for output timeseries and insures timeSeries table updates correctly.

## Where the following done:

do we want to update the legacy documentation for compexec?


java/opendcs/src/main/java/decodes/tsdb/comprungui/CompRunGuiFrame.java
```diff
@@ -1074,6 +1131,8 @@ private void runSelectedComps(Collection<DbComputation> compVector, Logger origi
 			@Override
 			public List<CTimeSeries> doInBackground()
 			{
+				myinputs.clear();
```
MikeNeilson, line 1134: This should probably be around line 1129 with the myoutputs.clear(). Just for consistency.

java/opendcs/src/main/java/decodes/tsdb/comprungui/CompRunGuiFrame.java
```diff
@@ -1135,6 +1194,7 @@ public List<CTimeSeries> doInBackground()
 							{
 								runme.addTimeSeries(ts);
 								inputs.add(ts);
+								myinputs.add(ts);
```
MikeNeilson, line 1197: I'm having a difficult time understanding why we need a 2nd array of the same inputs.

java/opendcs/src/main/java/decodes/tsdb/comprungui/RunComputationsFrameTester.java
```diff
@@ -1,7 +1,14 @@
 package decodes.tsdb.comprungui;
```
MikeNeilson, line 1: I know this existed already but would it make sense to just add the functionality to the main CompRunGuiFrame class?

MikeNeilson (changes requested):

java/opendcs/src/main/java/decodes/tsdb/comprungui/RunComputationsFrameTester.java
oskarhurst: To my knowledge(which I may be wrong about) is that runComputationsFrameTester extends TsdbAppTemplate, where CompRunGuiFrame extends TopFrame. So CompRunGuiFrame does not have the capability to be run is a stand alone mode, so it needs the TsdbAppTemplate to initialized the gui then open it.

MikeNeilson (approved):

PR #868, opened 2024-12-31
Refactor/rledit dbenum usage

MikeNeilson:
## Problem Description

Additional demonstration and usage of the new `OpenDcsDatabase::getDao` and `DataTransaction` system.

## Solution

1. Modified RefListFrame to extract Enumeration controls to own panel for separation of concerns.
2. Modified RefListEditor and RefListFrame to take an OpenDcsDatabase
3. Change ComboBox model to be DbEnum instead of string, modified usages to follow suite
4. Created functions as appropriate to control and retrieve state of the enumerations.
5. Where Database.getDb was directly used for Enum operations was changed to use new style
6. Saving of values was moved off GUI thread using technique from `RoutingSpecEditPanel` to report database saving exceptions to user (NOTE: will be turning that work flow into generic usage in separate PR. all but doInBackground and refernce value are duplicated.)
7. Saving of values is done in a transaction using the new style DAO.

## how you tested the change

Manually ran RlEdit to see original behavior still exists.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/decodes/rledit/RefListEditor.java
+++ b/java/opendcs/src/main/java/decodes/rledit/RefListEditor.java
@@ -1,6 +1,10 @@
 package decodes.rledit;
 
 import javax.swing.UIManager;
+
+import org.opendcs.database.DatabaseService;
+import org.opendcs.database.api.OpenDcsDatabase;
+
 import java.util.*;
 
 import ilex.gui.WindowUtility;
@@ -18,11 +22,13 @@ public class RefListEditor
     private static ResourceBundle genericLabels = null;
     private static ResourceBundle labels = null;
     boolean packFrame = false;
+    final private OpenDcsDatabase database;
 
     /** Construct the application. */
-    public RefListEditor() 
+    public RefListEditor(OpenDcsDatabase database)
     {
-        RefListFrame frame = new RefListFrame();
+        this.database = database;
+        RefListFrame frame = new RefListFrame(database);
         //Validate frames that have preset sizes
         //Pack frames that have useful preferred size info, e.g. from their layout
         if (packFrame) {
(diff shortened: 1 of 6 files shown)


java/opendcs/src/main/java/decodes/rledit/RefListEditor.java
```diff
@@ -97,19 +103,10 @@ public static void main(String[] args)
 
         DecodesSettings settings = DecodesSettings.instance();
         DecodesInterface.setGUI(true);
-
+        OpenDcsDatabase database = DatabaseService.getDatabaseFor("RefListEditor", settings);
+        database.getLegacyDatabase(Database.class).get().initializeForEditing();
         // Construct the database and the interface specified by properties.
         Database db = new decodes.db.Database();
```
ktarbet, line 109:
```suggestion
```

ktarbet, line 109: Database db = new decodes.db.Database();  // isn't used.

java/opendcs/src/main/java/decodes/rledit/RefListFrame.java
```diff
@@ -24,1489 +37,1186 @@
...
-		// right to close the window. We will catch the closing event and
-		// do the same thing as if user had hit File - Exit.
-		setDefaultCloseOperation(WindowConstants.DO_NOTHING_ON_CLOSE);
-		addWindowListener(
-			new WindowAdapter()
+			@Override
+			protected Boolean doInBackground() throws Exception
 			{
-				public void windowClosing(WindowEvent e)
+				Database db = Database.getDb();
```
ktarbet, line 955: Is the intent to remove all calls like this  Database.getDb() ?   I see several , that the github diff for this file isn't showing.

ktarbet (commented):
to be continued.

java/opendcs/src/main/java/decodes/rledit/RefListFrame.java
MikeNeilson: Eventually, at this time only the EnumDAO has been setup in the new style so only the operations on the enumerations are getting changed to keep the PR smaller.

ktarbet (approved):

PR #869, opened 2025-01-02
Disable documentation build, but don't fail the build itself.

MikeNeilson:
## Problem Description

The ant build allowed either disabling the documentation build, or ignoring the documentation build if python wasn't available.

## Solution

Provide the same behavior in the Gradle build.

## how you tested the change

Manually.
1. cleaned the build, set the `-Pno.docs=true` flag and verified no documentation was built and the distribution could still be created.
2. added a check for pythons existence and also disabled the build. (Check was done by intentionally misnaming the python executable defined at the top of the file.

## Where the following done:

### DIFF
--- a/docs/build.gradle
+++ b/docs/build.gradle
@@ -5,9 +5,24 @@ plugins {
 def pythonExe = osWin ? "python.exe" : "python3"
 def docsVenv = "$buildDir/sphinx"
 def docsDir = projectDir
-def venvPath = (osWin ? "${docsVenv}/Scipts" : "${docsVenv}/bin")
+def venvPath = (osWin ? "${docsVenv}/Scripts" : "${docsVenv}/bin")
+def noDocs = project.findProperty("no.docs") ?: false
+def havePython = checkPython(pythonExe)
 
-
+def checkPython(executableStr) {
+    def result = exec {
+        ignoreExitValue = true
+        if (osWin) {
+            executable "cmd"
+            args += "/c"
+        } else {
+            executable executableStr
+        }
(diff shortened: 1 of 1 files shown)


docs/build.gradle
```diff
@@ -5,9 +5,24 @@ plugins {
...
+            executable "cmd"
+            args += "/c"
+        } else {
+            executable executableStr
+        }
+        args += "-c"
+        args += "import sys; sys.exit(0)"
+    }
+    return result.getExitValue() == 0
+}
```
ktarbet, line 25:
```suggestion
def checkPython(pythonStr) {
    def result = exec {
        ignoreExitValue = true
        if (osWin) {
            executable "cmd"
            args += "/c"
            args += pythonStr
        } else {
            executable pythonStr
        }
        args += "-c"
        args += "import sys; print('checking for python'); sys.exit(0)"
    }
    return result.getExitValue() == 0
}
```

ktarbet: I was seeing error in the output

-c is not a valid command; added pythonStr command to args , consistent with unix.

MikeNeilson: Oops, thanks.

ktarbet:
It is also possible for pip to not be installed; but this check is a good improvement.

ktarbet (approved):

MikeNeilson:
> It is also possible for pip to not be installed; but this check is a good improvement.

I consider that, but decided it should be a failure. Mostly because I didn't want to add yet-another check.

PR #872, opened 2025-01-03
Alarm Email testing

ktarbet:
This PR is to add testing for the AlarmMailer, so we can have more confidence in upgrading  [com.sun.mail:mailapi](https://github.com/eclipse-ee4j/mail) from 1.6.2 to 2.0.1.  as suggested by dependabot.


java/opendcs/src/test/java/decodes/tsdb/alarm/mail/AlarmMailerTest.java
```diff
@@ -0,0 +1,84 @@
...
+import decodes.tsdb.alarm.EmailAddr;
+import decodes.sql.DbKey;
+
+
+public class AlarmMailerTest {
+
+	private static GreenMail smtpServer;
+	private AlarmMailer alarmMailer;
+	private Properties props;
+    private final String fromAddr = "karl@opendcs.org";
```
MikeNeilson, line 27: I would use the standard example.org, or whatever the "valid" non-existent domain is. If we ever did setup email for the domain, we're going to get a lot of junk at some point.

java/opendcs/src/test/java/decodes/tsdb/alarm/mail/AlarmMailerTest.java
```diff
@@ -0,0 +1,84 @@
...
+    private final static String fromName = "Karl";
```
MikeNeilson, line 28: Name doesn't match the second email below it.

MikeNeilson (changes requested):
Looks great, only concern is the use of an actual valid domain in test code.

MikeNeilson (approved):

MikeNeilson (approved):

PR #873, opened 2025-01-06
make Java languageVersion dynamic in opendcs.java-conventions.gradle

ktarbet:
languageVersion was fixed at java 8.

I could not compile on java 11.

The github actions don't need this change for some reason.  

Thanks @oskarhurst  for assistance on debugging this.

### DIFF
--- a/buildSrc/src/main/groovy/opendcs.java-conventions.gradle
+++ b/buildSrc/src/main/groovy/opendcs.java-conventions.gradle
@@ -24,7 +24,9 @@ java {
 
 tasks.withType(JavaCompile).configureEach {
     javaCompiler = javaToolchains.compilerFor {
-        languageVersion = JavaLanguageVersion.of(8)
+        languageVersion = JavaLanguageVersion.of(
+            JavaVersion.current().majorVersion.toInteger()
+        )
     }
 }
 
--- a/buildSrc/src/main/groovy/opendcs.java-conventions.gradle
+++ b/buildSrc/src/main/groovy/opendcs.java-conventions.gradle
@@ -22,12 +22,9 @@ java {
     withJavadocJar()
 }
 
-tasks.withType(JavaCompile).configureEach {
-    javaCompiler = javaToolchains.compilerFor {
-        languageVersion = JavaLanguageVersion.of(
-            JavaVersion.current().majorVersion.toInteger()
-        )
-    }
+tasks.withType(JavaCompile) {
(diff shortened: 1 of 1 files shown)


MikeNeilson (commented):
This breaks the output by targetting the class files at your current JDK level instead of the required output of java 8.

it's possible using "release" or compat is better than what's currently done; however, this is not correct for the project needs.

ktarbet:
Is there another way to compile with java 11?   Here is what I'm trying:

C:\project\opendcs>echo %JAVA_HOME%
C:\bin\jdk-11.0.20.1+1

C:\project\opendcs>where java
C:\bin\jdk-11.0.20.1+1\bin\java.exe

C:\project\opendcs>gradlew opendcs:jar

> Configure project :
The 'sonarqube' task depends on compile tasks. This behavior is now deprecated and will be removed in version 5.x. To avoid implicit compilation, set property 'sonar.gradle.skipCompile' to 'true' and make sure your project is compiled, before analysis has started.
The 'sonar' task depends on compile tasks. This behavior is now deprecated and will be removed in version 5.x. To avoid implicit compilation, set property 'sonar.gradle.skipCompile' to 'true' and make sure your project is compiled, before analysis has started.

> Configure project :docs
checking for python

> Configure project :install
Using app launcher_start

FAILURE: Build failed with an exception.

* Where:
Build file 'C:\project\opendcs\integrationtesting\opendcs-tests\build.gradle' line: 39

* What went wrong:
A problem occurred evaluating project ':testing:opendcs-tests'.
> Could not resolve all dependencies for configuration ':opendcs:runtimeClasspath'.
   > Failed to calculate the value of task ':opendcs:compileJava' property 'javaCompiler'.
      > Cannot find a Java installation on your machine matching this tasks requirements: {languageVersion=8, vendor=any, implementation=vendor-specific} for WINDOWS on x86_64.
         > No locally installed toolchains match and toolchain download repositories have not been configured.

ktarbet:
@MikeNeilson  Thanks for the idea to set sourceCompatibility and targetCompatibility  instead.  
Will this work?

MikeNeilson:
Yes, I was able to compile with a 17 compiler and verify the class file version was the correct output, and the javadocs compiled without errors.

MikeNeilson (approved):
Though I swear there was a reason I went with the tool chain. However, clearly that requires that the actual JDK be present, not just a compatible JDK.

PR #874, opened 2025-01-07
Added decorators

tlelv:
## Changes

* Added Algorithm, Input, Output, Override, and Propertyspec decorators.
* Added respective imports. 
* Removed AW comments and methods.

### DIFF
--- a/java/opendcs/src/main/java/decodes/tsdb/algo/RelativeHumidity.java
+++ b/java/opendcs/src/main/java/decodes/tsdb/algo/RelativeHumidity.java
@@ -12,13 +12,10 @@ import org.opendcs.annotations.algorithm.Output;
 public class RelativeHumidity extends AW_AlgorithmBase
 {
 
-	@Input
-	public double temperature;
-	@Input
-	public double dewPoint;
+	@Input public double temperature;
+	@Input public double dewPoint;
 
-	@Output
-	public NamedVariable output = new NamedVariable("output", 0);
+	@Output public NamedVariable output = new NamedVariable("output", 0);
 
 
 	// Allow javac to generate a no-args constructor.
@@ -26,20 +23,16 @@ public class RelativeHumidity extends AW_AlgorithmBase
 	/**
 	 * Algorithm-specific initialization provided by the subclass.
 	 */
+	@Override
 	protected void initAWAlgorithm() throws DbCompException
 	{
-//AW:INIT
 		_awAlgoType = AWAlgoType.TIME_SLICE;
-//AW:INIT_END
-
-//AW:USERINIT
-		// Code here will be run once, after the algorithm object is created.
-//AW:USERINIT_END
 	}
 	
 	/**
 	 * This method is called once before iterating all time slices.
 	 */
+	@Override
 	protected void beforeTimeSlices() throws DbCompException
 	{
 		
@@ -68,6 +61,7 @@ public class RelativeHumidity extends AW_AlgorithmBase
 	 * @throws DbCompException (or subclass thereof) if execution of this
 	 *        algorithm is to be aborted.
 	 */
+	@Override
 	protected void doAWTimeSlice() throws DbCompException
 	{
 		if (isMissing(temperature) || isMissing(dewPoint))
@@ -97,6 +91,7 @@ public class RelativeHumidity extends AW_AlgorithmBase
 	/**
 	 * This method is called once after iterating all time slices.
 	 */
+	@Override
 	protected void afterTimeSlices() throws DbCompException
 	{
 
--- a/java/opendcs/src/main/java/decodes/tsdb/algo/Resample.java
+++ b/java/opendcs/src/main/java/decodes/tsdb/algo/Resample.java
@@ -18,68 +18,49 @@ import decodes.tsdb.IntervalIncrement;
 import decodes.tsdb.ParmRef;
 import decodes.tsdb.VarFlags;
 import decodes.tsdb.algo.AWAlgoType;
+import org.opendcs.annotations.PropertySpec;
+import org.opendcs.annotations.algorithm.Algorithm;
+import org.opendcs.annotations.algorithm.Input;
+import org.opendcs.annotations.algorithm.Output;\
 
-//AW:IMPORTS
-// Place an import statements you need here.
-//AW:IMPORTS_END
 
-//AW:JAVADOC
-/**
-Resample an input to an output with a different interval.
-Output must not be irregular.
-Input may be irregular or any interval greater than or less than the output.
- */
-//AW:JAVADOC_END
+@Algorithm(description ="Resample an input to an output with a different interval. Output must not be irregular. Input may be irregular or any interval greater than or less than the output." )
 public class Resample
 	extends decodes.tsdb.algo.AW_AlgorithmBase
 {
-//AW:INPUTS
-	public double input;	//AW:TYPECODE=i
-	String _inputNames[] = { "input" };
-//AW:INPUTS_END
+	@Input public double input;	//AW:TYPECODE=i
+	@Input String _inputNames[] = { "input" };
 
-//AW:LOCALVARS
-	/** The next output time */
 	private Date nextOutputTime = null;
 	private Date lastOutputTime = null;
 	private IntervalIncrement outputIncr = null;
 	private Date lastTimeSlice = null;
 
-//AW:LOCALVARS_END
+	@Output public NamedVariable output = new NamedVariable("output", 0);
+	@Output String _outputNames[] = { "output" };
 
-//AW:OUTPUTS
-	public NamedVariable output = new NamedVariable("output", 0);
-	String _outputNames[] = { "output" };
-//AW:OUTPUTS_END
-
-//AW:PROPERTIES
-	public String method = "interp";
-	String _propertyNames[] = { "method" };
-//AW:PROPERTIES_END
+	@PropertySpec public String method = "interp";
+	@PropertySpec String _propertyNames[] = { "method" };
 
 	// Allow javac to generate a no-args constructor.
 
 	/**
 	 * Algorithm-specific initialization provided by the subclass.
 	 */
+	@Override
 	protected void initAWAlgorithm( )
 		throws DbCompException
 	{
-//AW:INIT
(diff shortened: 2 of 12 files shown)


java/opendcs/src/main/java/decodes/tsdb/algo/RelativeHumidity.java
```diff
@@ -12,34 +12,27 @@
 public class RelativeHumidity extends AW_AlgorithmBase
 {
 
-	@Input
```
MikeNeilson, line 15: Our standard is to have these on the line above even for variables.

Sorry, we need to actually provide a proper style config to make this easier on people.

java/opendcs/src/main/java/decodes/tsdb/algo/Resample.java
```diff
@@ -18,68 +18,49 @@
...
+@Algorithm(description ="Resample an input to an output with a different interval. Output must not be irregular. Input may be irregular or any interval greater than or less than the output." )
 public class Resample
 	extends decodes.tsdb.algo.AW_AlgorithmBase
 {
-//AW:INPUTS
-	public double input;	//AW:TYPECODE=i
-	String _inputNames[] = { "input" };
-//AW:INPUTS_END
+	@Input public double input;	//AW:TYPECODE=i
+	@Input String _inputNames[] = { "input" };
```
MikeNeilson, line 32: _inputNames, _outputNames, and _propertyNames should just be deleted.

java/opendcs/src/main/java/decodes/tsdb/algo/ScalerAdder.java
```diff
@@ -10,96 +10,71 @@
...
-	public double input3;	//AW:TYPECODE=i
-	public double input4;	//AW:TYPECODE=i
-	public double input5;	//AW:TYPECODE=i
-	public double input6;	//AW:TYPECODE=i
-	public double input7;	//AW:TYPECODE=i
-	public double input8;	//AW:TYPECODE=i
-	public double input9;	//AW:TYPECODE=i
-	public double input10;//AW:TYPECODE=i
-	String _inputNames[] = { "input1", "input2", "input3", "input4", "input5", 
+	@Input public double input1;	//AW:TYPECODE=i
```
MikeNeilson, line 25: Okay, so on this one, ignore the comment about keeping `@Input` on the line above. While technically not the standard, in this case due to volume it's more readable.

java/opendcs/src/main/java/decodes/tsdb/algo/RunningAverageAlgorithm.java
```diff
@@ -37,39 +37,31 @@
...
-//AW:OUTPUTS_END
+	@Output public NamedVariable average = new NamedVariable("average", 0);
+	@Output String _outputNames[] = { "average" };
 
-//AW:PROPERTIES
-	public long minSamplesNeeded = 1;
-	public boolean outputFutureData = false;
-	String _propertyNames[] = { "minSamplesNeeded", "outputFutureData" };
-//AW:PROPERTIES_END
+	@PropertySpec public long minSamplesNeeded = 1;
```
MikeNeilson, line 62: Due to a limitation of the java annotation system, the default value can't be picked up by the annotation processor, each `@PropertySpec` needs to have the value set manually. Example below. All of the default values are a string representation of the value. Generally you can copy the value as in between some quotes.

```suggestion
	@PropertySpec(value = "1") public long minSamplesNeeded = 1;
```

java/opendcs/src/main/java/decodes/tsdb/algo/RunningAverageAlgorithm.java
```diff
@@ -122,26 +108,25 @@ protected void beforeTimeSlices()
 	 * @throws DbCompException (or subclass thereof) if execution of this
 	 *        algorithm is to be aborted.
 	 */
+	@Override
```
MikeNeilson, line 111: Thanks for adding these.

MikeNeilson (changes requested):
Okay, so this was definitely a good start for someone accustomed to python decorators vs java annotations (similar concept, very different implementation.)

There's comments on a few files of required changes, this applies to all unless it was stated otherwise. I didn't want to overwhelm you with that many individual comments of the same thing.

java/opendcs/src/main/java/decodes/tsdb/algo/RunningAverageAlgorithm.java
tlelv: When it's a string like (public String method = "interp";), should I add something like (@PropertySpec(String = "Interp") public blah blah

MikeNeilson: It would be the same `@PropertySpec(value = ""interp")` regardless of the type. (The system doesn't currently deal with the property type.

tlelv (commented):
Changes have been made

java/opendcs/src/main/java/decodes/tsdb/algo/RdbRating.java
```diff
@@ -81,50 +72,37 @@ public void clearTable()
...
-    boolean failIfNoTable = false;
-    String interp = "log"; // possibilities are log and linear
-    String filePrefix = "";
-    String fileSuffix = ".rdb";
-    String _propertyNames[] = { "exceedLowerBound", "tableDir", "exceedUpperBound",
-        "applyShifts", "interp",
-        "filePrefix", "fileSuffix", "failIfNoTable" };
-//AW:PROPERTIES_END
+    @PropertySpec(value = "false")boolean exceedLowerBound = false;
+    @PropertySpec(value = "$DECODES_INSTALL_DIR/rdb")String tableDir = "$DECODES_INSTALL_DIR/rdb";
```
MikeNeilson, line 79: The `@PropertySpec` here should be above for readability. Notice that the actual variable declarations are a bit... all over the place.

MikeNeilson: Also when it does make sense to put on same line, make sure there's a space after the ).

java/opendcs/src/main/java/decodes/tsdb/algo/RelativeHumidity.java
```diff
@@ -12,10 +12,13 @@
...
 
-	@Input public double temperature;
-	@Input public double dewPoint;
+	@Input
+	public double temperature;
+	@Input
+	public double dewPoint;
 
-	@Output public NamedVariable output = new NamedVariable("output", 0);
+	@Output(type = Double.class)
```
MikeNeilson, line 20: This is arguably "better", but Double is the default type and thus not actually required.

However; I suggest leaving in as it does make our future intention of supporting additional output types more clear.

java/opendcs/src/main/java/decodes/tsdb/algo/TabRating.java
```diff
@@ -97,18 +97,16 @@ public void clearTable()
...
-	@PropertySpec public boolean exceedUpperBound = false;
-	@PropertySpec public String tableNameSuffix = ".tab";
-	@PropertySpec String interp = "log"; // possibilities are log and linear
-	@PropertySpec public String nametype = "";
-	@PropertySpec public String _propertyNames[] = { "exceedLowerBound", "tableDir", "tableName", 
-			"exceedUpperBound", "tableNameSuffix", "interp", "nametype" };
+	@Output(type = Double.class)
+	public NamedVariable dep = new NamedVariable("dep", 0);
+
+	@PropertySpec(value = "false") public boolean exceedLowerBound = false;
```
MikeNeilson, line 103: These should also be on line above. Or at least it would be nice if things just "read better".

Since you probably haven't looked at this in a few hours just scan it with your eyes again, and let your brain mull on what might make sense. I don't actually have a specific direction so just attempt some judgement if you think of something that seems better than spec on line above.

MikeNeilson (commented):
Okay, I have a super nitpick. I'm going to leave this as "comment" and not approved or changes required since everything is technically correct.

@ktarbet and @oskarhurst can provide the final judgement of "were the heck that annotation goes."

tlelv:
I made the judgement calls so there are two versions since its just a formatting issue.

I chose to keep things consistent except for both of the "ScalerAdder" files. They just seem to be the format-obnoxious exceptions.

tlelv (commented):
Still leaving it to you, Oskar, and Karl for the call but the fixes definitely weren't a challenge to change

MikeNeilson (approved):

java/opendcs/src/main/java/decodes/tsdb/algo/VirtualGage.java
```diff
@@ -8,72 +8,61 @@
...
+import org.opendcs.annotations.algorithm.Input;
+import org.opendcs.annotations.algorithm.Output;
+
+
+@Algorithm(description = "Compute a virtual elevation at an intermediate point between two other gages.\n" +
+"Inputs are the upstream and downstream elevation,\n" +
+"Properties specify Upstream position (e.g. Mile number), Downstream position, and Virtual Gage position.\n" +
+"The positions are required to do proper interpolation. Default values place the virtual gage halfway between up & downstream gages.\n" +
+"If provided, you may set gagezero properties for each of the locations, thus the output can be in gage height or elevation.
+")
```
MikeNeilson, line 22: This `"` needs to be on the line above at the end of the string.

MikeNeilson: @tlelv This still needs fixing.

MikeNeilson (changes requested):
Try doing a `./gradlew clean build` locally, there are still build failures.

It also looks like PythonAlgorithm has some quirks that make this difficult, I would suggest just reverting for now.

tlelv:
I apologize for taking a while to get back to this, would a better solution be to take a more recent, compiling version of openDCS and add the decorators to that?

MikeNeilson:
It appears to be all issues with your changes, definitely rebase or merge from the latest main if you haven't already, that's good practice, but then look at the build output and see what's actually wrong.

For one, it appears to just be a missing import, others may required more investigation.

PR #885, opened 2025-01-14
Strawman work on RawMessages as a stream.

MikeNeilson:
Okay, so there's definitely more details to work out here, but I've been thinking about how we could make use of the `java.util.streams` library to cleanup the flow of the routing spec processing.

The current logic of RoutingSpecThread is
1. Duplicated a surprising amount
2. A smidge spaghetti with the data
3. Uses exceptions for flow control, which is not great in java.
4. Basically processing a "stream" of messages, just in an archaic way.

This is incomplete and currently built as a strawman for discussion.

But the basic principal is:
- DataSourceExec returns a Stream of RawMessage (technically a wrapped message for error handling)
- To process this appropriate filters and mappings, such as to decodeMessage (which has it's own logic duplication issue), and output message
- Use a collector to handle pulling values from the stream, updating various statistics measurements, return a reason for stream processing to end.

While this mostly just eliminates an exclicit while loop, it has the added benefit of making it very difficult to embedded application (RoutinSpecThread, MessageBrowser) in the main flow. However, this usages could easily include peek operations on the stream to extract information desired during that usage.

The nice thing about this is that while the current getRawMessage could use some design improvements it integrations with this new scheme in a way that wouldn't interfere with existing code.

I'd like to gather some thoughts on the topic and if there's more general agreement I'll start working on something more concrete and useful.

### DIFF
--- a/java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
+++ b/java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
@@ -20,8 +20,14 @@
 */
 package decodes.datasource;
 
+import java.util.Iterator;
 import java.util.Properties;
+import java.util.Spliterators;
 import java.util.Vector;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+import org.opendcs.utils.FailableResult;
 import java.util.concurrent.DelayQueue;
 import java.util.concurrent.Delayed;
 import java.util.concurrent.TimeUnit;
@@ -267,6 +273,42 @@ public abstract class DataSourceExec implements PropertiesOwner
 	*/
 	protected abstract RawMessage getSourceRawMessage()
 		throws DataSourceException;
+
+	public Stream<FailableResult<RawMessage,DataSourceException>> getRawMessages() throws DataSourceException
+	{
+		return StreamSupport.stream(Spliterators.spliteratorUnknownSize(new Iterator<FailableResult<RawMessage,DataSourceException>>()
+		{
+			private FailableResult<RawMessage,DataSourceException> nextMessage = null;
+			@Override
+			public boolean hasNext() 
+			{
+				try
+				{
+					nextMessage = FailableResult.success(getRawMessage());
+					if (nextMessage != null)
+					{
+						return true;
+					}
+				}
+				catch (DataSourceException ex)
+				{
+					if (!(ex instanceof DataSourceEndException))
+					{
+						nextMessage =FailableResult.failure(ex);
+						return true;
+					}
+				}
+				return false;
+			}
+
+			@Override
+			public FailableResult<RawMessage,DataSourceException> next()
+			{
+				return nextMessage;
+			}
+
+		}, 0), false );
+	}
 
 	/**
 	  Find the matching transport medium for this platform.
--- a/java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
+++ b/java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
@@ -274,17 +274,26 @@ public abstract class DataSourceExec implements PropertiesOwner
 	protected abstract RawMessage getSourceRawMessage()
 		throws DataSourceException;
 
-	public Stream<FailableResult<RawMessage,DataSourceException>> getRawMessages() throws DataSourceException
+	/**
+	 * Provide a stream of DataMessage that can be chained to other operations.
+	 * 
+	 * The stream provided by the default implementation closes only when the source throws DataSourceEndException.
+	 * Other Errors are provided in the Error Result so that callers can design how to handle errors.
+	 * 
+	 * @return Either a valid DataMessage or the exception thrown by getSourceRawMessage.
+	 * @throws DataSourceException exception DataSourceEndException, which terminates the stream.
+	 */
+	public Stream<FailableResult<DataMessage,DataSourceException>> stream() throws DataSourceException
 	{
-		return StreamSupport.stream(Spliterators.spliteratorUnknownSize(new Iterator<FailableResult<RawMessage,DataSourceException>>()
+		return StreamSupport.stream(Spliterators.spliteratorUnknownSize(new Iterator<FailableResult<DataMessage,DataSourceException>>()
 		{
-			private FailableResult<RawMessage,DataSourceException> nextMessage = null;
+			private FailableResult<DataMessage,DataSourceException> nextMessage = null;
 			@Override
 			public boolean hasNext() 
 			{
 				try
 				{
-					nextMessage = FailableResult.success(getRawMessage());
+					nextMessage = FailableResult.success(getDataMessage());
 					if (nextMessage != null)
 					{
 						return true;
@@ -294,7 +303,7 @@ public abstract class DataSourceExec implements PropertiesOwner
 				{
 					if (!(ex instanceof DataSourceEndException))
 					{
-						nextMessage =FailableResult.failure(ex);
+						nextMessage = FailableResult.failure(ex);
 						return true;
 					}
 				}
@@ -302,7 +311,7 @@ public abstract class DataSourceExec implements PropertiesOwner
 			}
 
 			@Override
-			public FailableResult<RawMessage,DataSourceException> next()
+			public FailableResult<DataMessage,DataSourceException> next()
 			{
 				return nextMessage;
 			}
--- a/java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
+++ b/java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
@@ -20,9 +20,7 @@
 */
 package decodes.datasource;
 
-import java.util.Iterator;
 import java.util.Properties;
-import java.util.Spliterators;
 import java.util.Vector;
 import java.util.stream.Stream;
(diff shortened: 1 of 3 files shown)


java/opendcs/src/test/java/decodes/datasource/DataSourceTest.java
```diff
@@ -0,0 +1,175 @@
...
+        @Override
+        public RawMessage getRawMessage() throws DataSourceException
+        {
+            if (index >= messages.length)
+            {
+                throw new DataSourceEndException("No more data.");
+            }
+            if (index > 0)
+            {
+                /* outputing message with some delay to see the affects */
```
ktarbet, line 100:
```suggestion
                /* outputting message with some delay to see the affects */
```

ktarbet (commented):
Could this Streaming be useful to 'mock' a DataSource for testing purposes?

ktarbet (commented):
Looks good to me, I'm not a streaming expert , but can see the 'flow'

MikeNeilson:
> Looks good to me, I'm not a streaming expert , but can see the 'flow'

that last statement is what i looking for, that it overall makes sense. probably be a bit before i dig back into it but good motivation.

MikeNeilson:
As this doesn't currently affect operations, I would like merge it in as-is and then continue improvements until we eventually cut over usages to stream based.

java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
```diff
@@ -268,6 +274,51 @@ waiting for a message and the timeout (as defined in the properties
...
+	{
+		return StreamSupport.stream(Spliterators.spliteratorUnknownSize(new Iterator<FailableResult<DataMessage,DataSourceException>>()
+		{
+			private FailableResult<DataMessage,DataSourceException> nextMessage = null;
+			@Override
+			public boolean hasNext() 
+			{
+				try
+				{
+					nextMessage = FailableResult.success(getDataMessage());
```
adamkorynta: This doesn't look right if you can `hasNext()` twice in a row, it should give you the same message instead of pulling the next message.

java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
```diff
@@ -268,6 +274,51 @@ waiting for a message and the timeout (as defined in the properties
...
+					if (nextMessage != null)
```
adamkorynta: I don't think `FailableResult.success` can return null. I think the return true would only be if the actual message is not null?

java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
```diff
@@ -268,6 +274,51 @@ waiting for a message and the timeout (as defined in the properties
...
+	 * 
+	 * The stream provided by the default implementation closes only when the source throws DataSourceEndException.
+	 * Other Errors are provided in the Error Result so that callers can design how to handle errors.
+	 * 
+	 * @return Either a valid DataMessage or the exception thrown by getSourceRawMessage.
+	 * @throws DataSourceException exception DataSourceEndException, which terminates the stream.
+	 */
+	public Stream<FailableResult<DataMessage,DataSourceException>> stream() throws DataSourceException
+	{
+		return StreamSupport.stream(Spliterators.spliteratorUnknownSize(new Iterator<FailableResult<DataMessage,DataSourceException>>()
```
adamkorynta: Why not a custom Spliterator instead of an Iterator?

java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
```diff
@@ -268,6 +274,51 @@ waiting for a message and the timeout (as defined in the properties
...
 		throws DataSourceException;
 
+	/**
+	 * Provide a stream of DataMessage that can be chained to other operations.
+	 * 
+	 * The stream provided by the default implementation closes only when the source throws DataSourceEndException.
+	 * Other Errors are provided in the Error Result so that callers can design how to handle errors.
+	 * 
+	 * @return Either a valid DataMessage or the exception thrown by getSourceRawMessage.
+	 * @throws DataSourceException exception DataSourceEndException, which terminates the stream.
```
adamkorynta: `stream()` itself shouldn't be throwing `DataSourceException`

java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
```diff
@@ -268,6 +274,51 @@ waiting for a message and the timeout (as defined in the properties
...
+	 */
+	public Stream<FailableResult<DataMessage,DataSourceException>> stream() throws DataSourceException
```
adamkorynta: Streams are closeable, should resources get cleaned up on close?

adamkorynta: The contract of Stream allows for parallel, should probably document or figure out how to disallow parallel given this is required to be sequential message processing

java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
```diff
@@ -268,6 +274,51 @@ waiting for a message and the timeout (as defined in the properties
...
+				{
+					nextMessage = FailableResult.success(getDataMessage());
+					if (nextMessage != null)
+					{
+						return true;
+					}
+				}
+				catch (DataSourceException ex)
+				{
+					if (!(ex instanceof DataSourceEndException))
```
adamkorynta: This implies that all other `DataSourceException`'s are recoverable

java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
adamkorynta: Asked ChatGPT to turn into a spliterator for an example. I think that helps to simplify the hasNext vs next method implementations:

```
public Stream<FailableResult<DataMessage, DataSourceException>> stream()
{
    Spliterator<FailableResult<DataMessage, DataSourceException>> spliterator =
        new AbstractSpliterator<>(
            Long.MAX_VALUE,
            Spliterator.ORDERED | Spliterator.NONNULL)
        {
            private boolean finished;

            @Override
            public boolean tryAdvance(
                Consumer<? super FailableResult<DataMessage, DataSourceException>> action)
            {
                Objects.requireNonNull(action);

                if (finished)
                {
                    return false;
                }

                try
                {
                    DataMessage message = getDataMessage();

                    if (message == null)
                    {
                        return false;
                    }

                    action.accept(FailableResult.success(message));
                    return true;
                }
                catch (DataSourceEndException ex)
                {
                    finished = true;
                    return false;
                }
                catch (DataSourceException ex)
                {
                    action.accept(FailableResult.failure(ex));
                    return true;
                }
            }
        };

    return StreamSupport.stream(spliterator, false);
}
```

adamkorynta (changes requested):
I do think the hashNext vs next contract and null handling should be corrected before merge

java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
MikeNeilson: Probably yes, though we'll have to see how that interacts with the current DataSource setup.

MikeNeilson: Thanks, shouldn't be parallel, or at least should be non-parallel by default with a given DataSource able to state what it can do.

java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
MikeNeilson: I vaguely remember there was a reason, though it might have been because I thought of the iterator first.

java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
MikeNeilson: Yeah, that would be an issue. Will have to find a way to add a test for that.

java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
MikeNeilson: Yeah, that was probably left over from copied boiler plate. Thanks for pointing that out.

java/opendcs/src/main/java/decodes/datasource/DataSourceExec.java
MikeNeilson: That is up to the implementation, where this one *is* a definitive end. Hence why otherwise we just return the error result and let the usage decide.

But that's a fair point, probably good to have this particular portion of the handling be provided by the given DataSource so things can properly make sense for each.

MikeNeilson (commented):
Thanks, was sure I was missing things, not sure when I'll get back to it, but will likely move the Spliterator/Iterator/Stream Creation code to it's own class and see about shifting some of the error handling to allow the DataSources to have better control.

adamkorynta (approved):

MikeNeilson:
@adamkorynta Thanks! I suspect some things are still wrong. But it's at least a start.

Once #2077 also merged in we'll have a bit better setup to start experimenting with the behavior and integration.

PR #890, opened 2025-01-17
Decodes/CsvFunction bug fix

tlelv:
CsvFunction would not properly work with addMissing Function. 
Adds tests relating to this function


MikeNeilson (approved):

java/opendcs/src/main/java/decodes/decoder/CsvFunction.java
```diff
@@ -133,7 +133,7 @@ private void processColumn(int col, String sample, DecodedMessage decmsg, int li
 		if (sensorNumber < 0)
 			return;
 		trace("Processing column " + col + " value='" + sample + "'");
-		if (sample.length() == 0 || sample.startsWith("M") || sample.startsWith("/"))
```
ktarbet, line 136: I think we should keep 'M' as a default missing symbol.  This feels like a change in behavior removing 'M' from the check.

To keep behavior the DecodesScript.isMissingSymbol method could check for 'M' or or the list of missingSymbols in DecodesScript should be initialized with a 'M'

MikeNeilson: Yep, I honestly thought that was already there but I just checked an apparently not.

PR #894, opened 2025-01-22
fixes #856

ktarbet:
## Problem Description

Updating mail dependency to Jakarta Mail 1.6.7 (last one for Java 8) 

Fixes  #856

unit tests were added previously for the mailapi.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -16,7 +16,6 @@ jep="2.4.1"
 jama="1.0.2"
 dropwizard="4.1.0-rc2"
 passay="1.3.1"
-sun-mailapi="1.6.7"
 jsch="0.1.55"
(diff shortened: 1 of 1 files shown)


gradle/libs.versions.toml
```diff
@@ -16,7 +16,7 @@ jep="2.4.1"
 jama="1.0.2"
 dropwizard="4.1.0-rc2"
 passay="1.3.1"
-sun-mailapi="1.6.2"
+sun-mailapi="1.6.7"
```
MikeNeilson, line 19: put the cant go higher than 8 msg here, or move to the same section as jdbi.

MikeNeilson (changes requested):

MikeNeilson (approved):

PR #898, opened 2025-01-24
Add DatabaseIO Method Implementations for REST API PresentationResource Controller

zack-rma:
## Problem Description

Methods to support the refactored PresentationResources controller in the REST API are needed.

## Solution

Adds methods and makes changes necessary to support the refactored PresentationResources controller in the REST API.

## how you tested the change

Tested with testing harness in REST API against OpenTSDB instance. The test harness changes are not present in this PR.

Integration testing to come if needed.

## Where the following done:


java/opendcs/src/main/java/decodes/sql/SqlDatabaseIO.java
```diff
@@ -665,6 +665,26 @@ public synchronized void readDataTypeSet(DataTypeSet dts)
         }
     }
 
+    /**
+     Performs a lookup for a matching data-type object based on the data type code.
+     @param dtCode the data type code to look up
+     @return the data type object or null if not found
+     */
+    @Override
+    public synchronized DataType lookupDataType(String dtCode)
```
adamkorynta, line 674: Why add this method to `SqlDatabaseIO` when you can just get access to it from the `DataTypeDAI`?

adamkorynta (changes requested):

java/opendcs/src/main/java/decodes/db/DatabaseIO.java
```diff
@@ -399,6 +406,17 @@ public abstract void deletePresentationGroup( PresentationGroup pg )
 	public abstract Date getPresentationGroupLMT(PresentationGroup pg)
 		throws DatabaseException;
 
+	/**
+	 * If the presentation group referenced by groupId is used by one or more routing
+	 * specs, return a list of routing spec IDs and names. If groupId is not used,
```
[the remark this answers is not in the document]
zack-rma: Updated doc language

java/opendcs/src/main/java/decodes/sql/SqlDatabaseIO.java
zack-rma: Removed methods from the IO classes, updated controller to use `DataTypeDAI` implementation directly

java/opendcs/src/main/java/decodes/db/DatabaseIO.java
```diff
@@ -399,6 +398,17 @@ public abstract void deletePresentationGroup( PresentationGroup pg )
 	public abstract Date getPresentationGroupLMT(PresentationGroup pg)
 		throws DatabaseException;
 
+	/**
+	 * If the presentation group referenced by groupId is used by one or more routing
+	 * specs, return a string concatenated list of routing spec IDs and names. If groupId is not used,
```
adamkorynta, line 403: Don't return a concatenated list, let the calling API decide how to represent the list

java/opendcs/src/main/java/decodes/db/DatabaseIO.java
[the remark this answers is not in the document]
zack-rma: Updated return type to a list of RoutingSpec objects

java/opendcs/src/main/java/decodes/db/DatabaseIO.java
zack-rma: Returned as list of objects, moved string conversion to controller

java/opendcs/src/main/java/decodes/db/DatabaseIO.java
```diff
@@ -399,6 +398,17 @@ public abstract void deletePresentationGroup( PresentationGroup pg )
 	public abstract Date getPresentationGroupLMT(PresentationGroup pg)
 		throws DatabaseException;
 
+	/**
+	 * If the presentation group referenced by groupId is used by one or more routing
+	 * specs, return a list of routing specs populated with only IDs and names. If groupId is not used,
+	 * return null.
```
adamkorynta, line 404: we shouldn't return null collections (and also the implementations don't return null)

java/opendcs/src/main/java/decodes/db/DatabaseIO.java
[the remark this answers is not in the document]
adamkorynta: you didn't fill out all of the metadata for a RoutingSpec object. This function should just be returning `List<String>` with the names.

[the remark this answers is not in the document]
zack-rma: This is a convention of these DAO methods to specify partially-read objects via the javadocs. I'll update the javadocs to reflect this.

adamkorynta (approved):

PR #899, opened 2025-01-24
Add DatabaseIO Method Implementations for REST API NetlistResources Controller

zack-rma:
## Problem Description

Some functionality used by the refactored REST API controller is not present in the Toolkit.

## Solution

Adds new methods to support the REST API's NetlistResources controller.

## how you tested the change

Tested via test harness within REST API against OpenTSDB instance. The test harness is not included in these changes.

Integration test coverage will occur within the REST API.

## Where the following done:


java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
```diff
@@ -244,6 +244,112 @@ public void read(NetworkListList nll)
...
+				continue;
+			}
+			NetworkListEntry nle = new NetworkListEntry(
+					nl, transportId);
+			if (getDatabaseVersion() >= DecodesDatabaseVersion.DECODES_DB_11)
+			{
+				nle.setPlatformName(rs_nle.getString(3));
+				nle.setDescription(rs_nle.getString(4));
+			}
+			else // Pre version 11 derived name and description from platform & site.
```
adamkorynta, line 325: @MikeNeilson correct me if I'm wrong, but we shouldn't support versions before 11 in new code

MikeNeilson: bit of a judgement call, but I certainly wouldn't hold anyone to doing so.

adamkorynta: The `DecodesDatabaseVersion` doc says 11 is from OpenDCS 6.1 so wouldn't operating on the latest version of OpenDCS require the schema to be updated?

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
```diff
@@ -244,6 +244,112 @@ public void read(NetworkListList nll)
 			nl.addEntry(nle);
 		}
 	}
+
+	/**
+	 Reads the NetworkList and NetworkListEntry tables.
+	 @param nll the list of network lists
+	 @param tmType the transport medium type to filter on
+	 */
+	public void read(NetworkListList nll, String tmType)
```
adamkorynta, line 253: refactor `read(NetworkListList)` to call into this method with a null/wildcard `tmType`

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
MikeNeilson: Only to get any of the newest features. We're definitely going to be changing the policy a bit but in the past OpenDCS had always provided support for all previous database schema versions. Hence all the checks you see.

This is very useful for upgrades so we'll be keep backwards compatibly for a few schema version, but yeah, new code especially doesn't need to support anything past the "68" schema.

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
zack-rma: Refactored as discussed

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
```diff
@@ -162,16 +162,32 @@ private void initDb(DatabaseObject dbObj)
...
+			if (tmType.equalsIgnoreCase("goes"))
+			{
+				qtmt = "'goes', 'goes-self-times', 'goes-random'";
+			}
+			else
+			{
+				qtmt = "'" + tmType + "'";
+			}
+
+			q = q + " WHERE lower(transportMediumType) IN (" + qtmt + ")";
```
adamkorynta, line 188: should lower both sides in case parameterized `tmType` is capitalized

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
```diff
@@ -209,6 +225,10 @@ public void read(NetworkListList nll)
 					+ id + ", ignored.");
 				continue;
 			}
+			if (tmType != null && !nl.transportMediumType.equalsIgnoreCase(tmType))
```
adamkorynta, line 228: if `tmType` is `"goes"` and `transportMediumType` is `"goes-self-times"` or `"goes-random"`, won't this skip incorrectly?

java/opendcs/src/main/java/decodes/xml/XmlDatabaseIO.java
```diff
@@ -748,6 +748,18 @@ public void readNetworkListList( NetworkListList nll )
 		}
 		catch(java.io.IOException e) { }
 	}
+
+	/**
+	 * Returns the list of NetworkList objects defined in this database.
```
adamkorynta, line 753: javadoc does not reflect the implementation

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
MikeNeilson: Any concept of correctness here is definitely going to need a larger discussion.
But the NetworkLists themselves should only be doing a minimum of filtering. But any of us I'd have to really dig into this IO and how it's used to really know.

zack-rma: I've added some extra logic to check for `goes`, `goes-self-times`, and `goes-random`.

java/opendcs/src/main/java/decodes/xml/XmlDatabaseIO.java
zack-rma: Updated javadoc

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
zack-rma: lowered qtmt variable

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
adamkorynta: good to know thanks! very generous to support that many versions back

java/opendcs/src/main/java/decodes/xml/XmlDatabaseIO.java
adamkorynta: The revisions are very confusing, it should just say it isn't implemented

zack-rma: Fixed

adamkorynta (approved):

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
```diff
@@ -162,16 +162,32 @@ private void initDb(DatabaseObject dbObj)
...
+			if (tmType.equalsIgnoreCase("goes"))
+			{
+				qtmt = "'goes', 'goes-self-times', 'goes-random'";
+			}
+			else
+			{
+				qtmt = "'" + tmType + "'";
+			}
+
+			q = q + " WHERE lower(transportMediumType) IN (" + qtmt.toLowerCase() + ")";
```
adamkorynta, line 188: bind variable

adamkorynta (changes requested):

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
zack-rma: Replaced with bind variable

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
```diff
@@ -173,29 +175,37 @@ public void read(NetworkListList nll, String tmType)
...
-		if (rs != null) 
+		try (Connection conn = connection();
+			 PreparedStatement stmt = conn.prepareStatement(q))
 		{
-			while (rs.next()) 
+			if (tmType != null)
+			{
+				stmt.setString(1, qtmt);
+			}
+			ResultSet rs = stmt.executeQuery();
```
adamkorynta, line 201: try-with-resources

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
```diff
@@ -204,73 +214,80 @@ public void read(NetworkListList nll, String tmType)
...
 		q = "SELECT " + nleColumns
-		  + " FROM NetworkListEntry a, NetworkList b "
-		  + "WHERE a.networkListId = b.id ";
-		
-		ResultSet rs_nle = stmt.executeQuery( q );
+			+ " FROM NetworkListEntry a, NetworkList b "
+			+ "WHERE a.networkListId = b.id ";
 
-		while (rs_nle != null && rs_nle.next()) 
+		try (Statement stmt = createStatement())
```
adamkorynta, line 225: try-with-resources on the ResultSet

 is this sharing the same connection as the query above?

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
```diff
@@ -204,73 +214,80 @@ public void read(NetworkListList nll, String tmType)
...
-			NetworkList nl = _networkListList.getById(id);
-			if (nl == null)
-			{
-				Logger.instance().log(Logger.E_WARNING,
-				  "Orphan network list entry with invalid network list ID "
-					+ id + ", ignored.");
-				continue;
-			}
-			if (tmType != null)
+			while(rs_nle != null && rs_nle.next())
```
adamkorynta, line 229: ResultSet can never be null right?

adamkorynta (changes requested):

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
MikeNeilson: I believe it can.

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
zack-rma: Put in try-with-resources

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
zack-rma: Added to try-with-resources and reused connection from above

java/opendcs/src/main/java/decodes/sql/NetworkListListIO.java
adamkorynta: doesn't make any sense. for sure not from Oracle JDBC, but I'm not as familiar with Postgres drivers

MikeNeilson: well, I have been known to be wrong from time-to-time. And in fact I am: https://docs.oracle.com/javase/7/docs/api/java/sql/Statement.html#executeQuery%28java.lang.String%29

zack-rma: Removed null check

adamkorynta (approved):

PR #900, opened 2025-01-24
Add ComputationDAI Method Implementations for REST API ComputationResources Controller

zack-rma:
## Problem Description

The newly refactored ComputationResources controller in the REST API requires methods not implemented in the Toolkit.

## Solution

Adds methods to support the functionality of the ComputationResources controller in the REST API.

## how you tested the change
Integration tested using a test harness in the REST API against an OpenTSDB instance. The test harness is not included in these changes.

## Where the following done:


java/opendcs/src/main/java/decodes/tsdb/CompRefFilter.java
```diff
@@ -0,0 +1,270 @@
...
+package decodes.tsdb;
+
+import java.util.function.Predicate;
+import decodes.util.DecodesSettings;
+
+/**
+ * A filter used by the Comp Ref retrieval method of the REST API. This filter is used
+ * to filter the Db Computations that show up in the response.
+ */
+public class CompRefFilter
```
MikeNeilson, line 17: I feel like this should derived from `Predicate<DbComputation>`. I mean it's certainly fine now, but why limit the filter concept if we don't really have to?

@adamkorynta

MikeNeilson (changes requested):
This one will very likely get used soonish in GUI upgrades and should have tests.

The current integration tests load some computations, no need to go crazy just basic verification. 

Alternatively non-integration tests of the `CompRefFilter` behavior; that class is fairly complex.

java/opendcs/src/main/java/decodes/tsdb/CompRefFilter.java
adamkorynta: agreed, I would leave this class within the REST API and have the DAO interface take in a `Predicate<DbComputation>`. If future work is going to use similar behavior then a refactor can be done.  As the class stands. As the class stands it has too much unnecessary mutability and other cleanup that should be done.

MikeNeilson: a few more things:
1. make the method name generic. it's just getting comps matching a filter.
2. create a public static final filter that matches all for future default usage

the is basically the same as `listCompsForGUI` but the semantics of just providing a predicate is more clear.

zack-rma:
> This one will very likely get used soonish in GUI upgrades and should have tests.
> 
> The current integration tests load some computations, no need to go crazy just basic verification.
> 
> Alternatively non-integration tests of the `CompRefFilter` behavior; that class is fairly complex.

I've added basic test coverage of the retrieval method with filtering.

java/opendcs/src/main/java/decodes/tsdb/CompRefFilter.java
adamkorynta: I don't like #2 as that should just be a method overload that doesn't take in a filter. Alternatively `()->true` is much more transparent versus `defaultFilter` which requires knowledge of implementation.

java/opendcs/src/main/java/opendcs/dao/ComputationDAO.java
```diff
@@ -589,6 +589,34 @@ public ArrayList<DbComputation> listCompsForGUI(CompFilter filter)
...
+	 */
+	@Override
+	public List<DbComputation> listComps(Predicate<DbComputation> filter)
+			throws DbIoException
+	{
+		debug1("listComps " + filter);
+
+		if (compCache.size() == 0)
+		{
+			fillCache();
```
adamkorynta, line 605: I didn't realize the toolkit operated on an internal cache within the DAO layer. I do not see any cache invalidation being done, which means the REST API will never obtain updated results during the application lifecycle.

MikeNeilson: .. yeah, checking by size it not correct here. What i've been doing with the other DAOs is moving to a model of calling fillCache, and having fillCache check the last time it was loaded. (There are several that do that time check in multiple places.)

java/opendcs/src/main/java/decodes/tsdb/CompRefFilter.java
zack-rma: Swapped the `defaultFilter` variable for `() -> true`

java/opendcs/src/main/java/opendcs/dao/ComputationDAO.java
zack-rma: I've swapped the empty cache check for a time check. It looks like there are other DAOs that still do an empty cache check before filling the cache.

MikeNeilson: I'm not surprised. There's, to put it simply, a lot of work to be done to improve caching.

java/opendcs/src/main/java/opendcs/dao/ComputationDAO.java
```diff
@@ -589,6 +594,34 @@ public ArrayList<DbComputation> listCompsForGUI(CompFilter filter)
...
+	 * @param filter the computation filter containing site, algorithm, datatype, interval, process, and group names
+	 * @return List of computations
+	 */
+	@Override
+	public List<DbComputation> listComps(Predicate<DbComputation> filter)
+			throws DbIoException
+	{
+		debug1("listComps " + filter);
+
+		if (System.currentTimeMillis() - lastCacheFill > CACHE_TIME_LIMIT)
```
MikeNeilson, line 608: put the time check inside fillCache. makes it easier for the other methods to do the same without duplicating logic.

integrationtesting/opendcs-tests/src/test/java/org/opendcs/regression_tests/CompDependsDaoTestIT.java
MikeNeilson, line 1: Wrong class, this is for CompDepends, you're testing Computations. It's okay to make a new test class to keep the groupings consistent.

java/opendcs/src/main/java/opendcs/dao/ComputationDAO.java
zack-rma: Moved time check into fillCache

integrationtesting/opendcs-tests/src/test/java/org/opendcs/regression_tests/CompDependsDaoTestIT.java
zack-rma: Created a new test class and moved test case there

java/opendcs/src/main/java/opendcs/dao/ComputationDAO.java
```diff
@@ -171,88 +173,98 @@ public Connection getConnection()
 
 	private void fillCache()
 	{
-		debug1("ComputationDAO.fillCache()");
+		if (System.currentTimeMillis() - lastCacheFill > CACHE_TIME_LIMIT)
```
adamkorynta, line 176: invert condition and return in order to avoid pyramid of doom

adamkorynta (changes requested):

java/opendcs/src/main/java/opendcs/dao/ComputationDAO.java
zack-rma: Inverted condition and added return

adamkorynta (approved):

MikeNeilson (approved):

adamkorynta (approved):

PR #901, opened 2025-01-24
Update DatabaseIO Method Implementations for REST API ConfigResources Controller

zack-rma:
## Problem Description

The newly-refactored ConfigResources controller in the REST API needs method alterations to successfully retrieve data and report errors appropriately.

## Solution

Modifies methods used by the controller to properly return data and throw appropriate errors.

## how you tested the change

## Where the following done:


MikeNeilson (commented):
The overall design here is fine; however, I think it's going to be over-come-by-events as we transition everything to the DAO model. 

So this might end up causing more work than it's worth at the moment. Going approve the run and let the tests reveal any issues before I decide but I am inclined to just close this PR at the moment.

Nothing wrong with the change, just timing with other work.

java/opendcs/src/main/java/decodes/xml/XmlDatabaseIO.java
```diff
@@ -1409,13 +1408,14 @@ public void deletePlatform( Platform p ) throws DatabaseException
...
 	 */
-	public void readConfig( PlatformConfig pc ) throws DatabaseException
+	public PlatformConfig readConfig( PlatformConfig pc ) throws DatabaseException
 	{
 		String fn = "";
 		try
 		{
 			fn = makePath(PlatformConfigDir, pc.makeFileName());
 			myParser.parse(new File(fn), pc);
+			return pc;
```
adamkorynta, line 1418: you are returning the variable that was already parameterized. This is breaking the contract outlined in the javadoc that is specifying the input parameter will be modified to contain the full config.

adamkorynta (changes requested):

java/opendcs/src/main/java/decodes/xml/XmlDatabaseIO.java
zack-rma: In my testing, the parameterized variable is only modified to contain the DbKey and is otherwise empty. It may work as intended within the Toolkit, but does not appear to have the same behavior here. I could only get data to populate when the variable was returned.

adamkorynta (approved):

java/opendcs/src/main/java/decodes/sql/ConfigListIO.java
```diff
@@ -358,17 +359,21 @@ public PlatformConfig readConfig(DbKey id)
             stmt = createStatement();
 
             String q = "SELECT id, name, description, equipmentId " +
-                       "FROM PlatformConfig WHERE ID = " + id;
+                       "FROM PlatformConfig WHERE ID = " + pc.getId();
```
adamkorynta, line 362: bind variable

adamkorynta (changes requested):

java/opendcs/src/main/java/decodes/sql/ConfigListIO.java
zack-rma: Fixed

java/opendcs/src/main/java/decodes/sql/ConfigListIO.java
```diff
@@ -352,33 +353,27 @@ private void readConfigSensorProps(PlatformConfig cfg, Statement stmt)
     public void readConfig(PlatformConfig pc)
         throws DatabaseException, SQLException
     {
-        Statement stmt = null;
+        String q = "SELECT id, name, description, equipmentId " +
+                "FROM PlatformConfig WHERE ID = (?)";
```
adamkorynta, line 357: remove parenthesis

zack-rma: Removed

adamkorynta (approved):

adamkorynta (approved):

PR #902, opened 2025-01-24
Updated DataSourceListIO Method Error Handling for DataSourceResource controller in REST API

zack-rma:
## Problem Description 

The data retrieval methods in the DatasourceResource controller in the REST API need to be able to identify when a requested resource is not present in the database.

## Solution

Updated retrieval methods in DataSourceListIO with a throwable cause so the REST API can identify a not found error.

## how you tested the change

Integration tested in the REST API using a testing harness against an OpenTSDB instance. The test harness is not included in these changes.

Integration testing coverage will be present in the REST API.

## Where the following done:


java/opendcs/src/main/java/decodes/sql/DataSourceListIO.java
```diff
@@ -203,8 +204,11 @@ public DataSource readDS(DbKey id)
 
 			// There should be only one row in the result set
 			if (rs == null || !rs.next())
+			{
+				Throwable throwable = new ValueNotFoundException("No DataSource found with id " + id);
```
MikeNeilson, line 208: We're probably better off just changing ValueNotFound to derive from DatabaseException. 

I just checked and DatabaseException does derive from DecodesException.

MikeNeilson (changes requested):

java/opendcs/src/main/java/decodes/sql/DataSourceListIO.java
zack-rma: I've changed ValueNotFound to derive from DatabaseException and updated the usage

adamkorynta (approved):

MikeNeilson (approved):

PR #907, opened 2025-01-30
Add DatabaseIO Method Implementations for REST API PlatformResources Controller

zack-rma:
## Problem Description

The newly-refactored PlatformResources controller in the REST API needs method alterations to successfully retrieve data.

## Solution

Modifies methods used by the controller to properly return data.

## how you tested the change

## Where the following done:


java/opendcs/src/main/java/decodes/sql/PlatformListIO.java
```diff
@@ -230,6 +230,130 @@ public void read(PlatformList platformList)
...
+     *   <li>site</li>
+     *   <li>transportMedia - partial data</li>
+     *   <li>isReadComplete - should be false</li>
+     * </ul>
+     * The partial data of the TransportMedium table are the MediumType and
+     * MediumId fields.
+     * @param platformList the PlatformList object to populate
+     * @param tmType the transport medium type to filter on
+     */
+    public void read(PlatformList platformList, String tmType)
```
adamkorynta, line 254: `read(PlatformList)` should call into this method

java/opendcs/src/main/java/opendcs/dao/PlatformStatusDAO.java
```diff
@@ -65,6 +66,47 @@ public synchronized PlatformStatus readPlatformStatus(DbKey platformId)
 		}
 		return null;
 	}
+
+	@Override
+	public List<PlatformStatus> readPlatformStatusList(DbKey netlistId)
```
adamkorynta, line 71: Isn't this the same as `listPlatformStatus()` but with a filter?

zack-rma: Yes, I've refactored this method to use `listPlatformStatus()` if the filter is null

adamkorynta: `listPlatformStatus()` should call this method, not the other way around

java/opendcs/src/main/java/decodes/sql/PlatformListIO.java
```diff
@@ -154,24 +154,69 @@ private String getColTpl()
...
+                if (tmType.equals("goes"))
+                {
+                    filter = "goes, goes-self-timed, goes-random";
+                }
+                else
+                {
+                    filter = "goes";
+                }
+
+                q = q + " where exists(select PLATFORMID from TRANSPORTMEDIUM where lower(MEDIUMTYPE) IN ("
```
adamkorynta, line 217: should be using bind variables

adamkorynta (changes requested):

java/opendcs/src/main/java/opendcs/dao/PlatformStatusDAO.java
zack-rma: Swapped method call around

java/opendcs/src/main/java/decodes/sql/PlatformListIO.java
zack-rma: Added bind variable

java/opendcs/src/main/java/decodes/sql/PlatformListIO.java
```diff
@@ -154,27 +155,79 @@ private String getColTpl()
...
+                else
+                {
+                    filter = "goes";
+                }
+
+                q = q + " where exists(select PLATFORMID from TRANSPORTMEDIUM" +
+                        " where lower(MEDIUMTYPE) IN (?) and PLATFORM.ID = PLATFORMID)";
+            }
+
+            PreparedStatement stmt = conn.prepareStatement(q);
```
adamkorynta, line 223: preparedstatement needs to be in try-with-resources

java/opendcs/src/main/java/opendcs/dao/PlatformStatusDAO.java
```diff
@@ -65,6 +66,67 @@ public synchronized PlatformStatus readPlatformStatus(DbKey platformId)
...
+			if (db.isCwms())
+				q = q + " where platform_id in (select distinct a.id from platform a, "
+						+ "transportmedium b where a.id = b.platformid)";
+
+			ArrayList<PlatformStatus> ret = new ArrayList<PlatformStatus>();
+
+			if (db.getDecodesDatabaseVersion() < DecodesDatabaseVersion.DECODES_DB_11)
+				return ret;
+
+			ResultSet rs = doQuery(q);
```
adamkorynta, line 115: try-with-resources

java/opendcs/src/main/java/decodes/sql/PlatformListIO.java
zack-rma: Fixed

java/opendcs/src/main/java/opendcs/dao/PlatformStatusDAO.java
zack-rma: Fixed

adamkorynta (approved):

PR #909, opened 2025-02-03
Add Schedule Entry, Routing Spec, and Dacq Event Methods and Classes

zack-rma:
## Problem Description 

The newly-refactored RoutingResources controller in the REST API needs method alterations to successfully retrieve data. It also needs methods to properly store test data to verify functionality.

## Solution

Adds and modifies methods used by the controller to properly return data. Adds methods to Configuration class to allow for test data storage.

## how you tested the change

## Where the following done:

### DIFF
--- a/integrationtesting/opendcs-tests/src/test/java/org/opendcs/spi/configuration/Configuration.java
+++ b/integrationtesting/opendcs-tests/src/test/java/org/opendcs/spi/configuration/Configuration.java
@@ -1,13 +1,24 @@
+/*
+ * Copyright 2025 OpenDCS Consortium and its Contributors
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
 package org.opendcs.spi.configuration;
 
 import java.io.File;
 import java.util.Map;
 
-import decodes.db.DatabaseException;
 import decodes.db.DatabaseIO;
-import decodes.db.ScheduleEntryStatus;
-import decodes.polling.DacqEvent;
-import decodes.sql.DbKey;
 import decodes.util.DecodesSettings;
 
 import decodes.db.Database;
@@ -55,7 +66,7 @@ public interface Configuration
     public File getPropertiesFile();
     public File getUserDir();
     public boolean isSql();
-    default public boolean isTsdb()
+    public default boolean isTsdb()
     {
         return false;
     }
@@ -73,13 +84,13 @@ public interface Configuration
      * @return The timeseries database if it can be made.
      * @throws Throwable any issue with the creation of the TimeSeriesDb object
      */
-    default public TimeSeriesDb getTsdb() throws Throwable
+    public default TimeSeriesDb getTsdb() throws Throwable
     {
         return null;
     }
 
     /**
-     * Returns an independent instance of the {@decodes.db.Database} Decodes Database for this configuration.
+     * Returns an independent instance of the {@link decodes.db.Database} Decodes Database for this configuration.
      *
      * @return Instance of the Decodes Database for this run/test.
      * @throws Throwable
@@ -95,7 +106,7 @@ public interface Configuration
         return db;
     }
 
-    default public boolean implementsSupportFor(Class<? extends TsdbAppTemplate> appClass)
+    public default boolean implementsSupportFor(Class<? extends TsdbAppTemplate> appClass)
     {
         return false;
     }
@@ -108,18 +119,10 @@ public interface Configuration
     default public boolean supportsDao(Class<? extends DaoBase> dao)
     {
         return false;
-    };
+    }
 
     /* The name of this configuration
     * @return
     */
     public String getName();
-
-    void storeScheduleEntryStatus(ScheduleEntryStatus status) throws DatabaseException;
-
-    void deleteScheduleEntryStatus(DbKey scheduleEntryId) throws DatabaseException;
-
-    void storeDacqEvent(DacqEvent event) throws DatabaseException;
-
-    void deleteDacqEventForPlatform(DbKey platformId) throws DatabaseException;
 }
--- a/integrationtesting/opendcs-tests/src/test/java/org/opendcs/spi/configuration/Configuration.java
+++ b/integrationtesting/opendcs-tests/src/test/java/org/opendcs/spi/configuration/Configuration.java
@@ -1,29 +1,13 @@
-/*
- * Copyright 2025 OpenDCS Consortium and its Contributors
- * Licensed under the Apache License, Version 2.0 (the "License")
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- *  Unless required by applicable law or agreed to in writing, software
- *  distributed under the License is distributed on an "AS IS" BASIS,
- *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- *  See the License for the specific language governing permissions and
- *  limitations under the License.
- *
- */
-
 package org.opendcs.spi.configuration;
 
 import java.io.File;
 import java.util.Map;
 
-import decodes.db.DatabaseIO;
-import decodes.util.DecodesSettings;
-
 import decodes.db.Database;
+import decodes.db.DatabaseIO;
 import decodes.tsdb.TimeSeriesDb;
 import decodes.tsdb.TsdbAppTemplate;
+import decodes.util.DecodesSettings;
 import opendcs.dao.DaoBase;
 import uk.org.webcompere.systemstubs.environment.EnvironmentVariables;
 import uk.org.webcompere.systemstubs.properties.SystemProperties;
@@ -66,7 +50,7 @@ public interface Configuration
     public File getPropertiesFile();
     public File getUserDir();
     public boolean isSql();
-    public default boolean isTsdb()
+    default public boolean isTsdb()
     {
         return false;
     }
@@ -84,13 +68,13 @@ public interface Configuration
      * @return The timeseries database if it can be made.
      * @throws Throwable any issue with the creation of the TimeSeriesDb object
      */
-    public default TimeSeriesDb getTsdb() throws Throwable
+    default public TimeSeriesDb getTsdb() throws Throwable
     {
         return null;
     }
 
     /**
-     * Returns an independent instance of the {@link decodes.db.Database} Decodes Database for this configuration.
+     * Returns an independent instance of the {@decodes.db.Database} Decodes Database for this configuration.
      *
      * @return Instance of the Decodes Database for this run/test.
      * @throws Throwable
@@ -106,7 +90,7 @@ public interface Configuration
         return db;
     }
 
-    public default boolean implementsSupportFor(Class<? extends TsdbAppTemplate> appClass)
+    default public boolean implementsSupportFor(Class<? extends TsdbAppTemplate> appClass)
     {
         return false;
     }
@@ -119,7 +103,7 @@ public interface Configuration
     default public boolean supportsDao(Class<? extends DaoBase> dao)
     {
         return false;
-    }
+    };
 
     /* The name of this configuration
     * @return
--- a/integrationtesting/opendcs-tests/src/test/java/org/opendcs/spi/configuration/Configuration.java
+++ b/integrationtesting/opendcs-tests/src/test/java/org/opendcs/spi/configuration/Configuration.java
@@ -3,11 +3,16 @@ package org.opendcs.spi.configuration;
 import java.io.File;
 import java.util.Map;
 
-import decodes.db.Database;
+import decodes.db.DatabaseException;
 import decodes.db.DatabaseIO;
+import decodes.db.ScheduleEntryStatus;
+import decodes.polling.DacqEvent;
+import decodes.sql.DbKey;
+import decodes.util.DecodesSettings;
+
+import decodes.db.Database;
 import decodes.tsdb.TimeSeriesDb;
 import decodes.tsdb.TsdbAppTemplate;
-import decodes.util.DecodesSettings;
 import opendcs.dao.DaoBase;
 import uk.org.webcompere.systemstubs.environment.EnvironmentVariables;
 import uk.org.webcompere.systemstubs.properties.SystemProperties;
@@ -109,4 +114,12 @@ public interface Configuration
     * @return
     */
     public String getName();
+
+    void storeScheduleEntryStatus(ScheduleEntryStatus status) throws DatabaseException;
+
+    void deleteScheduleEntryStatus(DbKey scheduleEntryId) throws DatabaseException;
+
+    void storeDacqEvent(DacqEvent event) throws DatabaseException;
+
+    void deleteDacqEventForPlatform(DbKey platformId) throws DatabaseException;
 }
--- a/java/opendcs/src/main/java/decodes/db/DatabaseIO.java
+++ b/java/opendcs/src/main/java/decodes/db/DatabaseIO.java
@@ -1,4 +1,19 @@
 /*
+ * Copyright 2025 OpenDCS Consortium and its Contributors
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+/*
 *  $Id$
 */
 package decodes.db;
@@ -217,7 +232,7 @@ public abstract class DatabaseIO
 	and primary display attributes only).
 	  @param rsl the list to populate
 	*/
-	public abstract RoutingSpecList readRoutingSpecList(RoutingSpecList rsl)
+	public abstract void readRoutingSpecList(RoutingSpecList rsl)
 		throws DatabaseException;
 
 	/**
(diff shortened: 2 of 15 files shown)


integrationtesting/opendcs-tests/src/test/java/org/opendcs/spi/configuration/Configuration.java
```diff
@@ -109,4 +114,12 @@ default public boolean supportsDao(Class<? extends DaoBase> dao)
     * @return
     */
     public String getName();
+
+    void storeScheduleEntryStatus(ScheduleEntryStatus status) throws DatabaseException;
```
adamkorynta, line 118: These methods don't belong in this interface, you can get access to all these DAO's from the REST API already

java/opendcs/src/main/java/decodes/db/DatabaseIO.java
```diff
@@ -220,9 +217,19 @@ Objects in this list may be only partially populated (key values
 	and primary display attributes only).
 	  @param rsl the list to populate
 	*/
-	public abstract void readRoutingSpecList(RoutingSpecList rsl)
+	public abstract RoutingSpecList readRoutingSpecList(RoutingSpecList rsl)
```
adamkorynta, line 220: The input parameter is supposed to be filled by reference. Adding an output parameter is muddling the API.

integrationtesting/opendcs-tests/src/test/java/org/opendcs/spi/configuration/Configuration.java
MikeNeilson, line 118: These are not part of the "configuration" and you've rather oddly duplicated code.

Tt does make some sense to have them as a helper but they should be in their own class. I could also see an argument for `AppTestBase` as we do it that way for CDA, but Configuration's purpose is to setup the database access for use by the downstream tests and provide instances to said tests. It shouldn't handle calls like this.

java/opendcs/src/main/java/decodes/db/RoutingExecStatus.java
MikeNeilson, line 1: ... This was definitely needed. I'm glad you did it.

java/opendcs/src/main/java/decodes/sql/RoutingSpecListIO.java
```diff
@@ -176,6 +182,175 @@ public void read(RoutingSpecList rsList)
...
+        String q = "select rs.ID, se.NAME, se.SCHEDULE_ENTRY_ID,"
+                + " ses.SCHEDULE_ENTRY_STATUS_ID, ses.RUN_START_TIME,"
+                + " ses.RUN_COMPLETE_TIME, ses.LAST_MESSAGE_TIME, ses.HOSTNAME,"
+                + " ses.RUN_STATUS, ses.NUM_MESSAGES, ses.NUM_DECODE_ERRORS,"
+                + " ses.NUM_PLATFORMS, ses.LAST_SOURCE, ses.LAST_CONSUMER,"
+                + " ses.LAST_MODIFIED"
+                + " from ROUTINGSPEC rs, SCHEDULE_ENTRY_STATUS ses, SCHEDULE_ENTRY se"
+                + " where ses.SCHEDULE_ENTRY_ID = se.SCHEDULE_ENTRY_ID"
+                + " and se.ROUTINGSPEC_ID = rs.ID"
+                + " and se.SCHEDULE_ENTRY_ID = " + scheduleEntryId.getValue()
```
MikeNeilson, line 201: Dude! Bind variable!

java/opendcs/src/main/java/decodes/sql/RoutingSpecListIO.java
```diff
@@ -176,6 +182,175 @@ public void read(RoutingSpecList rsList)
...
+                + " order by ses.RUN_START_TIME desc";
+
+        try (Statement stmt = createStatement();
```
MikeNeilson, line 204: You're already using `try(Statement)`, change to PreparedStatement, this should have access to a Connection object.

MikeNeilson: You can also create a `DaoHelper` instance to use the new handlers.

java/opendcs/src/main/java/decodes/sql/RoutingSpecListIO.java
```diff
@@ -206,13 +381,19 @@ public void readRoutingSpec(RoutingSpec routingSpec) throws DatabaseException
         try (Statement stmt = createStatement();
              ResultSet resultSet = stmt.executeQuery(q);)
         {
+            Throwable thr = new ValueNotFoundException("No RoutingSpec found with id " + routingSpec.getId());
```
adamkorynta, line 384: creating an exception generates a stack trace which is not a trivial operation, only create the exception conditionally

java/opendcs/src/main/java/decodes/sql/SqlDatabaseIO.java
```diff
@@ -887,7 +887,7 @@ public synchronized void readRoutingSpecList(RoutingSpecList rsList)
         {
             conn = getConnection();
             _routingSpecListIO.setConnection(conn);
-            _routingSpecListIO.read(rsList);
+            return _routingSpecListIO.read(rsList);
```
adamkorynta, line 890: if `_routingSpecListIO.read` isn't updating by reference, then there is a bug in the implementation

java/opendcs/src/main/java/decodes/sql/SqlDatabaseIO.java
```diff
@@ -904,6 +904,45 @@ public synchronized void readRoutingSpecList(RoutingSpecList rsList)
         }
     }
 
+    @Override
+    public synchronized List<RoutingStatus> readRoutingSpecStatus() throws DatabaseException
+    {
+        try (Connection conn = getConnection())
+        {
+            _routingSpecListIO.setConnection(conn);
+           return _routingSpecListIO.readRoutingSpecStatus(makeLoadingAppDAO());
```
MikeNeilson, line 913: Move this loading app to the try-with so it gets closed.

java/opendcs/src/main/java/decodes/xml/XmlDatabaseIO.java
```diff
@@ -991,6 +993,24 @@ public void readPresentationGroupList( PresentationGroupList pgl ) throws Databa
 		}
 	}
 
+	/**
+	 * Returns the list of RoutingSpec statuses for the routing specs
+	 * stored to the database. XML implementation is not yet available.
+	 * @throws DatabaseException upon error
+	 */
+	public List<RoutingStatus> readRoutingSpecStatus()
```
MikeNeilson, line 1001: Shouldn't these have the `@Override`

java/opendcs/src/main/java/decodes/xml/XmlScheduleEntryDAO.java
```diff
@@ -122,6 +122,12 @@ public ScheduleEntry readScheduleEntry(String name) throws DbIoException
 		return null;
 	}
 
+	@Override
+	public ScheduleEntry readScheduleEntry(DbKey id) throws DbIoException
+	{
+		throw new UnsupportedOperationException("readScheduleEntry by ID not implemented for XML.");
```
MikeNeilson, line 128: Why not?

java/opendcs/src/main/java/decodes/sql/RoutingSpecListIO.java
```diff
@@ -176,6 +182,175 @@ public void read(RoutingSpecList rsList)
...
+                                ars.setAppName(appRef.getAppName());
+                                break;
+                            }
+            }
+            catch (DbIoException ex)
+            {
+                throw new DatabaseException("Cannot list computation apps", ex);
+            }
+
+            Collections.sort(ret,
```
adamkorynta, line 338: use the [Comparator.comparing API](https://www.baeldung.com/java-8-comparator-comparing)

java/opendcs/src/main/java/opendcs/dai/DacqEventDAI.java
```diff
@@ -93,6 +95,22 @@ public int readEventsForScheduleStatus(DbKey scheduleEntryStatusId, ArrayList<Da
...
+	 * @param evtList the list to add events to
+	 * @param appId the application ID
+	 * @param routingExecId the routing execution ID
+	 * @param platformId the platform ID
+	 * @param backlog the backlog
+	 * @param httpSession the HTTP session
+	 * @return number of events added to the list
+	 * @throws DbIoException
+	 */
+	public int readEvents(ArrayList<DacqEvent> evtList, DbKey appId, DbKey routingExecId, DbKey platformId, String backlog, HttpSession httpSession)
```
MikeNeilson, line 112: not that we can really get away with it but I'm okay with pretending this DAI doesn't actually exist.

java/opendcs/src/main/java/decodes/sql/RoutingSpecListIO.java
```diff
@@ -176,6 +182,175 @@ public void read(RoutingSpecList rsList)
...
+                RoutingStatus rstat = new RoutingStatus(DbKey.createDbKey(rs.getLong(1)));
+                rstat.setRoutingSpecId(DbKey.createDbKey(rs.getLong(1)));
+                rstat.setName(rs.getString(2));
+                ret.add(rstat);
+            }
+            q = "select se.SCHEDULE_ENTRY_ID, se.LOADING_APPLICATION_ID, se.ROUTINGSPEC_ID, "
+                    + "se.RUN_INTERVAL, se.ENABLED, se.NAME "
+                    + "from SCHEDULE_ENTRY se";
+
+            ResultSet resultSet = stmt.executeQuery( q );
```
adamkorynta, line 259: ResultSet needs to be in try-with-resources

java/opendcs/src/main/java/decodes/sql/RoutingSpecListIO.java
```diff
@@ -176,6 +182,175 @@ public void read(RoutingSpecList rsList)
...
+                }
+            }
+
+            q = "select se.SCHEDULE_ENTRY_ID, ses.LAST_MODIFIED, ses.NUM_MESSAGES, ses.NUM_DECODE_ERRORS, ses.LAST_MESSAGE_TIME"
+                    + " from SCHEDULE_ENTRY se, SCHEDULE_ENTRY_STATUS ses"
+                    + " where se.SCHEDULE_ENTRY_ID = ses.SCHEDULE_ENTRY_ID"
+                    + " and ses.SCHEDULE_ENTRY_STATUS_ID = "
+                    + "  (select max(SCHEDULE_ENTRY_STATUS_ID) from SCHEDULE_ENTRY_STATUS "
+                    + "   where SCHEDULE_ENTRY_ID = se.SCHEDULE_ENTRY_ID)";
+            resultSet = stmt.executeQuery( q );
```
adamkorynta, line 296: ResultSet needs to be in try-with-resources

java/opendcs/src/main/java/opendcs/dao/DacqEventDAO.java
```diff
@@ -236,6 +243,86 @@ public int readEventsForPlatform(DbKey platformId, ArrayList<DacqEvent> evtList)
...
+		if (routingExecId != null)
+		{
+			q = q + " " + c + " SCHEDULE_ENTRY_STATUS_ID = ?";
+			c = "AND";
+			parameters.add(routingExecId);
+		}
+		if (platformId != null)
+		{
+			q = q + " " + c + " PLATFORM_ID = ?";
+			c = "AND";
```
MikeNeilson, line 272: you might want to add the spaces around AND just in case something happens below. Nothing wrong with extra spaces, definite problems with missing spaces.

java/opendcs/src/main/java/decodes/db/RoutingExecStatus.java
```diff
@@ -0,0 +1,196 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package decodes.db;
+
+import java.util.Date;
+
+/**
+ * This bean holds the status info for a single execution of a routing spec.
```
adamkorynta, line 20: Need to distinguish between `RoutingSpecStatus`

MikeNeilson (changes requested):
Most of this is reasonable. the additional functions within Configuration seems rather... wonky to me.

adamkorynta (changes requested):

integrationtesting/opendcs-tests/src/test/java/org/opendcs/spi/configuration/Configuration.java
zack-rma: I've removed these methods and determined a way to access the data without them.

java/opendcs/src/main/java/decodes/sql/RoutingSpecListIO.java
zack-rma: Swapped for bind variable

java/opendcs/src/main/java/decodes/sql/RoutingSpecListIO.java
zack-rma: Swapped for PreparedStatement

java/opendcs/src/main/java/decodes/sql/SqlDatabaseIO.java
zack-rma: Moved

java/opendcs/src/main/java/decodes/xml/XmlDatabaseIO.java
zack-rma: Added overrides

java/opendcs/src/main/java/decodes/xml/XmlScheduleEntryDAO.java
zack-rma: I've implemented this method since it's only a few lines, but other DAO methods will not be getting XML implementations per the scope of work.

java/opendcs/src/main/java/opendcs/dao/DacqEventDAO.java
zack-rma: Added spaces

java/opendcs/src/main/java/decodes/db/DatabaseIO.java
zack-rma: Reverted to fill by reference

java/opendcs/src/main/java/decodes/sql/RoutingSpecListIO.java
zack-rma: Made exception generation conditional

java/opendcs/src/main/java/decodes/sql/SqlDatabaseIO.java
zack-rma: Reverted to update by reference

java/opendcs/src/main/java/decodes/sql/RoutingSpecListIO.java
zack-rma: Used Comparator API as suggested

java/opendcs/src/main/java/decodes/sql/RoutingSpecListIO.java
zack-rma: Moved to try-with-resources

java/opendcs/src/main/java/decodes/sql/RoutingSpecListIO.java
zack-rma: Moved

java/opendcs/src/main/java/decodes/db/RoutingExecStatus.java
zack-rma: Updated javadoc to represent differences

adamkorynta (approved):

java/opendcs/src/main/java/opendcs/dai/DacqEventDAI.java
```diff
@@ -23,6 +23,8 @@
 import java.util.ArrayList;
 import java.util.Date;
 
+import javax.servlet.http.HttpSession;
```
MikeNeilson, line 26: while already have a dependency that pulls this in, I don't think an HttpSession belongs in this DAI. Is there a more generic way to handle that information? a whole HttpSession risks a lot of information leakage given just how uncontrolled the calling code is, from the perspective of the web server.

java/opendcs/src/main/java/opendcs/dao/DacqEventDAO.java
```diff
@@ -236,6 +258,86 @@ public int readEventsForPlatform(DbKey platformId, ArrayList<DacqEvent> evtList)
...
+		{
+			q = q + " " + c + " PLATFORM_ID = ?";
+			c = " AND ";
+			parameters.add(platformId);
+		}
+		if (backlog != null && !backlog.trim().isEmpty())
+		{
+			if (backlog.equalsIgnoreCase("last"))
+			{
+				Object lastDacqEventId = httpSession.getAttribute(LAST_DACQ_ATTRIBUTE);
```
MikeNeilson, line 294: This does look somewhat useful generically; however, the specific attribute value should be passed in with the appropriate datatype. This class has no business understanding the contents of an HttpSession.

MikeNeilson (changes requested):

java/opendcs/src/main/java/opendcs/dai/DacqEventDAI.java
zack-rma: I've updated the handling so the HttpSession is not passed into the DAO

java/opendcs/src/main/java/opendcs/dao/DacqEventDAO.java
zack-rma: Removed HttpSession from DAO

java/opendcs/src/main/java/opendcs/dai/DacqEventDAI.java
```diff
@@ -150,4 +147,29 @@ public int readEventsAfter(DbKey eventId, ArrayList<DacqEvent> evtList)
 		throws DbIoException;
 
 
+	class DacqEventAttr
```
adamkorynta, line 150: don't love inner classes being defined in an interface.  This Object is used as an event id? shouldn't that be a number or String? The remove state also seems like it shouldn't be in the DAO unless the DAO is doing the removing.

zack-rma: I can move the inner class. The data value comes from the HttpSession in the controller as an Object, so I didn't want to cast it without knowing precisely what data is present. The remove state is tracked so the controller can determine whether the object should be removed from the HttpSession. I'm open to suggestions for better ways to handle this.

MikeNeilson (approved):
That's nice and generic.

java/opendcs/src/main/java/opendcs/dai/DacqEventDAI.java
zack-rma: Removed this class and refactored per our discussion

adamkorynta (approved):

PR #913, opened 2025-02-05
Add cache removal to TimeSeries deletion

zack-rma:
## Problem Description

When deleting TimeSeries, the data is removed from the database but not the cache. This causes the data to be returned after deletion until the cache is reloaded.

## Solution

Adds a removal-by-id for the cached TimeSeries.

## how you tested the change

Integration tested in REST API against OpenTSDB instance. Test harness code is not included here.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/opendcs/opentsdb/OpenTimeSeriesDAO.java
+++ b/java/opendcs/src/main/java/opendcs/opentsdb/OpenTimeSeriesDAO.java
@@ -63,7 +63,6 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Optional;
 import java.util.TimeZone;
 
 import org.opendcs.database.ExceptionHelpers;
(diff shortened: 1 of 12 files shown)


java/opendcs/src/main/java/opendcs/opentsdb/OpenTimeSeriesDAO.java
```diff
@@ -1342,6 +1343,34 @@ public ArrayList<TimeSeriesIdentifier> listTimeSeries()
...
+			TimeSeriesIdentifier tsid = tsidit.next();
+			if (tsid instanceof CwmsTsId)
+			{
+				CwmsTsId ctsid = (CwmsTsId) tsid;
+				if (!activeOnly || ctsid.isActive())
+				{
+					ret.add(tsid);
+				}
+			}
+			else if (!activeOnly || tsid.getSite().isActive())
```
zack-rma, line 1367: I'm unsure if this is an appropriate way to identify active status for non-CwmsTsId TimeSeriesIdentifiers, since they do not contain the `active` boolean

PR #928, opened 2025-02-19
Fixed PresentationGroup deletion from cache

zack-rma:
## Problem Description

Presentation groups were failing to be successfully removed from the cache upon deletion. This caused deleted groups to show up upon retrieval.

## Solution

Added removal from cache to presentation group deletion method.

## how you tested the change

Integration tested within the REST API against OpenTSDB and CWMS database instances.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/decodes/sql/PresentationGroupListIO.java
+++ b/java/opendcs/src/main/java/decodes/sql/PresentationGroupListIO.java
@@ -721,6 +721,15 @@ public class PresentationGroupListIO extends SqlDbObjIo
 
         q = "DELETE FROM PresentationGroup WHERE ID = " + id;
         executeUpdate(q);
+
+        if (pg.getDatabase().getDbIo() != null)
+        {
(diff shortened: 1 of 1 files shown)


java/opendcs/src/main/java/decodes/sql/PresentationGroupListIO.java
```diff
@@ -721,6 +721,15 @@ public void delete(PresentationGroup pg)
 
         q = "DELETE FROM PresentationGroup WHERE ID = " + id;
         executeUpdate(q);
+
+        if (pg.getDatabase().getDbIo() != null)
```
adamkorynta, line 725: Can't you do `_pgList.remove(pg);`?

zack-rma: Updated as suggested, tests show that does work.

adamkorynta (approved):

PR #930, opened 2025-02-20
Bugfix: Loading App deletion for CWMS database

zack-rma:
## Problem Description

Fixes https://github.com/opendcs/rest_api/issues/293. 

Deletion of a Loading App uses a query against the `ALARM_SCREENING` table with a field that is not present in the CWMS schema.

## Solution
Added a CWMS database check to skip the query on CWMS databases.

## how you tested the change

## Where the following done:


java/opendcs/src/main/java/opendcs/dao/LoadingAppDao.java
```diff
@@ -445,7 +445,7 @@ public void deleteComputationApp(CompAppInfo app)
                         + "assigned to it.");
             }
 
-            if (db.getDecodesDatabaseVersion() >= DecodesDatabaseVersion.DECODES_DB_68)
+            if (!db.isCwms() && db.getDecodesDatabaseVersion() >= DecodesDatabaseVersion.DECODES_DB_68)
```
MikeNeilson, line 448: This is a bug, the column shpuld be there, but will need to add it to  new database update migration. Simple alter table statememt if you want to just do it, but other wise just change the check to include cwms only db version greater than 6.8 would suffice.

MikeNeilson (changes requested):

java/opendcs/src/main/java/opendcs/dao/LoadingAppDao.java
zack-rma: Added check for database version higher than 6.8 for CWMS databases. My current database version showed 7.0, so I'm not sure this will work for all instances.

zack-rma: Is this versioning in progress? Or should I implement an alternative method of checking whether this column exists in the meantime?

MikeNeilson: Checking for the column probably makes the most sense. I'm not sure well I'll get to it.

zack-rma: Added column checks

adamkorynta: I recommend documenting a code with either a link to this discussion or an issue that describes how/when the column will get added.

MikeNeilson (approved):
Well, I think this can be a little cleaner, but I'm going to call it good enough for now.

At the least it's an improvement that will allow things to work and we can adjust after, or while, the tables are corrected

adamkorynta (approved):

PR #931, opened 2025-02-21
Fixed TimeSeries and TimeSeries Group deletion from cache

zack-rma:
## Problem Description

Deletion of timeseries and timeseries group objects only removes them from the database, not the cache. This results in deleted objects being returned upon retrieval.

## Solution

Adds removal from cache to both timeseries and timeseries group deletion methods.

## how you tested the change

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/opendcs/dao/TsGroupDAO.java
+++ b/java/opendcs/src/main/java/opendcs/dao/TsGroupDAO.java
@@ -460,6 +460,8 @@ public class TsGroupDAO
 
 		q = "DELETE from tsdb_group WHERE group_id = " + groupId;
 		doModify(q);
+
+		cache.remove(groupId);
 	}
 
 	@Override
--- a/integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/TimeSeriesDaoIT.java
+++ b/integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/TimeSeriesDaoIT.java
@@ -70,24 +70,24 @@ class TimeSeriesDaoIT extends AppTestBase
                 tsDao.saveTimeSeries(tsIn);
                 // Retrieve the TimeSeriesIdentifier that shouldn't been saved to the database.
                 // This will also fill in required metadata so that the retrieval operations are handled correctly.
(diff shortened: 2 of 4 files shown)


adamkorynta (approved):

java/opendcs/src/main/java/opendcs/dao/TsGroupDAO.java
```diff
@@ -460,6 +460,8 @@ public void deleteTsGroup(DbKey groupId)
 
 		q = "DELETE from tsdb_group WHERE group_id = " + groupId;
 		doModify(q);
+
+		cache.remove(groupId);
```
MikeNeilson, line 464: I really want to pressure you to move all those queries into a transaction, but I'm about to merge the big change anyways and I suspect it would make more work.

MikeNeilson (approved):

MikeNeilson:
It looks like you hit a similar rabbit hole I did went tweaking the caching, difficulties with timing of things during the integration tests... unfortunately I have no idea how to help you beyond that, your change looks perfectly reasonable.

zack-rma:
@MikeNeilson By integration test timing, are you referring to the failing Ubuntu runs? I'm also seeing failing tests only for Ubuntu here: https://github.com/opendcs/opendcs/pull/900

I'm planning on doing some more investigation, just wanted to see if you might have any insight before I do.

MikeNeilson:
no I kinda literally mean the timing within the test of computations itself. When I was updating how the caches were getting used in certain operations I had a hell of time with the tests because of how they were executed and what was or wasn't in process which effected when the cache was reloaded.

It seems deleting the item from the cache, which is definitely correct behavior, is causing issues with those tests. Which is rather odd; I was having issues with new timeseries *not* being in the cache.

zack-rma:
Huh, that is a strange issue.

Hopefully resolving these build errors will be relatively straightforward now that the behavior is correct. Fingers crossed.

zack-rma:
I've updated the test case that was failing - looks like it was trying to fill a timeseries after it had been deleted. This was working before, since the timeseries identifier was still in the cache and would result in an empty CTimeSeries object. I've updated the test to expect an exception since the identifier is now properly deleted.

MikeNeilson (dismissed):

PR #938, opened 2025-02-25
I can't see any usages of java\decodes\hdb\dbutils\DBAccess2.java

ktarbet:
java\decodes\hdb\dbutils\DBaccess2.java it is mostly a copy of DBAccess.java, and apparently not used?

## Problem Description

I was trying to fix some compiler warnings, and ran into DBAccess2.java that would be easier to delete than fix (assuming it isn't used?)  @agilmore2  what do you think?


AndrewPWRE:
I'm ok with removing this copy. Unfortunately, the main DBAccess code still has some uses outside this tree.

PR #943, opened 2025-02-26
drop decodes.hdb.dbutils.Logger

ktarbet:
When cleaning up javadocs, found this Logger that doesn't apper to be used.


MikeNeilson:
It would appear to be "well used".

agilmore2:
lol, thanks!

PR #945, opened 2025-02-27
more progress on javadoc warnings

ktarbet:


java/opendcs/src/main/java/org/opendcs/utils/properties/LrgsVariablesProvider.java
```diff
@@ -10,7 +10,7 @@
 
 /**
  * Handle variables required by Current LRGS reports.
- * See {@see DcsVariablesProvider} for justification as to way
+ * See {@link DcsVariablesProvider} for justification as to way
```
MikeNeilson, line 13: the text should also, say way twice when it should one `why`

MikeNeilson (changes requested):
though to be fair the one change is my own typo.

MikeNeilson (approved):

PR #949, opened 2025-02-28
Added Schedule Entry Retrieval by Status ID

zack-rma:
## Problem Description

When mapping data in the REST API, schedule entry data is needed from a schedule entry status identifier.

See: https://github.com/opendcs/rest_api/blob/3388e25b209884588134b00987daff563f09954d/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PlatformResources.java#L478

Used in: https://github.com/opendcs/rest_api/pull/346

## Solution

Added a method to retrieve a schedule entry object from the associated schedule entry status ID.

## how you tested the change

## Where the following done:


java/opendcs/src/main/java/opendcs/dao/ScheduleEntryDAO.java
```diff
@@ -466,6 +466,47 @@ public ArrayList<ScheduleEntryStatus> readScheduleStatus(
...
+	{
+		ArrayList<ScheduleEntryStatus> ret = new ArrayList<>();
+		if (db.getDecodesDatabaseVersion() < DecodesDatabaseVersion.DECODES_DB_10)
+		{
+			return null;
+		}
+
+		String q = "select " + sesColumns + " from " + sesTables + " where " + sesJoinClause;
+		ArrayList<Object> parameters = new ArrayList<>();
+		if (scheduleEntryStatusId != null && scheduleEntryStatusId != DbKey.NullKey)
```
MikeNeilson, line 486: how is "just give me the first one" valid in this context?

some tests with different data showing that would be useful for the discussion.

MikeNeilson (changes requested):

java/opendcs/src/main/java/opendcs/dao/ScheduleEntryDAO.java
zack-rma: I've refactored to reduce the number of calls to the DAO to get the data I'm actually after - the schedule entry object. I've added checks for multiple results coming back and also added a test to verify data is retrieved as expected.

java/opendcs/src/main/java/opendcs/dao/ScheduleEntryDAO.java
```diff
@@ -466,6 +467,50 @@ public ArrayList<ScheduleEntryStatus> readScheduleStatus(
...
+		{
+			return null;
+		}
+
+		String q = "select a.schedule_entry_id from " + sesTables + " where " + sesJoinClause
+				+ " and a.schedule_entry_status_id = ? order by a.run_start_time desc";
+
+		try
+		{
+			doQuery(q, rs ->
```
MikeNeilson, line 494: There is a `getSingleValue` method, and getSingleValueOr. Though I'm pretty sure getSingleValue returns null by default.

That seems like it would cleanup the code here vs using an ArrayList just to allow for a capture.

And since you know within the handler that there *MUST* be a row, you can just pass the result of DbKey.createDbkey to readScheduleEntry.

getSingleValue also handles error messaging for more than one value. (Though that's a decent specific error message so it might be worth updating that function to one that provides a default exception supplier and one that takes a specific one.

MikeNeilson (changes requested):
Test is reasonable.

Still concerned about the functions being available but not implemented in the XML database; definitely going to cause problems later. So FYI, the project will be updating the contributing language to avoid that in the future.

Getting there, I know I'm being really picky but we're also about to embark on a rather mass overhaul soon.

java/opendcs/src/main/java/opendcs/dao/ScheduleEntryDAO.java
zack-rma: Refactored to use the `getSingleValueOr` method for status retrieval. I also added an XML implementation, since it's just a few lines.

MikeNeilson (dismissed):
Thank you for providing the xml impl.

adamkorynta (approved):

PR #957, opened 2025-03-04
Fixes to Config and Platform Retrieval

zack-rma:
## Problem Description

Issues with retrieval of platforms and configs identified here: https://github.com/opendcs/rest_api/pull/362
Config items were not being correctly populated when retrieved. Platforms were not successfully removed from cache upon deletion, causing issues upon retrieval.

## Solution

Added cache removal to Platform deletion method. Added conditional copy of retrieved data to Configs.

## how you tested the change
Tested in REST API against OpenTSDB and CWMS instances.

## Where the following done:


java/opendcs/src/main/java/decodes/sql/ConfigListIO.java
```diff
@@ -372,7 +372,11 @@ public void readConfig(PlatformConfig pc)
                             String.format("No PlatformConfig found with ID %d", pc.getId().getValue()), thr);
                 }
 
-                putConfig(pc.getId(), rs);
+                PlatformConfig conf = putConfig(pc.getId(), rs);
+                if (!_dbio._isOracle)
```
MikeNeilson, line 376: why would all but Oracle need this and Oracle not?

MikeNeilson (changes requested):

java/opendcs/src/main/java/decodes/sql/ConfigListIO.java
zack-rma: I initially thought the error with retrieval I was running into was DB specific, but it seems there's some unreliable behavior differences between the usage of the method in the Toolkit and REST API. I'm currently investigating and hope to have a fix ready tomorrow.

zack-rma: I've updated the `readConfig` method with a content check. For some reason, data was correctly copied without the `pc.copyfrom()` within the Toolkit, but the REST API needed that line. The presence of the `copyfrom()` line was causing the tests in the Toolkit to fail, so it now checks whether the data was successfully copied before invoking the copy statement.

MikeNeilson: well. I definitely believe you, but looking at the code beyond what you changed that makes no sense.

Before I merge I'm going to step through it manually and see what I can see what's going on. It's definitely assigning the name from the ResultSet so super weird.

adamkorynta:
Fixing the PlatformConfig issue in PR: https://github.com/opendcs/opendcs/pull/958

MikeNeilson (approved):

PR #958, opened 2025-03-05
short circuit copyFrom tasks if input and this are same reference

adamkorynta:
## Problem Description

PlatformConfig was clearing the DecodesScript and ConfigSensors list when using the same reference. There are other classes that have similar copyFrom functionality that also may cause unintentional issues.

When operating on cached data, clearing state can inadvertently cause data to get cleared too early. Also, there is no need to perform redundant clone operations.

## Solution

Short circuit copyFrom if the same reference for all found methods.

## how you tested the change

Existing unit/integration tests should prove this change doesn't break anything.

## Where the following done:


adamkorynta:
Created this PR after diagnosing why DecodesTest was failing compared to REST API integration tests in: https://github.com/opendcs/opendcs/pull/957

MikeNeilson (approved):

PR #979, opened 2025-03-12
Initial algorithm setup for Hamon Evap method.

MikeNeilson:
## Problem Description

Fixes #1981. 

One of the USACE districts uses the Hamon Evap method. It is reasonable that other will also want to be able to use that method.

## Solution

Copied existing method in and upgrade formatting and certain design choices for project standards

## how you tested the change

Currently untested, hence the draft. Also waiting on updated algorithm test work from @oskarhurst 

## Where the following done:


MikeNeilson:
@krowvin can you sanity check the input and output data on this? I just pulled the last year's worth (well, September anyways, wanted to cross some DST boundaries.)

MikeNeilson:
Huh, things are passing in CWMS but not the Postgres impl, most likely a unit conversion error (probably with the langleys).

Still, it will fail in CWMS due to some data being missing, let me know which inputs/outputs are valid in regards to missing data.

UPDATE: yep, unit conversion, needed to use kph instead of km/h. CWMS was able to just figure it out.

integrationtesting/opendcs-tests/src/test/resources/data/Comps/EvapHamon/Test1/timeseries/expectedOutputs/data-Evap.tsimport
```diff
@@ -0,0 +1,244 @@
...
+2026/04/20-05:00:00, 0.3763488062615844, 0
+2026/04/21-05:00:00, 0.334658547353337, 0
+2026/04/22-05:00:00, 0.14402938520843916, 0
+2026/04/23-05:00:00, 0.06942190926692916, 0
+2026/04/24-05:00:00, 0.1404329570281935, 0
+2026/04/25-05:00:00, 0.07179078191110459, 0
+2026/04/26-05:00:00, 0.08643681908287813, 0
+2026/04/27-05:00:00, 0.2079476983426123, 0
+2026/04/28-05:00:00, 0.34881935809732295, 0
+2025/11/02-04:00:00, 48.86749999999999, 0
```
krowvin, line 241: This has an output but are the expected required inputs set?

krowvin (commented):
- The [DST-tail rows expect evap values even though solar radiation is blank](https://github.com/opendcs/opendcs/blob/ce7d41ff6b6bdc7d1d0c2f5c6cfe7a7d0f2a2f8f/integrationtesting/opendcs-tests/src/test/resources/data/Comps/EvapHamon/Test1/timeseries/expectedOutputs/data-Evap.tsimport#L241-L244
), and the 2026/03/08-05 and 2026/03/09-05 rows have all four inputs blank. Those rows do not sanity-check missing-data behavior; they encode output where the algorithm should not have enough inputs to calculate.
  
- [Several complete-input rows expect a blank output](https://github.com/opendcs/opendcs/blob/ce7d41ff6b6bdc7d1d0c2f5c6cfe7a7d0f2a2f8f/integrationtesting/opendcs-tests/src/test/resources/data/Comps/EvapHamon/Test1/timeseries/expectedOutputs/data-Evap.tsimport#L57), but the algorithm explicitly clamps negative daily evap to 0.0 and calls setOutput. My hand calculation for 2025/10/25, 2025/10/26, 2025/10/27, 2025/11/21, 2025/11/24, and 2026/03/05 produces negative millimeters before the clamp, so the expected value should be 0.0, not missing.

- [The fixture has relative humidity 100.07791666666667 on 2025/11/25, which makes es_ea negative](https://github.com/opendcs/opendcs/blob/ce7d41ff6b6bdc7d1d0c2f5c6cfe7a7d0f2a2f8f/java/opendcs/src/main/java/org/opendcs/algorithms/EvapHamon.java#L94); Math.pow(es_ea, 0.88) returns NaN in Java. The expected output hardcodes 0.07, but the current algorithm will not produce that reliably unless humidity is capped or this path is handled before the fractional power.

MikeNeilson:
@AndySWT can you take a look at this version of the Hamon Evap algorithm and the data used for testing.

Directly copied from code provided by @krowvin, so presumably what is currently running except for some name changes, and data from prod. The calculation itself appears to be correct; however, there are several points in time where your system has extra elements that probably shouldn't exist and I want to verify if that's true before I delete them from the test data set.

integrationtesting/opendcs-tests/src/test/resources/data/Comps/EvapHamon/Test1/timeseries/expectedOutputs/data-Evap.tsimport
MikeNeilson: Sorry, missed this, the inputs are in the "inputs" folder.

AndySWT:
I am satisfied that this algorithm is ready for merging.

MikeNeilson:
> I am satisfied that this algorithm is ready for merging.

@AndySWT the tests aren't passing due to descrepencies in the data. For the elements that are failing we need to confirm if the inputs are wrong, the actual output is wrong, or the expected output is wrong.

krowvin:
Spoke with members in our district there was some discussion on if the current algorithm was Hamon or another method. May need to rename this to the correct method. @AndySWT 

A followup would be needed for SWF to PR their method as well, they may have the correct Hamon Method. 
@SWF-WM

PR #991, opened 2025-03-21
Initial work on stateless variants of LoadingApp.

MikeNeilson:
## Problem Description

Fixes # 953. 

## Solution

Create states versions of methods in the interface, and implement them.
Also going to change the checkCompLock (new variant) to return a status object. Exception was being used for flow control; not a huge deal in this check but that's caused some major performance issues so we need to get out out of the habit.

## how you tested the change

Covered by existing tests

## Where the following done:


MikeNeilson:
we've started going in a different direction with the updated daos.

PR #1009, opened 2025-03-27
Current application launching. Fix logic error in EnumSqlDao.

MikeNeilson:
## Problem Description

Due to the various database changes, and a logic bug in enums, several applications would not start correctly.

## Solution

Ensure global database set after calls to getDatabaseFor.
Correct logic bug in EnumSQLDao

## how you tested the change

Run compedit directly, opened several applications through the launcher.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/decodes/launcher/LauncherFrame.java
+++ b/java/opendcs/src/main/java/decodes/launcher/LauncherFrame.java
@@ -1193,11 +1193,24 @@ Logger.instance().info("LauncherFrame ctor - getting dacq launcher actions...");
         try
         {
             // Load the DCS Tool Configuration
-            DecodesSettings settings = DecodesSettings.fromProfile(launchProfile);
+            final DecodesSettings settings = DecodesSettings.fromProfile(launchProfile);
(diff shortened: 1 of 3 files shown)


java/opendcs/src/main/java/decodes/launcher/LauncherFrame.java
```diff
@@ -1193,11 +1193,24 @@ private void doInitDecodes() throws DecodesException
...
+                catch (DecodesException ex)
+                {
+                    throw new RuntimeException(ex);
+                }
+            });
+
 
         }
-        catch (Exception ex)
+        catch (Throwable ex)
```
adamkorynta, line 1213: Should catch `RuntimException`, not `Throwable`. Unless there's a specific `Error` subclass you're targeting (scary), which should get a comment explainer

MikeNeilson: fair point, I was mostly being lazy but with `catch( Type 1 | Type 2 ex)` syntax there's not really any point in being that sloppy.

ktarbet (approved):

PR #1018, opened 2025-04-01
Refactor/1012 cwms oracle migration updates

### LINKED ISSUE #1012: Allow Database migration to be used with a non-sys DBA account
MigrationProviders, specially CWMS-Oracle

**Purpose of refactoring**
Migration assumes a large amount of "primiliary" work is done that can be handle if an account with the correct permissions are used.

Following the CWMS Database itself, this will either be done by the DBAs or in the case of an RDS database an "admin" account with the correct permissions is already used.

**Proposed method**
- Allow MigrationProviders to determine if schemas should be created
- Adapt tests to use new mechanism (sys should function in the same vain but may create a "lesser" admin user)
- Create a beforeMigration script to handle application of appropriate grants or other before schema migration details.

**Additional context**

MikeNeilson:
## Problem Description

Fixes #1012. 

## Solution

Transition to an "admin" (dba) account to create the schema instead of making the CCP user ahead of time.

Also discovered an order of operations problem within the manageDatabase application in regards to the configuration placeholder values.

## how you tested the change

The integration tests test this work.

## Where the following done:


java/opendcs/src/main/java/decodes/cwms/database/CwmsOracleProvider.java
```diff
@@ -45,23 +45,23 @@ public class CwmsOracleProvider implements MigrationProvider
...
         properties.add(
             new MigrationProperty(
                 "DEFAULT_OFFICE", String.class,
-                ""));
+                "Ascii value of the default office to assign"));
         properties.add(
             new MigrationProperty(
                 "TABLE_SPACE_SPEC", String.class,
-                ""));
+                "If data will be on a separate table space indicate the line here."));
```
ktarbet, line 64:
```suggestion
                "Name of table space, leave blank if you don't need a separate table space"));
```

ktarbet (commented):
looks good to me.  I've added one suggestion on  MigrationProperty description.

ktarbet (approved):

PR #1031, opened 2025-04-03
Bugfix/1019 cwms multi office

### LINKED ISSUE #1019: CWMS implementation does not correctly support multiple offices
**Describe the bug**
If data for multiple offices is used within the system a combination of the following will happen:
1. Data will only be saved as the default office configured at database install
2. Some tables do not include the office_code field and return results to queries when they shouldn't - some places in java handle this gracefully and some do not.
3. Subsubcription system only started for a single office at this time.

**To Reproduce**
Do things like dbimport with a profile with a different office.

**Expected behavior**
Data between offices stays independent

**Screenshots**

**Additional context**

MikeNeilson:
## Problem Description

Fixes #1019. 

## Solution

Add trigger to set office ID based on current session.

May need to update some tables to include the office_code field.

## how you tested the change

Existing tests should not break. Additional tests that insert data for a different office are needed.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/decodes/cwms/database/CwmsOracleProvider.java
+++ b/java/opendcs/src/main/java/decodes/cwms/database/CwmsOracleProvider.java
@@ -272,4 +272,14 @@ public class CwmsOracleProvider implements MigrationProvider
     {
         return true;
     }
+
+    @Override
+    public List<String> getAdminRoles()
+    {
+        ArrayList<String> roles = new ArrayList<>();
+        roles.add("CCP MGR");
+        roles.add("CCP PROC");
+        roles.add("CWMS Users");
+        return roles;
+    }
 }
--- a/java/opendcs/src/main/java/decodes/util/CmdLineArgs.java
+++ b/java/opendcs/src/main/java/decodes/util/CmdLineArgs.java
@@ -126,11 +126,10 @@ public class CmdLineArgs
             {
                 System.err.println("Invalid define '" + arg + "' -- should be "
                     + "in the form name=value.");
-                System.exit(1);
             }
             String name = arg.substring(0,idx).trim();
             String value = arg.substring(idx+1).trim();
-            if (name.length() == 0 || value.length() == 0)
+            if (name.length() == 0)
             {
                 System.err.println("Invalid define name='" + name
                     + "', value='" + value + "'");
--- a/java/opendcs/src/main/java/org/opendcs/database/ManageDatabaseApp.java
+++ b/java/opendcs/src/main/java/org/opendcs/database/ManageDatabaseApp.java
@@ -5,6 +5,7 @@ import decodes.util.CmdLineArgs;
 import decodes.util.DecodesSettings;
 import ilex.cmdline.StringToken;
 import ilex.cmdline.TokenOptions;
+import ilex.util.Pair;
 
 import javax.sql.DataSource;
 
@@ -17,6 +18,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Properties;
 
 /**
  * Simple Terminal application to prompt user for required information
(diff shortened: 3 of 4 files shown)


java/opendcs/src/main/java/decodes/cwms/database/CwmsOracleProvider.java
```diff
@@ -45,23 +45,23 @@ public class CwmsOracleProvider implements MigrationProvider
...
         properties.add(
             new MigrationProperty(
                 "DEFAULT_OFFICE", String.class,
-                ""));
+                "Ascii value of the default office to assign"));
         properties.add(
             new MigrationProperty(
                 "TABLE_SPACE_SPEC", String.class,
-                ""));
+                "If data will be on a separate table space indicate the line here."));
```
ktarbet, line 64:
```suggestion
                "Name of table space, leave blank if you don't need a separate table space"));
```

java/opendcs/src/main/resources/db/CWMS-Oracle/callbacks/beforeMigrate__set_schema_permissions.sql
```diff
@@ -0,0 +1,65 @@
...
+
+GRANT CREATE ANY CONTEXT TO ${CCP_SCHEMA};
+GRANT SELECT ON dba_scheduler_jobs to ${CCP_SCHEMA};
+GRANT SELECT ON dba_queue_subscribers to ${CCP_SCHEMA};
+GRANT SELECT ON dba_subscr_registrations to ${CCP_SCHEMA};
+GRANT SELECT ON dba_queues to ${CCP_SCHEMA};
+GRANT EXECUTE ON dbms_aq TO ${CCP_SCHEMA};
+GRANT EXECUTE ON dbms_aqadm TO ${CCP_SCHEMA};
+GRANT EXECUTE ON DBMS_SESSION to ${CCP_SCHEMA};
+GRANT EXECUTE ON DBMS_RLS to ${CCP_SCHEMA};
```
ktarbet, line 31: Other parts of the code have  suffix '_users'    , but why not on these grants?

I see one example below where both with and without the '_users' is used.

${CWMS_SCHEMA}.cwms_env

java/opendcs/src/main/resources/db/CWMS-Oracle/callbacks/afterMigrate__activateOfficeTrigger.sql
```diff
@@ -0,0 +1,18 @@
...
+
+  for rec in (
+    select table_name from user_tab_columns where column_name in ('DB_OFFICE_CODE')
+    )
+  loop
+    execute immediate 'alter table ' || rec.table_name || ' modify db_office_code default null';
+    the_sql := 
+    'create or replace trigger ' || rec.table_name || '_get_office' ||
+    ' before insert or update on ' || rec.table_name ||
+    ' for each row begin set_office_column(:new.db_office_code,:old.db_office_code); end;';
```
ktarbet, line 13: Will that work --  passing :new.db_office_code  by 'reference'  into set_office_column?  The :new.db_office_code feels like it is a value type.

ktarbet (commented):
Mostly some questions,  and one language change that didn't seem to copy from another PR.

java/opendcs/src/main/resources/db/CWMS-Oracle/callbacks/beforeMigrate__set_schema_permissions.sql
MikeNeilson: These grants are specific to the CCP schema "user". Their uses is through the CCP procedures which handle additional permissions checks.

the ${CCP_SCHEMA}_users is the name of the role granted to anyone who has an account and will use CCP.

java/opendcs/src/main/resources/db/CWMS-Oracle/callbacks/afterMigrate__activateOfficeTrigger.sql
MikeNeilson: `:new` and `:old` are triggerisms. They are provided by the engine that executes the trigger.

one or the other may be null, otherwise they will match the datatype of the insert itself. the set_office_column method then handles the various conditions; which due to how the triggers work I was not able to handle in the trigger directly. (There can be a lot of quirks in triggers.)

ktarbet (approved):

PR #1048, opened 2025-04-16
[draft]look at base class for annotations.   

ktarbet:
## Problem Description

Example:
DatchkScreeningAlgorithminherits from ScreeningAlgorithm,  but the annotation processor doesn't look at the base classes..

## Solution

look at the class hierarchy


MikeNeilson (commented):
Looks reasonable overall.

PR #1056, opened 2025-04-18
Wtp optimization

oskarhurst:
## Problem Description

Currently running multiple day of the ResEvap Algorithm is extremely slow.

## Solution

Describe your solution.

REsturcture ALgorithm base class to include new BeforeAll and AfterAll methods to prefore all data base calls only once, rather than per day. Also include check to only initializes resEvap and resevoir object once.

## how you tested the change

Resevap should still pass resevap algo tests. Profiles to data base are now constant, rather than O(n) to days.

## Where the following done:


MikeNeilson (commented):
This seems reasonable to me.

Would be could to get @agilmore2 thoughts. Looking at some of the HDB algorithms and how they use the database, this would likely help with some performance issues since we're moving away from sharing a single connection in all of the implementations.

ktarbet (approved):

PR #1074, opened 2025-04-26
add two more methods to DecodesScript builder

ktarbet:
This supported removing some UI code from ConfigEditPanel.


java/opendcs/src/main/java/decodes/db/DecodesScript.java
```diff
@@ -676,11 +676,14 @@ public static class DecodesScriptBuilder
         final private DecodesScriptReader scriptReader;
         private Supplier<PlatformConfig> platformSupplier;
         private Supplier<String> nameSupplier;
+        private String scriptType;
```
MikeNeilson, line 679: These should probably follow the same pattern as above an be a supplier.

MikeNeilson (dismissed):

MikeNeilson (approved):

MikeNeilson (approved):

PR #1090, opened 2025-05-05
Initial Commit of cp_computation indexing resolving full table scan

JeremyDKellett:
I noticed in Instance Viewer that cwms_ccp.notify_for_comp came up for having a lot of disk reads and buffer gets. On NAE's DB it's the #7 SQL with 14.3M disk reads and 677M buffer gets. On Omaha's DB it's #9 with 44.2M disk reads and 2.1G buffer gets.

The package calls:
```
ccp.cwms_cpp.notify_tsdatastored;
```

Which calls this SQL:

```
select distinct cc.loading_application_id
         from cp_comp_depends cd, cp_computation cc
         where cd.ts_id = :p_ts_code and cc.enabled = 'Y'
           and cc.loading_application_id is not null 
           and cd.computation_id = cc.computation_id;
```


java/opendcs/src/main/resources/db/CWMS-Oracle/schema/V7.1_index_cp_computation.sql
```diff
@@ -0,0 +1,12 @@
+------------------------------------------------------------------------------
+-- CWMS/CCP Performance Improvements 
+------------------------------------------------------------------------------
+
+-----------------------------------------------------------------------------
```
MikeNeilson, line 5: FYI, we're moving inline comments like this. Not a huge deal in the schema migrations but in the java files it just makes things messy.

java/opendcs/src/main/resources/db/CWMS-Oracle/schema/V7.1_index_cp_computation.sql
MikeNeilson, line 1: There needs to be two underscores after the V7.1 for the file to be picked up correctly by flyway.

e.g. `V7.1__index_cp_computations.sql`

java/opendcs/src/main/resources/db/CWMS-Oracle/schema/V7.1_index_cp_computation.sql
```diff
@@ -0,0 +1,12 @@
...
+-- 05MAY2025 - JDK - Initial Commit. CP_COMPUTATION indexing to remove table
+--                   scans and disk reads. Issue #1089.
+--                   https://github.com/opendcs/opendcs/issues/1089
+-----------------------------------------------------------------------------
+
+CREATE INDEX CP_COMPUTATION_IDX1 ON CP_COMPUTATION (COMPUTATION_ID ASC, ENABLED ASC, LOADING_APPLICATION_ID ASC, 'X' ASC) ;
```
MikeNeilson, line 11: What is the `'X' ASC` for?

MikeNeilson (changes requested):
Thanks for the PR. Considering how much this table is queried this should improve performance.

Only issues are formatting and style (that for SQL are admittedly *not* documented at all.)

java/opendcs/src/main/resources/db/CWMS-Oracle/schema/V7.1_index_cp_computation.sql
JeremyDKellett: Copy. Will remove.

java/opendcs/src/main/resources/db/CWMS-Oracle/schema/V7.1_index_cp_computation.sql
JeremyDKellett: It's neat. 

It's effectively makes the index a function based index (because it is indexing 'X' for each column) and thus the rows with null values for loading_application_id become indexed, which was the goal.

MikeNeilson: cool, please put *that* as a comment above this index. We have several other tables that may include nulls in primary keys and could use this.

MikeNeilson (approved):

PR #1092, opened 2025-05-06
More algorithm annotation work

tlelv:
## Problem Description

Updated outdated documentation, working to get rid of Algorithm Wizard.

## Solution

Describe your solution.

Swapped AW with annotations.

## how you tested the change

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/decodes/hdb/algo/EOPInterpAlg.java
+++ b/java/opendcs/src/main/java/decodes/hdb/algo/EOPInterpAlg.java
@@ -14,9 +14,6 @@ import decodes.tsdb.algo.AWAlgoType;
 // new class handles surrogate keys as an object
 import decodes.sql.DbKey;
 
-
-//AW:IMPORTS
-// Place an import statements you need here.
 import decodes.tsdb.ParmRef;
 import java.util.Calendar;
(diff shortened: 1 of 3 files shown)


java/opendcs/src/main/java/decodes/hdb/algo/EOPInterpAlg.java
```diff
@@ -30,22 +27,18 @@
...
+import org.opendcs.annotations.algorithm.Algorithm;
+import org.opendcs.annotations.algorithm.Input;
+import org.opendcs.annotations.algorithm.Output;
 
-//AW:JAVADOC
-/**
-Type a javadoc-style comment describing the algorithm class.
- */
-//AW:JAVADOC_END
+@Algorithm(description = "Type a javadoc-style comment describing the algorithm class.")
```
ktarbet, line 35: Looking at the algorithm I would suggest something like

'interpolate End-Of-Period value'

instead of the default javadoc template for the description.

ktarbet (changes requested):
Thanks for the progress.

tlelv:
Made the changes!

ktarbet (approved):

PR #1099, opened 2025-05-12
7.0 updates 2025 05 12

MikeNeilson:
Latest batch of fixes that should go into 7.0.15.

This should be the last batch. Everything else is on-going or will need a larger round of testing.


ktarbet (approved):
I manually tested the following:
* replaceValueWith function
* load from URL feature in Retrieve Sample Message dialog

PR #1111, opened 2025-05-16
Explore/rledit not working

### LINKED ISSUE #1110: rledit program won't start - java.lang.NullPointerException
**Describe the bug**
rledit (decodes.rledit.RefListEditor) is crashing on startup.

**To Reproduce**

 C:\project\opendcs\install\build\install\opendcs\bin>rledit
Exception in thread "main" java.lang.NullPointerException
        at decodes.rledit.EUTableModel.rebuild(EUTableModel.java:76)
        at decodes.rledit.EUTableModel.<init>(EUTableModel.java:67)
        at decodes.rledit.RefListFrame.<init>(RefListFrame.java:60)
        at decodes.rledit.RefListEditor.<init>(RefListEditor.java:31)
        at decodes.rledit.RefListEditor.main(RefListEditor.java:108)

**Expected behavior**
no crashing

ktarbet:
## Problem Description

Fixes #1110  rledit program is crashing

## Solution

restore functionality by connecting with legacy database using getLegacyDatabase(Database.class);

## how you tested the change

manually tested.

### DIFF
--- a/java/opendcs/src/main/java/decodes/rledit/EUTableModel.java
+++ b/java/opendcs/src/main/java/decodes/rledit/EUTableModel.java
@@ -52,11 +52,14 @@ public class EUTableModel extends AbstractTableModel
 	private ArrayList<EngineeringUnit> eus;
 	private int lastSortColumn;
 
+	Database db;
+
 	/**
 	 * No args constructor for JBuilder.
 	 */
-	public EUTableModel()
+	public EUTableModel(Database db)
 	{
+		this.db = db;
 		columnNames = new String[4];
 		columnNames[0] = labels.getString("EUTableModel.abbrev");
 		columnNames[1] = labels.getString("EUDialog.fullName");
@@ -73,7 +76,7 @@ public class EUTableModel extends AbstractTableModel
 	public void rebuild()
 	{
 		eus = new ArrayList<EngineeringUnit>();
-		EngineeringUnitList eul = Database.getDb().engineeringUnitList;
+		EngineeringUnitList eul = db.engineeringUnitList;
 		for(Iterator<EngineeringUnit> it = eul.iterator(); it.hasNext(); )
 		{
 			EngineeringUnit eu = (EngineeringUnit)it.next();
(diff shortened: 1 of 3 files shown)


java/opendcs/src/main/java/decodes/rledit/EUTableModel.java
```diff
@@ -52,11 +52,14 @@ public class EUTableModel extends AbstractTableModel
...
 	private int lastSortColumn;
 
+	Database db;
+
 	/**
 	 * No args constructor for JBuilder.
 	 */
 	public EUTableModel()
 	{
+		this.db = Database.getDb();
```
MikeNeilson, line 62: In the constructor would be better, but this is still a good improvement.

Also EngineeringUnitsList is a good candidate for migration to the new DAOs. While used just about everywhere it's contents are fairly constrained.

java/opendcs/src/main/java/decodes/rledit/RefListFrame.java
```diff
@@ -104,18 +103,39 @@ public class RefListFrame extends JFrame
...
      * constructor for RefListFrame
      */
-    public RefListFrame(OpenDcsDatabase database)
+    public RefListFrame(OpenDcsDatabase database) throws DecodesException
     {
         this.database = database;
+        log.info("RefListFrame:initDecodes()");
+		Optional<Database> db = this.database.getLegacyDatabase(Database.class);
+		if (db.isPresent()) 
+        {
```
MikeNeilson, line 120: This should only be done in a main function. not in individual classes, even the GUI ones.

MikeNeilson (changes requested):
Definitely an improvement.

The initializeForDecoding (though really should be ForEditing here) should happen in the RefListEditor main. 

No class that uses OpenDcsDatabase should need to know about initialization.

java/opendcs/src/main/java/decodes/rledit/EUTableModel.java
MikeNeilson: you would have to create the DAO and change all the usages. i wasn't suggesting to do it for this PR.

MikeNeilson (approved):

PR #1125, opened 2025-05-26
Feature/1112 starttls implementation

### LINKED ISSUE #1112: Implement STARTTLS support
https://github.com/opendcs/opendcs/pull/228 implements TLS directly.
This can make it difficult for servers to deal with various upgrade paths.

Implement STARTTLS on DDS Connections

Do nothing. Maybe we don't need STARTTLS and it should be all or nothing?

**Additional context**
Given we want to make a public LRGS, but many clients won't have upgraded yet, this does seem like a good idea.

Alternatively we should support connections on multiple ports like how http and https use 80 and 443 respectively.

MikeNeilson:
## Problem Description

Fixes #1112. 

## Solution

Instead of the current all or nothing approach move TLS setting to an Enum that supports:
1. NONE - No TLS
2. START_TLS - allow clients to begin encryption (DDS Command)
3. TLS - TLS is required for the initial connection to the socket.

TODO:

- [x] LrgsConnection and file
- [x] LrgsConnectionPanel
- [x] LrgsConfigDailog
- [x] Routing Spec
- [x] Lrgs DdsRecv Connections.

## how you tested the change

Automated tests were created to verify basic operations.
GUI elements will be manually tested.

## Where the following done:


MikeNeilson:
Still need to figure out why the windows test isn't passing. Likely something odd with how the windows file names in for temp files are handled.

@ktarbet didn't you recently fix and issue like this for the Xml Database Provider?

integrationtesting/lrgs/src/test/java/org/opendcs/fixtures/lrgs/LrgsTestInstance.java
```diff
@@ -44,6 +52,14 @@ public LrgsTestInstance(File lrgsHome) throws Exception
             fw.write("hritSourceCode=HR"+System.lineSeparator());
             fw.write("hritFileEnabled=true"+System.lineSeparator());
             fw.write("noTimeout=true"+System.lineSeparator());
+            fw.write("ddsListenPort=0"+System.lineSeparator());
+            fw.write("enableDdsRecv=true"+System.lineSeparator());
+            fw.write("ddsServerTlsMode="+tlsMode.name()+System.lineSeparator());
+
+            if (keyStore!=null) {
+                fw.write("keyStoreFile="+keyStore.getAbsolutePath()+System.lineSeparator());
```
ktarbet, line 60:
```suggestion
                String fileName =keyStore.getAbsolutePath();
                fileName = fileName.replace('\\','/');
                fw.write("keyStoreFile="+fileName+System.lineSeparator());
```

ktarbet (commented):
I've added a suggestion that gets the tests passing on my Windows computer.
I changed '\' to '/'

ktarbet (approved):

PR #1131, opened 2025-05-30
update command line PasswordFileEditor to work on windows  (for LRGS passwords)

ktarbet:
## Problem Description

note: PR into   #1112 

Add support for Windows.  Windows users were not able to run the editPassword program.

## Solution

Use java 1.8 feature of System.console().readPassword()  when using a terminal,  Remove legacy TTYEcho.java.

## how you tested the change

Manual testing on windows, and with batch file (see start-lrgs.rst addition)


java/opendcs/src/main/java/ilex/util/PasswordFileEditor.java
```diff
@@ -396,26 +402,34 @@ protected boolean isOkToQuit( )
 	* @param echo
 	* @return the response
 	*/
-	private String getResponse( String prompt, boolean echo )
+private String getResponse(String prompt, boolean echo)
```
MikeNeilson, line 405: We should just move this to a static utility function. It's identical in the two files.

MikeNeilson (approved):

PR #1135, opened 2025-06-01
Add database_properties file that DatabaseService now expects.

MikeNeilson:
## Problem Description

Fixes #. 

We setup the databaseFromDataSource to expect a specific table, that was not previously available

## Solution

Create a table (in OpenDCS-Postgres/Oracle and CWMS-Oracle)

## how you tested the change

Letting the integration tests run. This is currently only used by the rest_api

## Where the following done:


ktarbet:
Is there documentation for the config defined in database? (that is what this looks like) 


I'm seeing this error in a few places in the failed builds.
```txt
END;, Error Message = ORA-65040: Operation is not allowed from within a pluggable database.

```

ktarbet (approved):
Looks ok to me.
Thanks for VPD update.

PR #1157, opened 2025-06-09
Update TraceDialog.java

oskarhurst:
## Problem Description

When Trace Dialog reaches the maximum number of messages it becomes full and does not show any new messages.

## Solution

Describe your solution.

Trace Dialog now has two new buttons the clear and autoCycle buttons and shows an error when the dialog has reached its max.
AutoCycle automatically when the dialog fills up removed the top half of message to allow for more space.

## how you tested the change

Describe what was done to test the change. 

Manually tested fetured in the log heavy resEvap Comp.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/decodes/dbeditor/TraceDialog.java
+++ b/java/opendcs/src/main/java/decodes/dbeditor/TraceDialog.java
@@ -2,7 +2,7 @@ package decodes.dbeditor;
 
 import java.awt.*;
 import javax.swing.*;
-import javax.swing.text.DefaultCaret;
+import javax.swing.text.*;
 
 import java.awt.event.*;
 import java.util.ResourceBundle;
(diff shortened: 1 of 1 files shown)


java/opendcs/src/main/java/decodes/dbeditor/TraceDialog.java
```diff
@@ -131,17 +148,36 @@ public void addText(String text)
...
-		if (numMessages < maxMessages)
+		if (numMessages <= maxMessages || autoCycle.isSelected())
 		{
 			SwingUtilities.invokeLater(() ->
 			{
-				eventArea.append(text);
+
+				if (numMessages == maxMessages && autoCycle.isSelected())
+				{
+					errorLabel.setText("");
```
MikeNeilson, line 158: well, that's one way to reduce the memory allocations involved.

java/opendcs/src/main/java/decodes/dbeditor/TraceDialog.java
```diff
@@ -88,6 +91,14 @@ public void actionPerformed(ActionEvent e) {
 				closeButton_actionPerformed(e);
 			}
 		});
+		clearButton.setText(
+			genericLabels.getString("clear"));
+		clearButton.addActionListener(new java.awt.event.ActionListener() {
+			public void actionPerformed(ActionEvent e) {
+				clear();
+			}
+		});
```
MikeNeilson, line 100: Use the shorter lambda syntax here.
```suggestion
		clearButton.addActionListener(e -> clear());
```

MikeNeilson (changes requested):
Otherwise looks reasonable.

Realized we need to change the toggle buttons to use a resource string so we have a translation but we can do that in a separate PR.

MikeNeilson (approved):

PR #1158, opened 2025-06-10
Removing AW_AlgorithmTemplate.java

tlelv:
## Problem Description

Removing AW, and starting with the AW_AlgorithmTemplate.java. From my tests, things run like normal without this file present. 

## Solution

Describe your solution.

## how you tested the change

## Where the following done:


MikeNeilson:
I'm not sure we should remove this file yet. It's used by the AlgoEdit program at the moment and we haven't determine a path forward on that.

If we do want to go whole-hog into having people say, use VS Code, we need to document that and determine if we're going to make an extension to help or not.

PR #1200, opened 2025-06-26
Update sorting and model behavior. Correct filter of algorithm and pr…

### LINKED ISSUE #994: Sorting on compedit Computation IDs should be numeric
**Describe the bug**
When the ID column in the comp edit GUI is clicked, the IDs are sorted by lexical order. These are numbers and 99 should be less than 100.

**Screenshots**

Image

 - OS: Fedora 40
 - Version 7.0.15-RC03

MikeNeilson:
…ocess.

## Problem Description

Fixes #994. 

## Solution

- Return raw data types from getValueAt call instead of the forced String.
- Moved TableModel to own class
- Use DbComputation instances directly
- Delete ComputationInListComparator
- Added missing sort constraint for GUI elements that were present.
- Move refresh operation off of GUI thread.

## how you tested the change

Manually. Open CompEdit and verified correctly computations were open in the edit panels and that some 
some of the filters worked correctly.

Will need to test the group filtering as well, it appears that logic, while it uses the CompFilter class is within the ComputationDAO::compEditList method instead of nested within the CompFilter itself.

Filtering by group still works, it's just applied at the tablesorter level now.

## Where the following done:


ktarbet (approved):

ktarbet (commented):
I tested compexport,  and manually tested Computations Tab of Computation Editor.


I don't understand the following behavior in a test:

Type 'Flow' in the filter  :  "Has Param Code: "
    Result: 96/210 
Type 'FlowKarl' 
    Result: 210/210 

I was expecting 0/210.

MikeNeilson:
> I tested compexport, and manually tested Computations Tab of Computation Editor.
> 
> I don't understand the following behavior in a test:
> 
> Type 'Flow' in the filter : "Has Param Code: " Result: 96/210 Type 'FlowKarl' Result: 210/210
> 
> I was expecting 0/210.

Honestly, I'm a little confused myself on that one. But the behavior hasn't changed so it's a separate issue.

ktarbet (approved):

PR #1209, opened 2025-07-02
7.5 updates 2025 07 02

MikeNeilson:
Bring 7.5 in line with main.


ktarbet:
Could we add a 7.5 tag to the main branch and get the same result?  I'm hoping for a eaiser way to review such a large PR.

MikeNeilson:
Considering they idea is to just keep them inline with each other I'm not opposed.

@adamkorynta, thoughts?

MikeNeilson:
> The proposal would be to start tagging releases from main and then create hotfix or long-lived release branches as needed when patching previous versions? I don't see an issue there.
> 
> Given the concern with schema versioning - if that gets out of hand, we could consider moving the database schema(s) into a separate repository with a separate versioning cadence/scheme. That would also allow for running integration tests against multiple versions of the schema given the java codebase supports different schema versions.

Interesting idea. Hadn't though of that, testing itself might be a smidge awkward, not really a good way to have a test without the existing integration tests, but certainly not impossible to deal with.

MikeNeilson:
Alright, I'm good with the current suggestion and proposals and will go attempt to make a 7.5.0 RC from the main branch and see what happens.

PR #1214, opened 2025-07-03
Initial attempt at Weighted Average algorithm

agilmore2:
## Problem Description

See https://github.com/opendcs/opendcs/discussions/1210
Weighted average with a max of 8 components.

## Solution

Basic first algorithm

## how you tested the change

Manually tested and working.

## Where the following done:

### DIFF
--- a/integrationtesting/opendcs-tests/src/test/resources/data/Comps/WeightedAverage/Test1/WeightedAverageTest1.config
+++ b/integrationtesting/opendcs-tests/src/test/resources/data/Comps/WeightedAverage/Test1/WeightedAverageTest1.config
@@ -1,2 +1,0 @@
-EnableOn: OpenDCS-Postgres
-EnableOn: CWMS-Oracle
--- a/java/opendcs/src/main/java/decodes/tsdb/algo/WeightedAverage.java
+++ b/java/opendcs/src/main/java/decodes/tsdb/algo/WeightedAverage.java
@@ -67,7 +67,7 @@ public class WeightedAverage extends AW_AlgorithmBase
     public double weight8;
 
     @Output(type = Double.class)
-    public NamedVariable output = new NamedVariable("output", 0);
+    public NamedVariable output = new NamedVariable("output", 0.0d);
 
     // Allow javac to generate a no-args constructor.
 
@@ -105,14 +105,13 @@ public class WeightedAverage extends AW_AlgorithmBase
      * @throws DbCompException (or subclass thereof) if execution of this
      *        algorithm is to be aborted.
      */
-    @SuppressWarnings("checkstyle:LeftCurly")
     @Override
     protected void doAWTimeSlice()
             throws DbCompException
     {
         if (isMissing(weightTotal))
         {
-            log.debug("Skipping time slice with base time {} because of missing value for weightTotal", debugSdf.format(_timeSliceBaseTime));
+            log.debug("Skipping time slice with base time {} because of missing value for weightTotal", _timeSliceBaseTime);
             return;
         }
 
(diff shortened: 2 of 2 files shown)


java/opendcs/src/main/java/decodes/tsdb/algo/WeightedAverage.java
```diff
@@ -0,0 +1,110 @@
...
+	@Input
+	public double input5;
+	@Input
+	public double input6;
+	@Input
+	public double input7;
+	@Input
+	public double input8;
+	@Input
+	public double weight1;
```
MikeNeilson, line 32: Are these time series?

java/opendcs/src/main/java/decodes/tsdb/algo/WeightedAverage.java
```diff
@@ -0,0 +1,110 @@
...
+
+import decodes.tsdb.DbCompException;
+import ilex.var.NamedVariable;
+import org.opendcs.annotations.algorithm.Algorithm;
+import org.opendcs.annotations.algorithm.Input;
+import org.opendcs.annotations.algorithm.Output;
+
+@Algorithm(description = "Calculate Weighted Average values (eg Water Temperature) using: sum(input1n*input2n) / total2")
+public class WeightedAverage
+	extends AW_AlgorithmBase
```
MikeNeilson, line 11: put the extends ... on the same line as the class

java/opendcs/src/main/java/decodes/tsdb/algo/WeightedAverage.java
```diff
@@ -0,0 +1,110 @@
...
+{
+	@Input
+	public double weightTotal;
```
MikeNeilson, line 14: it seems like it would be better to calculate this instead of have it, but that may depend on a question below as to whether the weights are time series or not.

java/opendcs/src/main/java/decodes/tsdb/algo/WeightedAverage.java
```diff
@@ -0,0 +1,110 @@
...
+	 * @throws DbCompException (or subclass thereof) if execution of this
+	 *        algorithm is to be aborted.
+	 */
+	@Override
+	protected void doAWTimeSlice()
+		throws DbCompException
+	{
+		if(weightTotal > 0)
+		{
+			double tot = input1*weight1;
```
MikeNeilson, line 88: Look at https://github.com/opendcs/opendcs/blob/main/java/opendcs/src/main/java/decodes/tsdb/algo/ScalerAdder.java, specifically https://github.com/opendcs/opendcs/blob/c180e39ea32703b7e7db8118787ebd31e234d515/java/opendcs/src/main/java/decodes/tsdb/algo/ScalerAdder.java#L96-L100 for an example of "ignore this input if not defined."

MikeNeilson (commented):
I forgot about this one: https://github.com/opendcs/opendcs/blob/main/java/opendcs/src/main/java/decodes/tsdb/algo/GroupAdder.java

Though I'm not actually sure if it's been behaving as expected though I think there's at least one USACE user of it.

java/opendcs/src/main/java/decodes/tsdb/algo/WeightedAverage.java
agilmore2: Yes, this is conceptual, and these will need to be altered to timeseries references. Not sure what the original algorithm is doing here.

java/opendcs/src/main/java/decodes/tsdb/algo/WeightedAverage.java
agilmore2: Thanks for the reference!

agilmore2:
> I forgot about this one: https://github.com/opendcs/opendcs/blob/main/java/opendcs/src/main/java/decodes/tsdb/algo/GroupAdder.java
> 
> Though I'm not actually sure if it's been behaving as expected though I think there's at least one USACE user of it.

So GroupAdder seems to do a lot of the logic that the parent classes do for most other algorithms. Do you think this is the right way to handle these algorithms? I ask because, well... https://github.com/agilmore2/opendcs/tree/CULAlgs has a few *specialty* algorithms in it that I eventually want to merge...

Might be good to get some standardized supporting code in a utility class or something!

java/opendcs/src/main/java/decodes/tsdb/algo/ScalerAdder.java
```diff
@@ -109,7 +109,9 @@ protected void doAWTimeSlice()
 			pr.missingAction != MissingAction.IGNORE)
 		 || ((pr = getParmRef(t = "input8")) != null && isAssigned(t) && isMissing(input8) && 
 			pr.missingAction != MissingAction.IGNORE)
-		 || ((pr = getParmRef(t = "input0")) != null && isAssigned(t) && isMissing(input9) && 
+		 || ((pr = getParmRef(t = "input9")) != null && isAssigned(t) && isMissing(input9) &&
```
MikeNeilson, line 112: good catch, I'm going to yank this out into a separate PR though, scalaradder30 was copy paste and problem also has this mistake and a separate commit will be easier to port to the 7.0 branch.

java/opendcs/src/main/java/decodes/tsdb/algo/WeightedAverage.java
agilmore2: This is definitely a timeseries that the user supplies. It does behave poorly if the total doesn't match the sum of the weights though!

java/opendcs/src/main/java/decodes/tsdb/algo/WeightedAverage.java
agilmore2: I've tested this and it's functioning. I'm going to add a bit more logging, and finalize the PR for review.

MikeNeilson:
take a look at the algorithm integration system, @oskarhurst can help if you have questions.

nothing in the algo looks implementation specific so shouldn't be unreasonable to setup the tests.

agilmore2:
Not sure what the "algorithm integration" system is? Sounds useful...
I'm ready for this to be considered for merge, please let me know what else is needed.

agilmore2:
Oh good! I am most of the way to having all those files created, although
not sure how to run the tests locally yet.

Thanks!

MikeNeilson:
> Oh good! I am most of the way to having all those files created, although not sure how to run the tests locally yet. Thanks!
> […](#)
> Message ID: ***@***.***>

Assuming you're on a host with docker, one of:

```bash

./gradlew :testing:opendcs-tests:tests --info
./gradlew :testing:opendcs-tests:testOpenDCSPostgres --info # note that might be OpenDcsPostgres

```

I haven't gotten around to setting up a "bypass" for the OpenDcs Implementations yet.

integrationtesting/opendcs-tests/src/test/resources/data/Comps/WeightedAverage/Test1/WeightedAverageTest1.config
```diff
@@ -0,0 +1,2 @@
+EnableOn: OpenDCS-Postgres
+EnableOn: CWMS-Oracle
```
MikeNeilson: you can delete this to run on all tsdb impls (OpenDCS-Oracle is also a thing)

java/opendcs/src/main/java/decodes/tsdb/algo/WeightedAverage.java
```diff
@@ -0,0 +1,202 @@
...
+              pr.missingAction != MissingAction.IGNORE)
+         || ((pr = getParmRef(t = "weight8")) != null && isAssigned(t) && isMissing(weight8) &&
+              pr.missingAction != MissingAction.IGNORE))
+        {
+            log.debug("Skipping time slice with base time {} because of missing value for param {}",
+                    debugSdf.format(_timeSliceBaseTime), t);
+            return;
+        }
+        if (!isMissing(input1) && !isMissing(weight1))
+            tot += (input1 * weight1);
```
MikeNeilson: Brackets. yeah i know it's an extra two lines and characters. I've been bit by this syntax too many times

java/opendcs/src/main/java/decodes/tsdb/algo/WeightedAverage.java
```diff
@@ -0,0 +1,202 @@
...
+            tot += (input7 * weight7);
+        if (!isMissing(input8) && !isMissing(weight8))
+            tot += (input8 * weight8);
+
+        // Output only if nonzero total weight
+        if (weightTotal != 0)
+        {
+            tot /= weightTotal;
+            log.trace("doAWTimeSlice baseTime={}, input1={}, weight1={}, input2={}, weight2={}, tot={}",
+                    debugSdf.format(_timeSliceBaseTime), input1, weight1, input2, weight2, tot);
```
MikeNeilson:
```suggestion
                    _timeSliceBaseTime, input1, weight1, input2, weight2, tot);
```

the long knows how to handle dates, we just left all the debugsdf objects in because ... so many and there were a few non logging usages.

MikeNeilson (approved):

PR #1215, opened 2025-07-03
column names are needed for saving properties in site_property

ktarbet:
SQL Inserts into the ccp.site_property table need to consider the new numeric db_office_code column.   

This PR is an experiment to dynamically lookup the column names for property tables.


ktarbet:
All the property tables, in CCP, have consistent names:   (Prop_Name, Prop_Value), so dynamic columns should not be necessary.

PR #1258, opened 2025-07-28
Bugfix/1256 logging fixes

### LINKED ISSUE #1256: CompProc partial operations and failed loging.
**Describe the bug**
CompProc appears to be somewhat working (if very slow) but no log entries until restart after a point.

**To Reproduce**
1. run comppproc with a lot of data
2. wait
3. see various missing data, some saved, no log entries after a point.

**Expected behavior**
Continued logging and data processing.

**Screenshots**

 - Version [main, 7.0.15]

**Additional context**

MikeNeilson:
## Problem Description

Possibly Fixes #1256. 

## Solution

Original system was logging one entry at a time using `Queue::poll`
Before changes I printed the entries in queue vs queue size and artificially shrunk the queue. There did appear to be the possibility of things hanging on adding new log entries to the queue which could have gotten things stuck. Additionally it may have been that the log was rotated but never picked up correctly hence the lack of updated log messages.

This change makes the log rotation a tighter part of the writing loop and should handle that situation better.

## how you tested the change

Unit test was updated to verify behavior of rotation.
Manually tests with compproc instance to see if responses were reasonable.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/ilex/util/FileLogger.java
+++ b/java/opendcs/src/main/java/ilex/util/FileLogger.java
@@ -3,16 +3,19 @@
 */
 package ilex.util;
 
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.PrintStream;
 import java.util.ArrayList;
-import java.util.Date;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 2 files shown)


java/opendcs/src/main/java/ilex/util/FileLogger.java
```diff
@@ -26,6 +31,17 @@
...
 public class FileLogger extends Logger
 {
+	/**
+	 * Message Queue capacity.
+	 *
+	 * This number is currently rather arbitrary and may need adjustment in the future.
+	 */
+	private static final int MESSAGE_CAPACITY = 5000;
+	/**
+	 * Numbered of queued messages to log in a single batch
```
ktarbet, line 41:
```suggestion
	 * Number of queued messages to log in a single batch
```

ktarbet (approved):

PR #1271, opened 2025-08-05
Create LoggerFactory helper, demo usage, and update DaoBase for example.

### LINKED ISSUE #1267: LoggerFactory helper
Using the Slf4j LoggerFactory requires passing in the surrounding class which is prone to error.

Create a wrapper factory that determines the calling class.

Wait for SLF4j to implement such. However, they appear rather concerned, and reasonably so, about edge cases we won't see within our usage.

**Additional context**
We're about to go whole hog on updating the logging as the recent "multi-threading" modification to the `ilex.util.FileLogger` is not behaving correctly and rather than spend time analzying it @ktarbet and I decided to just suck it up and update all the logging usage to slf4j.

MikeNeilson:
## Problem Description

Fixes #1267. 

## Solution

Implement the request helper.

## how you tested the change

Unit test for basic behavior.
Ran integration tests and saw DaoBase message in the log as expected.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/opendcs/dao/DaoBase.java
+++ b/java/opendcs/src/main/java/opendcs/dao/DaoBase.java
@@ -23,7 +23,6 @@
  */
 package opendcs.dao;
 
-import ilex.util.Logger;
 import opendcs.dai.DaiBase;
 import opendcs.util.functional.ConnectionConsumer;
 import opendcs.util.functional.DaoConsumer;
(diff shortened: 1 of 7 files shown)


java/opendcs/src/main/java/opendcs/dao/DaoBase.java
```diff
@@ -235,19 +236,18 @@ public ResultSet doQuery(String q)
...
         {
-            String msg = "SQL Error in query '" + q + "': " + ex;
-            warning(msg);
-            throw new DbIoException(msg);
+            String msg = "SQL Error in query '" + q + "'";
+            if (log.isTraceEnabled())
+            {
+                log.atWarn().setCause(ex).log(msg);
+            }
+            throw new DbIoException(msg, ex);
```
ktarbet, line 250:
```suggestion
            log.atWarn().setCause(ex).log(msg);
            
            throw new DbIoException(msg, ex);
```

When would we need  if (log.isTraceEnabled()) ?   It seems simpler to just have the log.atWarn() do the filtering?

ktarbet (approved):

PR #1300, opened 2025-08-14
Create CwmsLocationLevelDAO.java

oskarhurst:
## Problem Description
Add Location levelDAO

Fixes #. 

## Solution

Describe your solution.

## how you tested the change

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/decodes/platstat/DbPollThread.java
+++ b/java/opendcs/src/main/java/decodes/platstat/DbPollThread.java
@@ -1,23 +1,23 @@
-/**
- * $Id: DbPollThread.java,v 1.2 2016/08/05 14:50:18 mmaloney Exp $
- * 
- * Open Source Software
- * 
- * $Log: DbPollThread.java,v $
- * Revision 1.2  2016/08/05 14:50:18  mmaloney
- * Station and Routing Status GUI updates.
- *
- * Revision 1.1  2016/07/20 15:40:12  mmaloney
- * First platstat impl GUI.
- *
- *
- */
+/*
+* Where Applicable, Copyright 2025 OpenDCS Consortium and/or its contributors
+* 
+* Licensed under the Apache License, Version 2.0 (the "License"); you may not
+* use this file except in compliance with the License. You may obtain a copy
+* of the License at
+* 
+*   http://www.apache.org/licenses/LICENSE-2.0
+* 
+* Unless required by applicable law or agreed to in writing, software 
+* distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+* WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+* License for the specific language governing permissions and limitations 
+* under the License.
+*/
 package decodes.platstat;
 
 import java.util.ArrayList;
 import java.util.HashMap;
 
-import ilex.util.Logger;
 import opendcs.dai.PlatformStatusDAI;
 import opendcs.dai.ScheduleEntryDAI;
 import decodes.db.PlatformStatus;
@@ -26,13 +26,16 @@ import decodes.db.ScheduleEntryStatus;
 import decodes.sql.DbKey;
 import decodes.tsdb.DbIoException;
 
+import org.opendcs.utils.logging.OpenDcsLoggerFactory;
+import org.slf4j.Logger;
+
 /**
  * Polls Time Series Database for process status and updates model.
  * @author mmaloney Mike Maloney, Cove Software, LLC
  */
-public class DbPollThread 
-	extends Thread
+public class DbPollThread extends Thread
 {
+	private static final Logger log = OpenDcsLoggerFactory.getLogger();
 	private PlatformMonitor platformMonitor = null;
 	private boolean _shutdown = false;
 	public static final long updateInterval = 5000L;
@@ -80,7 +83,7 @@ public class DbPollThread
 				}
 				catch(DbIoException ex)
 				{
-					Logger.instance().warning("Error reading platstat info: " + ex);
+					log.atWarn().setCause(ex).log("Error reading platstat info.");
 				}
 				finally
 				{
--- a/java/opendcs/src/main/java/decodes/platstat/PlatformMonitor.java
+++ b/java/opendcs/src/main/java/decodes/platstat/PlatformMonitor.java
@@ -1,3 +1,18 @@
+/*
+* Where Applicable, Copyright 2025 OpenDCS Consortium and/or its contributors
+* 
+* Licensed under the Apache License, Version 2.0 (the "License"); you may not
+* use this file except in compliance with the License. You may obtain a copy
+* of the License at
+* 
+*   http://www.apache.org/licenses/LICENSE-2.0
+* 
+* Unless required by applicable law or agreed to in writing, software 
+* distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+* WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+* License for the specific language governing permissions and limitations 
+* under the License.
+*/
 package decodes.platstat;
 
 import java.awt.event.WindowAdapter;
@@ -6,13 +21,16 @@ import java.awt.event.WindowEvent;
 import javax.swing.ImageIcon;
 
 import ilex.util.EnvExpander;
-import ilex.util.Logger;
 import lrgs.gui.DecodesInterface;
 import decodes.tsdb.TimeSeriesDb;
 import decodes.tsdb.TsdbAppTemplate;
 
+import org.opendcs.utils.logging.OpenDcsLoggerFactory;
+import org.slf4j.Logger;
+
 public class PlatformMonitor extends TsdbAppTemplate
 {
+	private static final Logger log = OpenDcsLoggerFactory.getLogger();
 	public static final String module = "PlatformMonitor";
 	private boolean exitOnClose = true;
 	private PlatformMonitorFrame frame = null;
@@ -75,10 +93,7 @@ public class PlatformMonitor extends TsdbAppTemplate
 		} 
 		catch (Exception ex)
 		{
-			String msg = module + " Can not initialize. " + ex.getMessage();
-			Logger.instance().failure(msg);
-			System.err.println(msg);
-			ex.printStackTrace(System.err);
+			log.atError().setCause(ex).log("Can not initialize.");
 		}
 	}
 
--- a/java/opendcs/src/main/java/decodes/platstat/PlatformMonitorFrame.java
+++ b/java/opendcs/src/main/java/decodes/platstat/PlatformMonitorFrame.java
@@ -1,19 +1,22 @@
-/**
- * $Id: PlatformMonitorFrame.java,v 1.2 2016/08/05 14:50:18 mmaloney Exp $
- *
- * $Log: PlatformMonitorFrame.java,v $
- * Revision 1.2  2016/08/05 14:50:18  mmaloney
- * Station and Routing Status GUI updates.
- *
- * Revision 1.1  2016/07/20 15:40:12  mmaloney
- * First platstat impl GUI.
- *
- */
+/*
+* Where Applicable, Copyright 2025 OpenDCS Consortium and/or its contributors
+* 
+* Licensed under the Apache License, Version 2.0 (the "License"); you may not
+* use this file except in compliance with the License. You may obtain a copy
+* of the License at
+* 
+*   http://www.apache.org/licenses/LICENSE-2.0
+* 
+* Unless required by applicable law or agreed to in writing, software 
+* distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+* WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+* License for the specific language governing permissions and limitations 
+* under the License.
+*/
 package decodes.platstat;
 
 import ilex.gui.EventsPanel;
 import ilex.util.LoadResourceBundle;
-import ilex.util.Logger;
 
 import java.awt.BorderLayout;
 import java.awt.Dimension;
@@ -39,6 +42,9 @@ import javax.swing.event.ListSelectionEvent;
 import javax.swing.event.ListSelectionListener;
 
 import opendcs.dai.DacqEventDAI;
+import org.opendcs.utils.logging.OpenDcsLoggerFactory;
+import org.slf4j.Logger;
+
 import decodes.db.Database;
 import decodes.db.DatabaseIO;
 import decodes.db.PlatformStatus;
@@ -56,9 +62,9 @@ import decodes.util.DecodesSettings;
  * @author mmaloney Mike Maloney, Cove Software, LLC
  */
 @SuppressWarnings("serial")
-public class PlatformMonitorFrame 
-	extends TopFrame 
+public class PlatformMonitorFrame extends TopFrame 
 {
+	private static final Logger log = OpenDcsLoggerFactory.getLogger();
 	private PlatformTableModel platstatModel = null;
 	private EventsPanel eventsPanel = new EventsPanel();
 	private JSplitPane outerPane = null;
@@ -225,7 +231,7 @@ public class PlatformMonitorFrame
 			}
 			catch (DbIoException ex)
 			{
-				Logger.instance().warning("Error reading events: " + ex);
+				log.atWarn().setCause(ex).log("Error reading events.");
 			}
 			finally
 			{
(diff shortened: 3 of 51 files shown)


MikeNeilson (changes requested):

integrationtesting/fixtures/src/main/java/org/opendcs/fixtures/configurations/cwms/CwmsOracleConfiguration.java
```diff
@@ -149,6 +153,33 @@ public void start(SystemExit exit, EnvironmentVariables environment, SystemPrope
         }
     }
 
+    /**
+     * Load location level test data for integration tests
+     */
+    private void loadLocationLevelTestData()
```
MikeNeilson, line 159: This is better done in the test itself or at least the test class.

This section is for baseline data, which is only the minimum required to even start the applications and have them behave correctly.

integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CwmsLocationLevelDAOTestIT.java
```diff
@@ -0,0 +1,286 @@
...
+        {
+            // For non-CWMS databases, skip these tests
+            dao = null;
+        }
+    }
+    
+    @Test
+    @EnableIfDaoSupported(CwmsLocationLevelDAO.class)
+    public void testGetLatestLocationLevelValue_Basic() throws Exception
+    {
```
MikeNeilson, line 56: There is an enableif annotation for specific implementations that works at the class and method level.

integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CwmsLocationLevelDAOTestIT.java
```diff
@@ -0,0 +1,286 @@
...
+    private OpenDcsDatabase db;
+    
+    private CwmsLocationLevelDAO dao;
+    
+    @BeforeEach
+    public void setUp() throws Exception
+    {
+        if (tsDb instanceof CwmsTimeSeriesDb)
+        {
+            dao = new CwmsLocationLevelDAO((CwmsTimeSeriesDb) tsDb);
```
MikeNeilson, line 44: This PR is to main, use the new style of stateless DAO and the `OpenDcsdatabase::getDao` method.

integrationtesting/opendcs-tests/src/test/resources/data/cwms/location_level_test_data.sql
MikeNeilson, line 1: everything appears commented out why is this even here?

java/opendcs/src/main/java/opendcs/dao/LocationLevelDAO.java
MikeNeilson, line 1: You can ditch this, the `OpenDcsDatabase::getDao` just checks for anything that implements the `OpenDcsDao` (or whatever we called it) and returns back the desired type using generics.

java/opendcs/src/main/java/decodes/cwms/CwmsLocationLevelDAO.java
MikeNeilson, line 1: Make this just the stateless variant. 

The old style only needs to exist in the 7.0 branch.

java/opendcs/src/main/java/opendcs/dai/LocationLevelDAI.java
```diff
@@ -0,0 +1,89 @@
...
+    
+    /**
+     * Get the latest location level value with unit conversion
+     * @param locationLevelId The location level identifier
+     * @param targetUnits The desired units for the value (e.g., "ft", "m", "in")
+     * @param sourceUnits The source units to filter by (if null, no filter applied)
+     * @return The latest LocationLevelValue or null if none found
+     * @throws DbIoException on database error
+     */
+    LocationLevelValue getLatestLocationLevelValue(String locationLevelId,
```
MikeNeilson, line 36: Why do the store units need to be known here? We should just let user request in the units desired and throw an appropriate exception if the units don't apply.

java/opendcs/src/main/java/opendcs/dai/LocationLevelDAI.java
```diff
@@ -0,0 +1,89 @@
+package opendcs.dai;
+
+import decodes.tsdb.DbIoException;
+import opendcs.dao.LocationLevelDAO.LocationLevelValue;
```
MikeNeilson, line 4: You've now created a circular dependency of an interface on it's implementation, this is always a terrible idea.

Move the LocationLevelValue and Spec to a separate package space dedicated to just being data.

MikeNeilson (changes requested):
Don't mix database styles.

main gets the stateless, 7.0 gets the old style.
Yes that means it's going to get more difficult to port things and eventually there will be things that we just don't.

integrationtesting/fixtures/src/main/resources/database/cwms_location_level_test_data.sql
MikeNeilson, line 1: This file should be moved to the opencs-tests test resource directory.

integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CwmsLocationLevelDAOTestIT.java
```diff
@@ -0,0 +1,361 @@
...
+            // Test passes whether data exists or not
+            // If value is not null, verify its structure
+            if (value != null)
+            {
+                assertNotNull(value.getLevelDate(), "Level date should not be null when value exists");
+                assertNotNull(value.getUnits(), "Units should not be null when value exists");
+                assertEquals(testLocationLevelId, value.getLocationLevelId(), 
+                    "Location level ID should match request");
+                
+                System.out.println("Found test data for location: " + testLocationLevelId);
```
MikeNeilson, line 114: Use logger.

NOTE, slf4j *should* also be used in tests, but java util logging is fine as well.

integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CwmsLocationLevelDAOTestIT.java
```diff
@@ -0,0 +1,361 @@
...
+            return; // Skip for non-CWMS databases
+        }
+        
+        String testLocationLevelId = "TEST_LOCATION.Stage.Const.0.Test";
+        String testLocationId = "TEST_LOCATION";
+        
+        DataTransaction tx = null;
+        try
+        {
+            tx = db.newTransaction();
```
MikeNeilson, line 228: DataTransaction implements `AutoCloseable` which means you can put it in try-with-resources but it's not going to get flagged if it's not in the try-with.

The root opening of a transaction *SHOULD* be in the try-with-resources.

MikeNeilson: See you're next test in this file.

integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CwmsLocationLevelDAOTestIT.java
```diff
@@ -0,0 +1,361 @@
...
+            }
+            dao.close();
+        }
+    }
+    
+    @Test
+    @EnableIfTsDb({"CWMS-Oracle"})
+    public void testErrorHandling_InvalidLocationId() throws Exception
+    {
+        if (dao == null)
```
MikeNeilson, line 264: Suggest retrieving the DAO in each test.

integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CwmsLocationLevelDAOTestIT.java
```diff
@@ -0,0 +1,361 @@
...
+            // Should not throw exception, just return null
+            LocationLevelValue value = dao.getLatestLocationLevelValue(tx, invalidLocationId);
+            
+            // Most likely will be null for invalid ID
+            // But if not null, it means the ID exists (unlikely for this test ID)
+            // Either way, no exception should be thrown
+        }
+        finally
+        {
+            dao.close();
```
MikeNeilson, line 283: The stateless DAO shouldn't have a close method.

java/opendcs/src/main/java/decodes/cwms/CwmsLocationLevelDAO.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+    
+    
+    /**
+     * Close the DAO and free resources
+     */
+    @Override
+    public void close()
+    {
+        // Nothing to close in stateless implementation
```
MikeNeilson, line 298: throw a runtime except to make sure this never gets called until we can ditch DaoBase

java/opendcs/src/main/java/decodes/cwms/CwmsLocationLevelDAO.java
```diff
@@ -0,0 +1,300 @@
...
+
+/**
+ * Data Access Object for CWMS Location Level data.
+ * This class provides methods to read and store location level data
+ * directly from/to the CWMS database.
+ *
+ * Implements LocationLevelDAI interface following stateless pattern.
+ * All operations require a DataTransaction for true stateless behavior.
+ */
+public class CwmsLocationLevelDAO extends DaoBase implements LocationLevelDAI
```
MikeNeilson, line 36: It doesn't look like anything from DaoBase is actually used.

java/opendcs/src/main/java/decodes/cwms/algo/ResEvapAlgo.java
```diff
@@ -303,13 +308,15 @@ public void beforeAllTimeSlices() throws DbCompException
         siteDAO = tsdb.makeSiteDAO();
         timeSeriesDAO = tsdb.makeTimeSeriesDAO();
         crd = new CwmsRatingDao((CwmsTimeSeriesDb) tsdb);
+        CwmsLocationLevelDAO locLevDAO = null;
```
MikeNeilson, line 311: Future work, you should be able to call OpenDcsDatabase::getDao here... however, clearly that never gets passed into the algorithm code.

java/opendcs/src/main/java/decodes/db/LocationLevelSpec.java
```diff
@@ -0,0 +1,94 @@
...
+import decodes.sql.DbKey;
+import opendcs.dao.CachableDbObject;
+
+import java.util.Date;
+
+/**
+ * Inner class representing a Location Level Specification
+ * Implements CachableDbObject for caching support
+ */
+public class LocationLevelSpec implements CachableDbObject
```
MikeNeilson, line 12: Don't put new code in decodes/db, put it in org/opendcs.

Also, unless we want to figure out a proper "OpenDCS" "Location Meta Data" these should be under a /cwms names to indicate they aren't generic.

java/opendcs/src/main/java/opendcs/dai/LocationLevelDAI.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.List;
+
+/**
+ * Abstract interface for Location Level Data Access Interface.
+ * This interface defines the core operations for reading location level data,
+ * which can be implemented for different database types (CWMS, HDB, OpenDCS, etc.)
+ * 
+ * All operations are stateless and require a DataTransaction.
+ */
+public interface LocationLevelDAI extends DaiBase, OpenDcsDao
```
MikeNeilson, line 20: Move to org.opendcs namespace. Additionally I think you can ditch the DaiBase now.

MikeNeilson (changes requested):

java/opendcs/src/main/java/opendcs/dai/LocationLevelDAI.java
oskarhurst: should I make a new DAI folder in the org.opendcs folder?

java/opendcs/src/main/java/opendcs/dai/LocationLevelDAI.java
MikeNeilson, line 1: Rename this to `SiteReferenceMetaData`, `CwmsLocationLevelDao` can keep it's name, but the form is what we came up with as a better "reference name" for the concept. 

The name, even without those that have interacted with cwms before, has caused confusion such that a retrieval was limited to only "length" based measurements. The actual CWMS "Location Levels" can return other measurements as well. Additionally this will, over time, become an interface into other possible meta data for a location.

java/opendcs/src/main/java/org/opendcs/model/cwms/LocationLevelValue.java
MikeNeilson, line 1: There's still likely a better name, but for the moment how does `ReferenceValue` sound?

MikeNeilson (commented):
Naming reads better, but we still have the bad "CWMS" naming at the root interface level. Have a suggestion, might also be worth just excepting the questionable names for the moment.

java/opendcs/src/main/java/decodes/cwms/algo/ResEvapAlgo.java
```diff
@@ -329,6 +336,17 @@ public void beforeAllTimeSlices() throws DbCompException
             throw new DbCompException("Failed to load rating table", ex);
         }
 
+        if (secchi == 0){
+            try (org.opendcs.database.api.DataTransaction tx = locLevDAO.getTransaction())
```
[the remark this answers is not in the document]
MikeNeilson: Fair, and it looks like there's bigger problems. However, in the future, let's try to have tests in place for any changes, especially something like this. That lack of tests that define behavior has been costing us a lot of time.

You have have to manually provide the casting.

There may be away in the future to handle that typing with generics, but that seems like another "future PR" thing.

integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CwmsLocationLevelDAOTestIT.java
```diff
@@ -268,11 +270,11 @@ public void testInterfaceCompatibility() throws Exception
             String testLocationLevelId = "TEST_LOCATION.Stage.Const.0.Test";
```
MikeNeilson, line 270: favor real, or at least realistic site names.

MikeNeilson (changes requested):
Looks better, but now i add an important snag.

"levels" can change over time, having a new effective date, with the design as of now how would one run an algorithm and use the appropriate effective value as of the time of calculation (the time in do time slice, or appropriate aggregate window start/end.)

MikeNeilson:
> Looks better, but now i add an important snag.
> 
> "levels" can change over time, having a new effective date, with the design as of now how would one run an algorithm and use the appropriate effective value as of the time of calculation (the time in do time slice, or appropriate aggregate window start/end.)

integrationtesting/opendcs-tests/src/test/resources/data/cwms/cwms_location_level_test_data.sql
```diff
@@ -57,65 +57,177 @@ BEGIN
...
+            P_OFFICE_ID        => v_office_id
+        );
+        DBMS_OUTPUT.PUT_LINE('Created FTPK-Lower.Stage.Const.0.Test level at -14 days');
+        
+        -- Value from 21 days ago
+        CWMS_20.CWMS_LEVEL.STORE_LOCATION_LEVEL(
+            P_LOCATION_LEVEL_ID => 'FTPK-Lower.Stage.Const.0.Test',
+            P_LEVEL_VALUE      => 119.75,
+            P_LEVEL_UNITS      => 'ft',
+            P_EFFECTIVE_DATE   => SYSDATE - 21,
```
MikeNeilson, line 125: don't worry about it in this one, for now, but in the future favor full specific date times.

MikeNeilson (changes requested):
Only one nitpick this time.

Otherwise seems like a reasonable start. Would like to not even use "Level" in the name of the functions but I can't think of a better name at the moment so that will have to wait.

java/opendcs/src/main/java/decodes/cwms/CwmsLocationLevelDAO.java
```diff
@@ -384,8 +384,8 @@ public CTimeSeries getLocationLevelRange(DataTransaction tx, String locationLeve
                             }
                             catch (Exception ex)
                             {
-                                log.warn("Failed to convert units from {} to {}: {}",
-                                        dbUnits, targetUnits, ex.getMessage());
+                                log.atWarn().setCause(ex).log("Failed to convert units from {} to {}: {}",
```
MikeNeilson, line 387:
```suggestion
                                log.atWarn().setCause(ex).log("Failed to convert units from {} to {}",
```

MikeNeilson (changes requested):
last nitpick for this one.

PR #1314, opened 2025-08-19
update logging part 8 (into main)

ktarbet:
progress on #1268

## Solution

update logging to slf4j


java/opendcs/src/main/java/decodes/db/Platform.java
```diff
@@ -757,14 +718,14 @@ else if (!thisps.equals(thatps))
 		// Check properties for equality.
 		if (!PropertiesUtil.propertiesEqual(properties, op.properties))
 		{
-			Logger.instance().debug3("Platform Properties differ");
+			log.trace("Platform Properties differ");
```
MikeNeilson, line 721: I would just remove the log statements from equals.

MikeNeilson (dismissed):

PR #1352, opened 2025-09-05
Start on OpenDCS SQLite implementation.

### LINKED ISSUE #1264: Create SQLite implementation
The XML Database, while it has worked for the better part of 3 decades, is causing a fair amount of consternation in how to support it in light of the new DAO system.

To move various features more inline with each other using SQLlite as the replacement for the XML "edit-db" will streamline a lot of code while reducing maintainance burden. There will be several DAOs that can be easily shared between the core OpenDCS implementations that would have to be independent for just the XML.

Create an OpenDCS-SQLlite `OpenDcsDatabase` implementation.

COntinue to use the XML database and deal with the general maintenance thereof and trying to sort out any sort of "transaction handling" we get for free with a relational database system.

**Additional context**
NOTE: XML as an implementation will not be removed until the SQLite implementation is sufficiently tested.
We will still need to update the XML implemenation, as the DatabaseIOs are their own maintenance issue, but knowing that an XML OpenDcsDatabase implementation could be assumed to be only for export/import should let the general implementation be a bit easier.

MikeNeilson:
## Problem Description

Fixes #1264. 

## Solution

- Create new provider for SQLite for database management app and testing
- Created RandomKeyGenerator (TODO: but likely not this PR, enable ability to use UUID as key)

## how you tested the change

Automated tests

Triggers not yet in place (different syntax in sqlite, of course) so those expected to fail.
Also seeing several "database locked" which is odd but is likely just pointing out areas of left open connections.

Frankly more worked immediately that I actually expected.

## Where the following done:


java/opendcs/src/main/java/decodes/tsdb/TimeSeriesDb.java
```diff
@@ -193,6 +194,16 @@ public TimeSeriesDb(String appName, javax.sql.DataSource dataSource, DecodesSett
         {
             determineTsdbVersion(conn, this);
             postConnectInit(appName, conn);
+            
+			DatabaseMetaData meta = conn.getMetaData();
+			if (meta.getDatabaseProductName().contains("PostgreSQL"))
```
ktarbet: Consider if a DRY approach for similar code in SqlDatabaseIO constructor would work.

java/opendcs/src/main/java/opendcs/util/sql/ConnectionInTransaction.java
```diff
@@ -28,4 +28,10 @@ public void setAutoCommit(boolean state) throws SQLException
             throw new SQLException("This connection is in a specific transaction. The auto commit state cannot be changed.");
         }
     }
+
+    @Override
+    public void close()   throws SQLException
+    {
+        throw new SQLException("cant vclosw ot mine");
```
ktarbet: spelling

ktarbet (commented):
Looking good!  checkpoint review.

java/opendcs/src/main/java/decodes/sql/SqlDatabaseIO.java
```diff
@@ -217,12 +219,21 @@ public SqlDatabaseIO(javax.sql.DataSource dataSource, DecodesSettings settings)
...
             determineVersion(conn);        
             setDBDatetimeFormat(conn);
+            DatabaseMetaData meta = conn.getMetaData();
+            if (meta.getDatabaseProductName().contains("PostgreSQL"))
+            {
+                listSubtraction = "except";
+            }
+            else
+            {
+                listSubtraction = "minus";
```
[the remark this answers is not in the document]
MikeNeilson: In this particular case, Postgres is the special case. minus is the SQL standard.

java/opendcs/src/main/java/decodes/tsdb/TimeSeriesDb.java
MikeNeilson: It will eventually. Until we've removed all of the SqlDatabaseIO based classes it's a fair amount of work to avoid some of that duplication.

ktarbet:
This experiment https://github.com/HydrologicEngineeringCenter/dss8-playground/tree/main/src/main/java/mil/army/usace/hec/sqldss  is worth watching for possible use in OpenDCS.

MikeNeilson:
I'm going to leave this open for now for reference, but I'm going to start this over again. Given several changes since this was started we can achieve better separation of concerns.

Like only implementing the new style DAOs and thus just dealing with what is already implemented vs total tests. Should let us get the baseline change itself merged in sooner which will help its development.

Issue #1269, opened 2025-08-05
7.0 - Update all logging usage to slf4j


MikeNeilson:
**Component to be refactored**

*... everything ... *

**Purpose of refactoring**
Eliminate custom code that is proving difficult to maintain.

**Proposed method**
replace all ilex.util.Logger usages with sl4fj.
delete ilex.util.Logger and implementations

**Additional context**
NOTE: this will initially break GUI logging
NOTE2: initial java.util.logging "implementation" will be provided by the install and tests. configuration still TDB.
NOTE3: also breaks the DACQEvent Logger, however, JUL and other systems provide ways to "hook" into them and get information so we should be able to determine some sort of adapter setup for downstream usages to implement.

PR #1397, opened 2025-10-01
Update logging to slf4j. Part 42

MikeNeilson:
Partial #1269

### DIFF
--- a/java/opendcs/src/main/java/lrgs/db/LrgsDatabase.java
+++ b/java/opendcs/src/main/java/lrgs/db/LrgsDatabase.java
@@ -1,13 +1,17 @@
 /*
+* Where Applicable, Copyright 2025 OpenDCS Consortium and/or its contributors
 *
-*  This is open-source software written by ILEX Engineering, Inc., under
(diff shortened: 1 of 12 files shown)


java/opendcs/src/main/java/lrgs/db/LrgsDatabase.java
```diff
@@ -321,7 +322,10 @@ public void commit()
 			conn.commit();
 			conn.clearWarnings();
 		}
-		catch(SQLException ex) {}
+		catch(SQLException ex)
+		{
+			throw new LrgsDatabaseException("Unable to commit changes.", ex);
```
ktarbet, line 327: behavior change?

ktarbet (dismissed):

java/opendcs/src/main/java/lrgs/db/LrgsDatabase.java
MikeNeilson: yeah, but it seemed appropriate, that's quite the silent error. Could just change it to logging though.

ktarbet (approved):

PR #1439, opened 2025-10-11
Features/1370 user tables

### LINKED ISSUE #1370: User tables
With the creation of the REST API it is not a good idea to provide each user a "database user", users should have an entry in a separate table that then can be tied to various identity providers.

Create a set of tables to contain:
1. OpenDCS Specific user information
2. User roles
3. Identify provider links.

Keeping using database accounts and have a giant security hole

**Additional context**
Will diagram more later, likely 4 tables "users", "user role mapping", "roles", and identity providers.

MikeNeilson:
## Problem Description

Fixes #1370 . 

## Solution

- create tables
- create dao interface
- create dao impl
- various mappers (demonstrating how to deal with joins in Jdbi).
- tests

Some more tweaks are likely needs but this should allow work to begin on the rest_api/wev_ui side

## how you tested the change

integration tests created

## Where the following done:


MikeNeilson:
@krowvin you've setup a lot of various services in your homelab and other places. Anything overall design-wise missing that needs to be accounted for general usage you can think of?

@DanielTOsborne same question. Goal here to is to let OpenDCS, especially the rest_api/webui properly support non-database users in various contexts, including standalone.

java/opendcs/src/main/java/org/opendcs/database/impl/opendcs/dao/UserManagementImpl.java
```diff
@@ -0,0 +1,458 @@
...
+{
+
+    private static final RoleMapper ROLE_MAPPER = RoleMapper.withPrefix(null);
+    private static final IdentityProviderMapper PROVIDER_MAPPER = IdentityProviderMapper.withPrefix(null);
+
+    @Override
+    public List<User> getUsers(DataTransaction tx, int limit, int offset) throws OpenDcsDataException
+    {
+        Handle handle = getHandle(tx);
+        handle.getJdbi().installPlugin(new Jackson2Plugin());
```
oskarhurst: Dont you only have to install the Jackson Plugin once per jdbi instance. wouldn't this install it again each time?

java/opendcs/src/main/resources/db/OpenDCS-Postgres/schema/V7.3__users.sql
```diff
@@ -0,0 +1,44 @@
+create table identity_provider (
+    id bigserial primary key,
+    name varchar(256) not null unique,
+    type varchar(256) not null,
+    config jsonb default '{}'::jsonb,
+    updated_at timestamptz not null --- TODO make trigger
+);
+
+create table opendcs_user(
+    id bigserial primary key,
```
oskarhurst: do you think it is worth adding cascading deletion?

java/opendcs/src/main/resources/db/OpenDCS-Postgres/schema/V7.3__users.sql
```diff
@@ -0,0 +1,44 @@
...
+create table user_roles(
+    user_id bigint references opendcs_user(id) not null,
+    role_id bigint references opendcs_role(id) not null,
+    primary key (user_id, role_id)
+);
+
+
+create table user_identity_provider(
+    user_id bigint references opendcs_user(id) not null,
+    subject text not null unique,
```
oskarhurst: are we sure the subject will always be unique?

oskarhurst (dismissed):
Everything looks good to me, I just had some questions. A lot of boiler plate but I didn't see anything missing.

java/opendcs/src/main/resources/db/OpenDCS-Postgres/schema/V7.3__users.sql
```diff
@@ -0,0 +1,44 @@
...
+    id bigserial primary key,
+    preferred_name_provider bigint references identity_provider(id),
+    email text not null unique,
+    preferences jsonb default '{}'::jsonb,
+    created_at timestamptz not null default now(),
+    updated_at timestamptz not null --- TODO make trigger
+);
+
+create table opendcs_user_password(
+    user_id bigint references opendcs_user(id) not null,
```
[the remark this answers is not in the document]
MikeNeilson: Oh, right, the user_id can be the primary key.

java/opendcs/src/main/resources/db/OpenDCS-Postgres/schema/V7.3__users.sql
MikeNeilson: Something to discuss. With it, you might accidentally delete, but it does make that operation easier. Without it, you're force to make sure everything is deleted. I'm not hard set on any particular method.

MikeNeilson:
Quick note, I intend to fill out the IdentityProvider, and especially BuiltInIdentityProvider mechanism in a follow up PR. That will involve a lot of code related to password management and some new dependencies and this is already fairly large so I don't want to overload reviewers.

oskarhurst (approved):
Everything Looks good to me.

PR #1461, opened 2025-10-15
Update logging to slf4j. Try with usage. part 65

MikeNeilson:
Removed slf4j/ilex helper utilities.
Removed dbupdate tools (we are using flyway now.)
Removed Rebalance tool (files it depends on where already moved.)

## Problem Description

Partial #1269

... This might be the last one of actual logging statements -- at least that aren't cleanup from us missing things. Will wait to merge everything more I run the `ilex.util.Logger` search again.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -20,7 +20,7 @@ jsch="0.1.55"
 aspectj="1.8.10"
 json="20231013"
 slf4j-api= { strictly = "2.0.17" }
-logback = "1.5.18" 
+logback = "1.3.15" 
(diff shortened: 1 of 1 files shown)


MikeNeilson:
... nope, not the last one, just found a few more.

java/opendcs/src/main/java/opendcs/dbupdate/DbUpdate.java
ktarbet: consider removing launch scripts: dbupdate, dbupdate.bat

ktarbet (approved):

PR #1489, opened 2025-10-31
Remove edgecase for -P option.

MikeNeilson:
## Problem Description

Discover that we aren't processing the -P option correctly anymore.

## Solution

Describe your solution.

## how you tested the change

## Where the following done:


MikeNeilson:
> So help me out with context. Why is this an edge case? We're processing -P differently now, or not at all?

i put code in to make sure a -P option would get passes on to the app correctly, and when wasn't required did the wrong correction. (this stems from fixing -pp for the launcher and mixing up the casing.) So -P is not longer an "edge" case and the pass through can use the simple form. Without this change it was stripping the -P before it got to the app.

PR #1494, opened 2025-11-05
Setup gradle build extension mechanism.

MikeNeilson:
Mechanism makes dealing with required upstream changes during development on downstream projects easier.

## Problem Description

As the RestAPI isn't merged in yet, while still needing upstream changes, dealing with the constant creation of releases to test small things will get beyond annoying.

## Solution

Setup gradle dependency substitution that can be triggered by a depending project so that local change are used instead of the fixed artifact

## how you tested the change

Not tested directly, mechanism has been used in several other projects. Single file add, no changes to any logic.

Do need to add docs to the dev-docs though.

## Where the following done:

### DIFF
--- a/docs/source/dev-docs.rst
+++ b/docs/source/dev-docs.rst
@@ -779,7 +779,7 @@ While the actual versioned migrations *MUST* stay the same, the other organizati
 if you think you have a superior organization for these data.
 
 
-Using OpenDCS Jars in you project
+Using OpenDCS Jars in your project
 =================================
 
 Except for tests and the installer zip, project jars are available on maven central.


ktarbet (dismissed):
In the rest API I think you will need something like the following in settings.gradle


```gradle
if(hasProperty('LibrarySourcesDir')) {
    def externalDir = new File("$librarySourcesDir")
    if (externalDir.exists()) {
        externalDir.eachDir() { directory ->
            def hecBuildExtFile = new File(directory, "opendcs-build-ext.gradle")
            if (hecBuildExtFile.exists()) {
                gradle.ext.externalLibDir = directory
                apply from: hecBuildExtFile
            }
        }
    }
}
```

ktarbet (dismissed):

ktarbet (approved):

PR #1497, opened 2025-11-06
Task/1495 java 25-- maybe, got at least to 21.

### LINKED ISSUE #1495: Update to Java 21 compiler and min target (main only)
As the title says.

MikeNeilson:
## Problem Description

Fixes #1495. 

## Solution

- Upgrade to Gradle 9.2, adjust mostly docs, some other tasks.
- Set target to Java 25... discovered our systemstubs library can't mock SystemExit even at newest version
  - It relied on the SecurityManager which is completely removed now.
- Set target to Java 21, at least one integration test passed.

I suspect I'll also need to update most of the CWMS Dependencies, but will also look further into the stubs. There may be a way, beyond just fixing all of our code to not need it during tests (which would be ideal.)

## how you tested the change

Integration tests that already exist

## Where the following done:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,6 +11,9 @@ on:
       - "main"
       - "7.0"
       - "8.0"
+# Update this value if we need a different Jdk to be used for sonar analysis.
+env:
+  sonarJdk: 21
 
 jobs:
   tests:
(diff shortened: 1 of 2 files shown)


MikeNeilson:
However, that's enough of a change it should likely be a followup PR.

ktarbet:
> Next question is, there are a _lot_ of dependencies that we couldn't update past their version due being stuck on Java 8. Should I handle that in this PR, or a follow up?

I prefer a separate PR for dependencies; This PR is large enough.

ktarbet (dismissed):

PR #1477, opened 2025-10-28
Feature/1473 logger

MikeNeilson:
## Problem Description

Fixes #1473. 

Describe the problem you are trying to solve.

## Solution

- Create default logback for production and integration tests
- Adapt decj (mostly bash script) to process logging arguments into logback parameters.
  - @ktarbet will look at decj.bat, work around provided in documentation.
- Update Docker images to use JSON exporter for structured logging
- Documentation.

NOTE: Rendering to GUI was intentionally not solved during this work.

## how you tested the change

Manually, verified elements logged. 

## Where the following done:

- [ ] Tests. Check all that apply:
   - [ ] Unit tests created or modified that run during ant test.
   - [ ] Integration tests created or modified that run during integration testing
         (Formerly called regression tests.)
   - [ ] Test procedure descriptions for manual testing
- [x] Was relevant documentation updated?
- [x] Were relevant config element (e.g. XML data) updated as appropriate

If you aren't sure leave unchecked and we will help guide you to want needs changing where.

PR #1502, opened 2025-11-09
update decj.bat to support -d , -l, and -a options

ktarbet:
initial update of decj.bat to be in sync with decj bash script.      More arguments can be added later.

followup on : https://github.com/opendcs/opendcs/pull/1477


MikeNeilson (approved):

MikeNeilson:
> We should consider moving the powershell for dec.bat etc.

I certainly don't see an issue with doing that. There aren't any windows platforms we'd support that wouldn't have powershell.

PR #1514, opened 2025-11-18
Bugfix/rc10 testing feedback

MikeNeilson:
## Problem Description

LockList and ReleaseLock were mistakenly removed.
Parameter processing was breaking several normal usages

## Solution

Restore classes - renamed package to avoid confusion.
Simply parameter processing. Provide standard "--" to indicate any remaining parameter should be presented as-is

## how you tested the change

Manually, using runApp and on cli to verify output was correct.

## Where the following done:


MikeNeilson:
closing to reopen so can target the right branch.

PR #1521, opened 2025-11-24
Refactor/prep-for-removal of jooq usage

### LINKED ISSUE #237: Rewrite CWMS Packages to use JDBI instead of JOOQ for access to the CWMSDB
CWMS Timeseries Database and Related components

**Purpose of refactoring**
While we *think* we might be validly using JOOQ here... better not to risk it long term.

**Proposed method**
Queries will be wrapped by Jdbi instead of going though the JOOQ layer.

**Additional context**
Pretty sure this is mostly paranoia, but it is a good first place to start demonstrating Jdbi within the main code base.

MikeNeilson:
## Problem Description

Fixes #237. 

## Solution

Have Connection object derived from a Jdbi instance. Update UserManagementDaoImpl to use Handle from Transaction.

Gave XmlDatabaseWrapper it's own newTransaction to keep old behavior as it is currently still needed.

## how you tested the change

Existing integration tests. No changes in behavior.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/org/opendcs/database/JdbiTransaction.java
+++ b/java/opendcs/src/main/java/org/opendcs/database/JdbiTransaction.java
@@ -9,7 +9,7 @@ import org.opendcs.database.api.OpenDcsDataException;
 
 import opendcs.util.sql.WrappedConnection;
 
-public class JdbiTransaction implements DataTransaction
+public final class JdbiTransaction implements DataTransaction
 {
     final Handle jdbiHandle;
 


java/opendcs/src/main/java/org/opendcs/database/JdbiTransaction.java
```diff
@@ -0,0 +1,63 @@
...
+import java.sql.Connection;
+import java.util.Optional;
+
+import org.jdbi.v3.core.Handle;
+import org.opendcs.database.api.DataTransaction;
+import org.opendcs.database.api.OpenDcsDataException;
+
+import opendcs.util.sql.WrappedConnection;
+
+public class JdbiTransaction implements DataTransaction
```
ktarbet: Should this be autoclosable?  Who is in charge of closing the handle?

MikeNeilson: yes, the `DataTransaction` is in charge of ending the transaction. with the way the Handle instance is retrieved some portion of the calling entity is in charge.

java/opendcs/src/main/java/org/opendcs/database/JdbiTransaction.java
```diff
@@ -0,0 +1,62 @@
...
+        {
+            return (Optional<T>)Optional.of(new WrappedConnection(jdbiHandle.getConnection(), c -> {}));
+        }
+        else if (Handle.class.equals(connectionType))
+        {
+            return (Optional<T>)Optional.of(jdbiHandle);
+        }
+        else
+        {
+            return Optional.empty();
```
adamkorynta: What kind of connection type is this case? If the XML case, shouldn't a specific class type get sent in? using any other type seems like an error

MikeNeilson: Nothing available, hence why the Test just passes in an integer.

I can see an argument for why maybe that it shouldn't be optional though this PR isn't the place to tweak that.

The XML database has it's own wrapper that sticks to the simple Connection for which we have an `XmlConnection` implementation to store the file location. The XML stuff is really a stop gap until we replace with SQLlite.

Though I do like Optional here. we will also have HTTP connections in the future and it's possible, though admittedly perhaps not likely, that a given implement way want to "i'd prefer there but I'll fallback on old behavior if I have to." That kind of control shouldn't be done with exceptions.

Perhaps returning `FailableResult` would be better here.

MikeNeilson: Interesting idea, I'm not sure if Closable it entirely correct but constraining to something should be done... I will investigate a bit.

adamkorynta: I think adding that `extends AutoCloseable` is enough. Without any constraint it really feels like the caller has to know implementation details for what kinds of things are supported.

PR #1522, opened 2025-11-25
add in ability to parameterize office id on database creation

adamkorynta:
## Problem Description

Fixes https://github.com/opendcs/rest_api/issues/464

Two issues: 
- connection pool cache is keyed without the office id which means the office can't change per user
- no way to parameterize in the office id while also reading from the tsdb_property table

## Solution

- Add office id to pool cache key.
- allow Properties parameterization in Database factory creation

## how you tested the change

Tested via opendcs-rest-api

## Where the following done:

See: https://github.com/opendcs/rest_api/pull/590


java/opendcs/src/main/java/org/opendcs/database/DatabaseService.java
```diff
@@ -67,13 +69,27 @@ public static OpenDcsDatabase getDatabaseFor(String appName, DecodesSettings set
      * @return
      * @throws DatabaseException any error with class initialization or loading of properties
      */
-    public static OpenDcsDatabase getDatabaseFor(javax.sql.DataSource dataSource) throws DatabaseException
+    public static OpenDcsDatabase getDatabaseFor(DataSource dataSource) throws DatabaseException
+    {
+        DecodesSettings settings = decodesSettingsFromJdbc(dataSource, new Properties());
```
MikeNeilson: I'm not sure I see the value of initializing the Properties Object here.

PR #1550, opened 2025-12-10
Task/1545 transaction context

### LINKED ISSUE #1545: Create Context object for OpenDcsDatabase and DataTransaction
DataTransaction, OpenDcsDatabase, various Daos

**Purpose of refactoring**
The DAOs at runtime need access to various information, like "is this query for Postgres" or schema version information to put together various queries. Or the KeyGenerator instance.

Initially The OpenDcsDatabase instance was injected into by Constructor into the new DAO; however, that means that while the DAO is stateless pr transaction, it's not stateless per database instance which it should be (and MUST be for some of the proposed Rest API CWMS office id usages.)

**Proposed method**
Create a "TransactionContext" object that can be provided to the DataTransaction instances so that DAOs can retrieve this information.

**Additional context**
There will be one more DAO that does this wrong, for #1223, this is intentional to allow work to continue. A follow up will correct that and others for more appropriate behavior.

At this time EnumSqlDao is the only other one with the issue.

MikeNeilson:
## Problem Description

Fixes #1545. 

## Solution

1. Cleanup (remove) Injection of the OpenDcsDatabase instance into the DAO (which made them stateful again)
2. Create a `TransactionContext` object that allows retrieving appropriate runtime database/implementation information
3. Apply ServiceAnnotation and usage Lookup classes from Netbeans/OpenIDE to retrieve the 3 DAOs adjusted. (CwmsLocationLevelDao is an additional candidate but I didn't want to deal with that in what is essentially a "demo/initial design PR".

Still need to account for differences in implementations which will likely require some changes to the lookup method or some additional annotations... or understanding of the "path" or other parameters to the `ServiceAnnotation`. We will need to create a `UserManagementDao` implementation for CWMS and we can address that behavior at that time.

## how you tested the change

No functional change. Existing Integration tests apply.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/org/opendcs/database/dai/EquipmentModelDao.java
+++ b/java/opendcs/src/main/java/org/opendcs/database/dai/EquipmentModelDao.java
@@ -34,7 +34,7 @@ public interface EquipmentModelDao extends OpenDcsDao
     /**
      * Retrieve EquipmentModel by name.
      * @param tx Transaction object for the request.
-     * @param name equipment model name name to look for. Comparison will be upper(name) = upper(name in database).
+     * @param name equipment model name name to look for. Comparison will be upper(name) = upper(name) in database.
      * @return Equipment Model if found, otherwise empty.
      */
     Optional<EquipmentModel> getEquipmentModel(DataTransaction tx, String name) throws OpenDcsDataException;
--- a/java/opendcs-api/src/main/java/org/opendcs/database/api/TransactionContext.java
+++ b/java/opendcs-api/src/main/java/org/opendcs/database/api/TransactionContext.java
@@ -8,7 +8,7 @@ public interface TransactionContext
 {
  
     /**
-     * KeyGenerator if auto generated keys aren't used.
+     * KeyGenerator alternative to auto generated keys.
      * @return
      */
     <T extends Generator> Optional<T> getGenerator(Class<T> generatorClass);


java/opendcs/src/main/java/opendcs/dao/EnumSqlDao.java
```diff
@@ -15,64 +15,81 @@
...
 import decodes.sql.DecodesDatabaseVersion;
+import decodes.sql.KeyGenerator;
 import decodes.tsdb.DbIoException;
+import decodes.util.DecodesSettings;
 
 /**
  * Data Access Object for writing/reading DbEnum objects to/from a SQL database
  * @author mmaloney Mike Maloney, Cove Software, LLC
  */
+@ServiceProvider(service = EnumDAI.class, path = "DEFAULT")
```
adamkorynta: "DEFAULT" seems equivalent to a null path, which implies support for XML. Should be typed to the different implementations:
```
@ServiceProviders({
		@ServiceProvider(service = EnumDAI.class, path = "OPENTSDB"),
		@ServiceProvider(service = EnumDAI.class, path = "CWMS"),
		@ServiceProvider(service = EnumDAI.class, path = "HDB"),		
})
```

For a case that we actually want a default fallback, adding the no path makes sense:
```
@ServiceProviders({
		@ServiceProvider(service = EnumDAI.class),
		@ServiceProvider(service = EnumDAI.class, path = "OPENTSDB"),
		@ServiceProvider(service = EnumDAI.class, path = "CWMS"),
		@ServiceProvider(service = EnumDAI.class, path = "HDB"),		
})
```

java/opendcs/src/main/java/org/opendcs/database/dai/EquipmentModelDao.java
```diff
@@ -34,7 +34,7 @@ public interface EquipmentModelDao extends OpenDcsDao
     /**
      * Retrieve EquipmentModel by name.
      * @param tx Transaction object for the request.
-     * @param name equipment model name name to look for. Comparison will be upper(name) = upper(name) in database.
+     * @param name equipment model name name to look for. Comparison will be upper(name) = upper(name in database).
```
ktarbet:
```suggestion
     * @param name equipment model name name to look for. Comparison will be upper(name) = upper(name) in database.
```

PR #1562, opened 2025-12-16
Untabify DBImport.java

agilmore2:
## Problem Description
Is this how we're untabifying things?

### DIFF
--- a/java/opendcs/src/main/java/decodes/xml/TopLevelParser.java
+++ b/java/opendcs/src/main/java/decodes/xml/TopLevelParser.java
@@ -46,675 +46,675 @@ import ilex.xml.*;
  */
 public class TopLevelParser implements XmlObjectParser
 {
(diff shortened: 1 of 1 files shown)


MikeNeilson (changes requested):
I think something went wonky with your find and replace.

You've stripped all blank lines.

the `\s+$` should only remove extra spaces at the end of a line with text on it.... granted I'm using VS Code's find and replace when I do that so your IDE may require a slightly different syntax.

agilmore2:
Interesting. I'm using VSCode, and when I regex search for \s+$ and replace with nothing, that's what happens.

I'll revert that, thanks!

MikeNeilson:
> Interesting. I'm using VSCode, and when I regex search for \s+$ and replace with nothing, that's what happens.
> 
> I'll revert that, thanks!

Strange, last time I wanted to remove newlines I had to really work to make it happen.

PR #1567, opened 2025-12-16
Set CRLF for .bat files

### LINKED ISSUE #1564: decj.bat behavior change - JAVA_HOME environment variable should be optional
**Describe the bug**
dcstool_start on windows doesn't work.

**To Reproduce**
Steps to reproduce the behavior:  reported by @jbatterman 

```
c:\opendcs-7.0.16-RC15\bin>dcstool_start -d3 -P "C:\Users\user1\AppData\Roaming\.opendcs\NWO-T7.profile" -l file.log
The system cannot find the batch label specified - _ParseDone
ERROR: JAVA_HOME environment variable is not set.
Please set JAVA_HOME to a Java + installation.

```

**Expected behavior**
use Path to find Java if JAVA_HOME is not set.

ktarbet:
fixes  #1564 

batch files need to keep CRLF line ending

Tested fix using Ubuntu to build distZip, then copying file to windows to verify.

### DIFF
--- a/install/src/main/dist/scripts/decj.bat
+++ b/install/src/main/dist/scripts/decj.bat
@@ -15,7 +15,7 @@
 ::   -D<prop>=<value>  Pass to JVM 
 ::
 :: Environment Variables:
-::   JAVA_HOME         Path to Java installation (required)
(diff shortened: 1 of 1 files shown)


MikeNeilson (approved):
Cherry-pick this into 7.0 as well.

PR #1584, opened 2025-12-21
Extend the element filter from TopLevelParser to the DatabaseParser,

### LINKED ISSUE #1575: "Platform only" dbimport as merged in #125 has never worked?
I can find no artifact, clear back to 7.0.2, where the "platform only" flag to dbimport works. The newer code is much easier to follow than the never ending function in 7.0.2 where it was merged, but it still doesn't work.

The element filter is created and applied to the XML Parser in 
https://github.com/opendcs/opendcs/blob/7.0/java/opendcs/src/main/java/decodes/dbimport/DbImport.java#L535

I don't see why it doesn't work, but I'm not that familiar with the XML parser. How much longer will XML even be available?

**To Reproduce**
1. Initialize a new local XML database
2. dbimport dev.xml.log to initialize database
3. dbimport -p prod.xml.log
4. Inspect log or resulting database to see that routing, schedule entries and datasources from prod.xml.log were imported

Filtered versions of subject XML files:
[dev.xml.log](https://github.com/user-attachments/files/24252761/dev.xml.log)
[prod.xml.log](https://github.com/user-attachments/files/24252762/prod.xml.log)
Files named .xml.log because .xml files are filtered out by GH.

**Expected behavior**
The -p flag is supposed to limit the dbimport targets to Platform, PlatformConfig, NetworkList, EquipmentModel, and Site.
In the delivered version to a customer, it works as designed.

 - OS: Windows/Linux
 - Version 7-cs1-2 works, every version from 7.0.2 onward doesn't seem to work.

**Additional context**
Discovered by BCHydro when attempting 7.0.14 upgrade.

agilmore2:
…where it will actually affect the import.

## Problem Description

Fixes #1575. 

## Solution

The DatabaseParser created by the TopLevelParser needs to have the same element filter propagated to it.

## how you tested the change

manually ran dbimport with the changes against an XML backend, noted that the sample xml files in #1575 are imported correctly. Further, the -W flag also correctly only overwrites the destination categories of objects, and doesn't delete categories filtered out.

I'd like to see some automatic tests confirming this code functions as expected, but not sure how to go about implementing such things. Any pointers?

## Were the following done:

   - [X] Test procedure descriptions for manual testing
- [N/A] Were relevant config element (e.g. XML data) updated as appropriate


agilmore2:
Ready for review with integration tests on DBImport/Export and platform only and overwrite flags. 
Picked up on a DBExport bug with a static field. Wouldn't ever show up unless we're calling it twice within the same program, but it's fixed.

integrationtesting/opendcs-tests/src/test/java/org/opendcs/regression_tests/DBImportExportTest.java
```diff
@@ -0,0 +1,124 @@
...
+        IOUtils.closeQuietly(exportXMLFile);
+        //assertEquals(golden, cleanedOut, "Output Doesn't match expected data."); //how to handle modified times?!?
+
+        try
+        {
+            int existingplats = configuration.getDecodesDatabase().platformList.size();
+            int existingrs = configuration.getDecodesDatabase().routingSpecList.size();
+            assertEquals(plats, existingplats, "Expected number of platforms not found!");
+            assertEquals(rs, existingrs, "Expected number of data sources not found!");
+        } catch (Throwable e)
```
MikeNeilson: try catch unneccessary here. 

If you are concerned about some sort of throwing situation use `assertDoesNotThrow`

integrationtesting/opendcs-tests/src/test/java/org/opendcs/regression_tests/DBImportExportTest.java
```diff
@@ -0,0 +1,124 @@
...
+            case "normal":
+                log.info("Loading platform, routing spec, etc with no extra flags.");
+                break;
+            case "platonly":
+                log.info("Loading platforms with platform only flag.");
+                platargs.add("-p");
+                break;
+            case "platover":
+                log.info("Loading platforms with platform only and overwrite flags.");
+                platargs.addAll(Arrays.asList("-W", "-y", "-p"));
```
MikeNeilson: Using `-W` here and above is the most likely issue, I don't believe the system is setup to do anything *other* than a total wipe.

MikeNeilson (approved):

PR #1680, opened 2025-12-31
#1605 Support for Computation Execution from REST API

### LINKED ISSUE #1605: Web UI Computations-Computations: No Run Computations action
Currently unable to manually Run Computations to test or for any other reason in the Web UI.   This is a highly used feature of the desktop client.

Add the Run Computations button/feature.

zack-rma:
## Problem Description

Fixes #1605. 

## Solution

Adds SSE endpoint to trigger computation execution from REST API. 

## how you tested the change

Integration testing in progress.

## Where the following done:


java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ComputationResources.java
```diff
@@ -477,4 +494,96 @@ public Response deleteComputation(@Parameter(required = true, description = "Uni
...
+			DbComputation comp = dai.getComputationById(DbKey.createDbKey(computationId));
+
+			List<TimeSeriesIdentifier> tsiList = new ArrayList<>();
+
+			for(String tsId : tsList)
+			{
+				tsiList.add(tsDai.getTimeSeriesIdentifier(tsId));
+			}
+
+			executor = Executors.newCachedThreadPool();
```
MikeNeilson: this should be thread pool configured at the container level, or at least higher level than here. this will create a new pool on each request and cold lead to resource exhaustion.

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ComputationResources.java
```diff
@@ -477,4 +494,96 @@ public Response deleteComputation(@Parameter(required = true, description = "Uni
...
+			tags = {"REST - Computation Methods"},
+			responses = {
+					@ApiResponse(responseCode = "202", description = "Successfully initiated execution of computation"),
+					@ApiResponse(responseCode = "400", description = "Missing required computationid parameter"),
+					@ApiResponse(responseCode = "404", description = "Computation with the specified ID not found"),
+					@ApiResponse(responseCode = "500", description = "Internal Server Error")
+			}
+	)
+	public Response runComputation(@Context SseBroadcaster broadcaster, @Suspended final AsyncResponse asyncResp,
+			@Parameter(required = true, description = "Unique Computation ID",
```
MikeNeilson: I want to say this should really be the comp name but then this api isn't really general purpose.

java/opendcs/src/main/java/decodes/tsdb/ComputationApp.java
```diff
@@ -353,40 +348,14 @@ else if (regressionTestModeArg.getValue()
...
-									 dataCollection.getTasklistHandle().markComputationFailed(rn);
-								}
-							}
-							comp.getTriggeringRecNums().clear();
-							log.debug("End of computation '{}'", comp.getName());
-						}
+						ComputationExecution execution = new ComputationExecution(theDb);
+
+						execution.execute(List.of(comps), dataCollection);
+						compsTried += execution.getComputesTried();
```
MikeNeilson: perhaps we should just return a `CompResults` record from `::execute`?

java/opendcs/src/main/java/decodes/tsdb/ComputationApp.java
```diff
@@ -830,15 +801,15 @@ private void executeTimedComp(DbComputation tc, Calendar timedCompCal, DataColle
 			{
 				try (MDCCloseable mdc = MDC.putCloseable("computation", concreteClone.getName()))
```
MikeNeilson: this MDC should be moved into the Computation execution class methods.

java/opendcs/src/main/java/decodes/tsdb/ComputationExecution.java
```diff
@@ -0,0 +1,175 @@
...
+
+	public ComputationExecution(TimeSeriesDb db)
+	{
+		this.db = db;
+	}
+
+	public DataCollection execute(List<DbComputation> toRun, DataCollection theData)
+	{
+		// Execute the computations
+		for(DbComputation comp :toRun)
```
MikeNeilson: shouldn't this eventually just call the execute single method?

java/opendcs/src/main/java/decodes/tsdb/ComputationExecution.java
MikeNeilson: overall i like it.

MikeNeilson (changes requested):
Seems like an excellent start.

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ComputationResources.java
```diff
@@ -477,4 +494,96 @@ public Response deleteComputation(@Parameter(required = true, description = "Uni
...
+			for(String tsId : tsList)
+			{
+				tsiList.add(tsDai.getTimeSeriesIdentifier(tsId));
+			}
+
+			executor = Executors.newCachedThreadPool();
+
+			executor.submit(() ->
+			{
+				OutboundEvent event = new OutboundEvent.Builder()
```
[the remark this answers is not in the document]
zack-rma: Added a generated UUID task ID

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ComputationResources.java
zack-rma: Leaving this for now since the ID can be grabbed from the catalog endpoint.

java/opendcs/src/main/java/decodes/tsdb/ComputationApp.java
zack-rma: Moved

java/opendcs/src/main/java/decodes/tsdb/ComputationExecution.java
zack-rma: Yes, swapped for single execute method

java/opendcs/src/main/java/decodes/tsdb/ComputationExecution.java
```diff
@@ -0,0 +1,239 @@
...
+					msg = msg + " -- No TSID assigned.";
+				else
+					msg = msg + " -- TSID '" + parmRef.tsid.getUniqueString() + "' does not exist in db.";
+				log.atWarn().setCause(ex).log(msg);
+				numErrors++;
+			}
+		}
+	}
+
+	public static final class CompResults
```
MikeNeilson: You can use the record type now. We're java 21+ in main.

MikeNeilson (commented):
Seems reasonable. I suspect a lot of challenges will be discovered after some actual usage but that's not something I'd expect to be fixed on the first go.

java/opendcs/src/main/java/decodes/tsdb/ComputationExecution.java
```diff
@@ -199,41 +201,27 @@ private void executeSingleComp(DbComputation tc, Date since, Date until, DataCol
...
-		{
-			return computesTried;
-		}
+	}
 
-		public boolean isSuccess()
-		{
-			return numErrors == 0;
-		}
+	public interface ProgressListener
```
MikeNeilson: i like the idea 
1. move to own class, will be reused
2. maybe take a logging event
3. look at [logging](https://github.com/opendcs/opendcs/blob/main/java/opendcs-api/src/main/java/org/opendcs/logging/spi/LoggingEventProvider.java) and how the gui uses it

it's likely to have a separate stream for the log and this for comp run state *or* filter to log output to the user... hence the MDC stuff. 

alt the listener itself gets a self contain named log to make that even easier. (e.g most logger instances are static but that isn't a hard requirement.

java/opendcs/src/main/java/decodes/tsdb/ProgressListener.java
```diff
@@ -0,0 +1,26 @@
+package decodes.tsdb;
+
+import org.opendcs.utils.logging.OpenDcsLoggerFactory;
+import org.slf4j.Logger;
+import org.slf4j.event.Level;
+
+public abstract class ProgressListener
+{
+	private static final Logger PROGRESS_LOGGER = OpenDcsLoggerFactory.getLogger();
```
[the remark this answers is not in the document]
zack-rma: Fixed

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ComputationResources.java
```diff
@@ -477,4 +284,262 @@ public Response deleteComputation(@Parameter(required = true, description = "Uni
...
+
+		try(ComputationDAI dai = getLegacyTimeseriesDB().makeComputationDAO();
+			TimeSeriesDAI tsDai = getLegacyTimeseriesDB().makeTimeSeriesDAO();
+			SiteDAI siteDai = getLegacyTimeseriesDB().makeSiteDAO())
+		{
+			DbComputation comp = dai.getComputationById(DbKey.createDbKey(computationId));
+
+			String taskID = UUID.randomUUID().toString();
+
+			final Instant startTime = start != null ? Instant.parse(start) : Instant.parse("1800-01-01T12:00:00Z");
```
MikeNeilson: ... let's fail if this aren't provided, or use a more sane default, like -1D to now.

This is fine for our tests but some places have 50+ years of data and this would tank things.

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/util/DTOMappers.java
MikeNeilson: This seems like a lot of general knowledge in one spot. Seems like something we can follow, but hopefully improve on the pattern CDA made, using the ServerAnnotation/Loader mechanisms... though again, future work. First make it work, then make it work and pretty.

java/opendcs/src/main/java/decodes/tsdb/ComputationExecution.java
```diff
@@ -0,0 +1,216 @@
...
+									// Should not happen! We checked first.
+									listener.onProgress("Unexpected duplicate time series.", Level.ERROR, ex);
+									numErrors++;
+								}
+							}
+						}
+					}
+				}
+
+				comp.apply(dataCollection, getTsDb());
```
MikeNeilson: Probably good to have an onProgress update here indicating success.

integrationtesting/opendcs-tests/src/test/java/org/opendcs/odcsapi/res/it/ComputationResourcesIT.java
```diff
@@ -501,4 +660,133 @@ void testPostAndDeleteComputation() throws Exception
...
+
+				int foundCount = 0;
+				for (InboundSseEvent sseEvent : events)
+				{
+					if (sseEvent.getName().equalsIgnoreCase("Results"))
+					{
+						foundCount++;
+					}
+				}
+				assertEquals(10, foundCount, "Expected 10 SSE Result events, found " + foundCount);
```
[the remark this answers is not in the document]
zack-rma: Updated to use the size of the expected TS list

integrationtesting/opendcs-tests/src/test/java/org/opendcs/odcsapi/res/it/ComputationResourcesIT.java
```diff
@@ -127,8 +174,126 @@ void init() throws Exception
...
+		tv2.setFlags(VarFlags.TO_WRITE);
+		ts2.addSample(tv2);
+		tv2 = new TimedVariable(Date.from(Instant.parse("2012-01-01T06:00:00Z")), 0.21, 0);
+		tv2.setFlags(VarFlags.TO_WRITE);
+		ts2.addSample(tv2);
+
+		storeTimeSeries(ts2);
+
+		// get TS DbKeys and save to the identifiers
+		response = given()
```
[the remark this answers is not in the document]
zack-rma: Updated storeTimeSeries method

MikeNeilson (changes requested):
Okay, once the timer names make more sense that should be it.

MikeNeilson (dismissed):

MikeNeilson:
> @zack-rma I suspect the current failure is just a timing issue, I'm rerunning it now.

Sorry, that unfortunately wasn't it. Not really sure why that would start failing though. Previously it was an issue with what wasn't or wasn't running inside the same JVM and the time series caching being too aggressive. That's where I'd start, but again, not totally sure.

PR #1681, opened 2026-01-05
#1651: Swapped ConfigSensor min/max presentation in Web UI

### LINKED ISSUE #1651: Web UI Config - Sensor Min/Max ranges swapped
**Describe the bug**
The Config Sensor Valid Range Min/Max values appear to be swapped in the Web UI.  For example, the Config MRR-AKIA Min = 36 and Max = 0.

**To Reproduce**
1. Open Web UI Config that contains sensor range information (MRR-AKIA for NWK profile on HEC test server)
2. Click on the first Stage sensor (sensor 2)
3. See Valid Range

**Expected behavior**
Expect Valid Range to match what is in the client.

**Screenshots**
Web UI:
Image

7.0.15 Client
Image

 - OS: Windows 11 Microsoft Edge
 - Client Version 7.0.15

zack-rma:
## Problem Description
Fixes #1651. 

## Solution
Swapped indexes for ConfigSensor minimum and maximum value fields in web UI. Added check for null min/max values to prevent empty values being represented as max float value in UI.

## how you tested the change

Tested locally in web UI.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/decodes/sql/ConfigListIO.java
+++ b/java/opendcs/src/main/java/decodes/sql/ConfigListIO.java
@@ -1,5 +1,5 @@
 /*
-* Where Applicable, Copyright 2026 OpenDCS Consortium and/or its contributors
+* Where Applicable, Copyright 2025 OpenDCS Consortium and/or its contributors
 * 
 * Licensed under the Apache License, Version 2.0 (the "License"); you may not
 * use this file except in compliance with the License. You may obtain a copy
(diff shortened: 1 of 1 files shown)


java/opendcs/src/main/java/decodes/sql/ConfigListIO.java
```diff
@@ -266,8 +266,8 @@ private void readConfigSensors(DbKey platformConfigId, PlatformConfig pc)
                             equipId, _pcList.getDatabase());
                 }
 
-                if (hasAbsMin) cs.absoluteMin = absMin;
-                if (hasAbsMax) cs.absoluteMax = absMax;
+				cs.absoluteMin = hasAbsMin ? absMin : Double.MIN_VALUE;
```
adamkorynta: I don't see an absolute compare against `Const.undefinedDouble`, this would come back as a valid value

PR #1712, opened 2026-01-16
Gradle updates (fix warnings , and update gradle

ktarbet:
gradle updated using:

```bash
 ./gradlew wrapper --gradle-version 9.2

```


MikeNeilson:
don't forget to update the Docker file for any updates.

never mind, looks like it just failed to pull a dependency from hec's nexus and that's why it failed.

PR #1713, opened 2026-01-17
Devops/UI layout improvements

MikeNeilson:
## Problem Description

Some initial cleanup for the React Web UI

## Solution

Strip various CSS, rearrange components using bootstrap tooling. Correct behavior or storybook play functions and decorators

## how you tested the change

Visual changes, checked with a combination of `npm run dev` and `npm run storybook` to see rendering behavior.

Added additional "Stories" for some components like the mode picker and made adjustments given that support.

Existing tests confirm previously defined behavior.

## Where the following done:


ktarbet (approved):
I was able to use:

npm run storybook
and
npm run dev.

when using npm run dev,  I stopped at the login screen,  I'm not sure how to start the rest-api.

I tried gradlew :opendcs-rest-api:tomcatRun 

based on the opendcs-rest-api/Readme.md  (it may need an update)

MikeNeilson:
Need to fix the failure, there's not actually a test involved so it really shouldn't be failing. Appears to have something to do with running in the CI environment and thus "mostly headless."

MikeNeilson:
> when using npm run dev, I stopped at the login screen, I'm not sure how to start the rest-api.

You have to start it manually, the easiest at the moment is to do a `docker compose up -d --build` from the project root. `./gradlew runWeb --info` is also suitable but make sure it's a different shell (doesn't run in the background).

MikeNeilson:
NOTE: will cleanup the extra action workflows before final merge.. They've been useful to testing so don't want to loose them just yet.

Issue #1779, opened 2026-02-11
[UI] DecodesScriptEditor component


MikeNeilson:
As users must be able to interact with DecodesScripts, we need a Decodes Script Editor.

PR #1789, opened 2026-02-11
Initial decodes scripts components.

MikeNeilson:
## Problem Description

Partial for #1779. 

## Solution

NOTE: for the purpose of these components it is assumed that the Decodes "Config" is already available, rendering and manipulation of the config itself will be handled next.

- Created
  - ScriptHeader component (name, data order, header type)
  - Format Statement editor (with available sensors and ability to set unit conversions
    - An initial setup of syntax highlighting was tested, however proper setup will be in
      a separate PR as it needs some specific work.
  - DecodesSampleArea
    - Can Render output from "/decodes" endpoint.
  - DecodesScript reducer

Known issues:
1. The input changes need to be debounced, at the moment this is a lot of flickering an internal/temporary state changes.
2. No way to add or remove statements (that's next)

## how you tested the change

Manually, still exploratory, may be pivoting direction given the breadth of components required to implement this, may barrel forward. Definitely won't connect to a database in this PR.

NOTE: will now be creating automatic storybook tests. A few exist but definitely needs more.

## Where the following done:

### DIFF
--- a/javascript/opendcs-web-ui-react/src/pages/decodes/configs/script/ClassicEditor.tsx
+++ b/javascript/opendcs-web-ui-react/src/pages/decodes/configs/script/ClassicEditor.tsx
@@ -190,8 +190,8 @@ const ClassicFormatStatementEditor: React.FC<
         width: "2em",
         render: renderDragArrows,
       },
-      { data: "label", width: "10em", render: renderLabel },
-      { data: "format", render: renderFormat },
+      { data: "label", width: "10em", render: renderLabel, defaultContent: "" },
+      { data: "format", render: renderFormat, defaultContent: "" },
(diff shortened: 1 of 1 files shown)


MikeNeilson:
> Running Firefox automatically opens a new tab and seems to be stuck on ProtectedRoute/loading

It's not stuck, that's all that particular story ever shows, what the loading screen looks like.

> Click Decodes ScriptEditor / Default and I get this error:
> 
> DataTables warning: table id=DataTables_Table_6 - Requested unknown parameter 'label' for row 0, column 1. For more information about this error, please see https://datatables.net/tn/4
> 
> Click Decodes ScriptEditor / Create from empty gives similar errors.


I'll check that out before merge, that shouldn't happen.

PR #1795, opened 2026-02-17
React web UI Implement Protected Route

oskarhurst:
## Problem Description

Unprotected routes can briefly flash before redirecting to login page.

## Solution

Implement a protected route to handle authorization and routing.

## how you tested the change

## Where the following done:

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,23 +1,23 @@
 [versions]
 #General/Shared Versions
 oracle = { strictly = "23.5.0.24.07" }
-postgres = "42.7.9"
+postgres = "42.7.10"
 commons-logging = "1.3.5"
 commons-net = "3.12.0"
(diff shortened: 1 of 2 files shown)


javascript/opendcs-web-ui-react/src/contexts/app/AuthProvider.tsx
```diff
@@ -10,21 +10,31 @@ interface ProviderProps {
...
     auth.logout().then((_value: void) => {
-      setUser(undefined);
+      setUser(null);
       navigate("/login");
     });
   };
 
   const authValue: AuthContextType = {
     user,
+    isLoading: user === undefined,
```
MikeNeilson: what happens on logout after you've set the user to `null` instead of `undefined`?

perhaps the default for the user should just be `null`? or stick with setting to undefined, I believe `setUser(undefined)` should work.

PR #1797, opened 2026-02-18
Update the react web ui to use scss with bootstrap

oskarhurst:
Update the react webUI project to us scss styling

## how you tested the change

## Where the following done:

### DIFF
--- a/javascript/opendcs-web-ui-react/src/App.tsx
+++ b/javascript/opendcs-web-ui-react/src/App.tsx
@@ -3,6 +3,7 @@ import Login from "./pages/auth/login";
 import { TopBar, SideBar, ProtectedRoute, PublicOnlyRoute } from "./components/layout";
 import { ModeIcons } from "./components/ModeIcon";
 
+import { Container } from "react-bootstrap";
 import { Navigate, Route, Routes } from "react-router-dom";
 import { Platforms } from "./pages/platforms";
 import { Algorithms } from "./pages/computations/algorithms";
@@ -18,7 +19,7 @@ function App() {
         onToggleSidebar={() => setSidebarOpen((prev) => !prev)}
         sidebarOpen={sidebarOpen}
       />
-      <div className="odcs-layout d-flex">
+      <Container fluid className="odcs-layout d-flex">
         <Routes>
           <Route element={<PublicOnlyRoute />}>
             <Route path="/login" element={<Login />} />
@@ -36,7 +37,7 @@ function App() {
             </Route>
           </Route>
         </Routes>
-      </div>
+      </Container>
     </>
   );
 }
--- a/javascript/opendcs-web-ui-react/src/styles/components/_sidebar.scss
+++ b/javascript/opendcs-web-ui-react/src/styles/components/_sidebar.scss
@@ -18,7 +18,6 @@
     padding: 1rem 1rem 0.5rem;
     font-size: 0.8125rem;
     font-weight: 600;
-    text-transform: uppercase;
     letter-spacing: 0.06em;
     color: var(--odcs-sidebar-section-text);
     user-select: none;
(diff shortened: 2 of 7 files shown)


MikeNeilson (changes requested):
Okay, so I was with this until i started seeing files for things we aren't likely to pull in.

The use of scss i agree with but espicially the modals we are trying to avoid so the styling is likely not required. While we need to move every concept over this transition is a good opprotunity to take lessoned learned and clean things up as we go. I sure i haven't done everything right so for but this PR is doing too much. I suggest sticking to the css that has already been ported, noting that some of it was intentionally deleted.


Please read the section of the readme discussing styles and more direct use of bootstrap without trying to fight it, the creation of an opendcs specific theme can come later.

oskarhurst:
> Okay, so I was with this until i started seeing files for things we aren't likely to pull in.
> 
> The use of scss i agree with but espicially the modals we are trying to avoid so the styling is likely not required. While we need to move every concept over this transition is a good opprotunity to take lessoned learned and clean things up as we go. I sure i haven't done everything right so for but this PR is doing too much. I suggest sticking to the css that has already been ported, noting that some of it was intentionally deleted.
> 
> Please read the section of the readme discussing styles and more direct use of bootstrap without trying to fight it, the creation of an opendcs specific theme can come later.

We can trim some of the changes to not include so many of the components and get rid of the modal components. Is your biggest concern that this is to many changes to make all at once or do you think it is best practice to use the default bootstrap styling with minimal css. My initial thoughts was that is might be best to rip the band aid off early and convert every thing to scss so if we ever decided to make more changes in the future there would be less we would have to change, but I am flexible with moving in a different direction.

MikeNeilson:
> We can trim some of the changes to not include so many of the components and get rid of the modal components. Is your biggest concern that this is to many changes to make all at once or do you think it is best practice to use the default bootstrap styling with minimal css. 

Yes

> My initial thoughts was that is might be best to rip the band aid off early and convert every >thing to scss so if we ever decided to make more changes in the future there would be less we would have to change, but I am flexible with moving in a different direction.

Yeah, the scss part makes sense. We do want an "OpenDCS Theme" in the future, and bootstrap itself supports scss to do so. I'm more concerned about bringing elements that appear to have been fighting with themselves. You included styling for select2, but we aren't currently using select2 - perhaps we will, but at the moment I've be sticking to "as simple and direct as possible" to focus on the large elements like general organization.

The statement in the readme of "just do it the bootstrap way" is to provide all of us "a target" and avoid paralysis of choice. The choice is made so we can focus on our data and systems and how users interact with them.

Also, I saw you set the top bar to 56px and pull the "lg" size of the button. that LG was intentional. I fought the items too small to be readable... my eyes suck.

One thing I agree is useful to bring back is the collapsing sidebar, just look at the bootstrap5 docs and see if there's an existing built-in way to do it. I believe so; a lot of that code was written before we settled specifically on bootstrap 5 and older versions didn't have the behavior, hence the extra code -- that we may not need anymore.

javascript/opendcs-web-ui-react/src/styles/components/_sidebar.scss
```diff
@@ -0,0 +1,97 @@
...
+  flex-direction: column;
+  flex-shrink: 0;
+  overflow-y: auto;
+  border-right: 1px solid rgba(255, 255, 255, 0.06);
+
+  .odcs-sidebar__section-title {
+    padding: 1rem 1rem 0.5rem;
+    font-size: 0.8125rem;
+    font-weight: 600;
+    text-transform: uppercase;
```
MikeNeilson: Same as other comment, if we want uppercase somewhere we'll do it intentionally.

javascript/opendcs-web-ui-react/src/App.tsx
```diff
@@ -1,34 +1,42 @@
...
   return (
     <>
       <ModeIcons />
-      <TopBar />
-      <Container fluid className="page-content d-flex">
+      <TopBar
+        onToggleSidebar={() => setSidebarOpen((prev) => !prev)}
+        sidebarOpen={sidebarOpen}
+      />
+      <div className="odcs-layout d-flex">
```
MikeNeilson: back to Container

MikeNeilson (changes requested):
Looks good, a few nitpicks.

MikeNeilson (approved):

PR #1824, opened 2026-03-10
Move organization DAO to new DAO design standards. Correct API output…

MikeNeilson:
… definition.

## Problem Description

Realized Organizations endpoint could use simplified behavior, and it need an output formatting correction.

## Solution

Output empty list instead of not-found for organizations not currently being supported.
Set schema implementation of list element to ApiOrganization instead of String.

## how you tested the change

Update integration test to expect data for CWMS and Empty list for OpenDCS Postgres
## Where the following done:


MikeNeilson:
The code that Sonar is saying is not covered, most definitely is.

PR #1832, opened 2026-03-17
Computations page

tlelv:
## Problem Description

Adds Computations page to react web ui

## Solution

Describe your solution.

## how you tested the change

## Where the following done:


javascript/opendcs-web-ui-react/src/pages/computations/computations/ComputationParamsTable.tsx
oskarhurst: @tlelv computation have different parameters than algorithms. to this need to be reworked to match computation parameters

Algorithm Param table:

Image

Computation Param table:

Image

MikeNeilson (changes requested):
Overall I like it, I have one change I think should happen, the passing of the algorithm instead of just the parameters.

I'd also like to find some time to experiment with the changes in the branch myself in case I'm forgetting something important. Hopefully today, might be tomorrow or Friday though.

MikeNeilson (commented):
That looks good. As I said before, definitely want to interact with it a bit. And I suspect @oskarhurst and @ktarbet will have some input as well.

MikeNeilson (changes requested):
Design looks good. My only gripe is that you appear to have created some rather useful utility functions in a way that isn't sharable with future work.

MikeNeilson (approved):

MikeNeilson (approved):

PR #1843, opened 2026-03-26
Feature/1220 datasource dao

### LINKED ISSUE #1220: DataSource DAO

None

**Additional context**

MikeNeilson:
## Problem Description

Fixes #1220. 

## Solution

Implement a DataSource in the new stateless way. Along with Appropriate JDBI Mappers to handle conversion from database to java object.

NOTE: I have intentionally *NOT* updated the REST API to use this. There is a data consistent issue with the original data source delete that @oskarhurst and I agree should be an error. The handling of which suggested some improved exception hierarchy that goes beyond the scope of this PR and thus I will do in a follow up.

The issue:
The Data Source delete will allow the data source to delete itself from *other* groups. This is wrong, as those groups assume they can depend on that Data Source this should cause an error. This does happen in the new code due to the foreign key constraints. However, the REST API code, and the test setup/teardown, assume the original behavior.

## how you tested the change

New Integration Test.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/org/opendcs/database/dai/DataSourceDao.java
+++ b/java/opendcs/src/main/java/org/opendcs/database/dai/DataSourceDao.java
@@ -15,7 +15,7 @@ public interface DataSourceDao extends OpenDcsDao
     /**
      * Retrieve a DataSource by DbKey.
      * 
-     * Unlikely other DAOs at this time the DataSource will not have been "prepared for exec."
+     * Unlike other DAOs, at this time, the DataSource will not have been "prepared for exec."
      * 
      * @param tx active transaction to use
      * @param id Identifier of the DataSource desired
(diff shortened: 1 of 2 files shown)


java/opendcs/src/main/java/org/opendcs/database/dai/DataSourceDao.java
```diff
@@ -0,0 +1,69 @@
...
+
+import decodes.db.DataSource;
+import decodes.sql.DbKey;
+
+public interface DataSourceDao extends OpenDcsDao
+{
+    /**
+     * Retrieve a DataSource by DbKey.
+     * 
+     * Unlikely other DAOs at this time the DataSource will not have been "prepared for exec."
```
ktarbet:
```suggestion
     * Unlike other DAOs, at this time, the DataSource will not have been "prepared for exec."
```

PR #1861, opened 2026-04-06
Initial design of properties source/builder.

MikeNeilson:
## Problem Description

Partial for #1860. 

Our usage of "Properties" from various sources is very consistent in behavior, and thus involves a rather large about of duplicate code. @ktarbet and I have discussed that multiple times having both recently modified code that followed these patterns of "if available from this source use, otherwise use the next one." I had some inspiration and decided to start on a more unified system.

An alternatively appropriate would be a bunch of methods, but I'm liking the `Property<T>` class more and more to handle various usages where either the `Property<T>` or `Property.Builder<T>` could be kept around and reused.

## Solution

Create a Property<T> and Property Builder object. After Property<T> is built allows retrieving the value with the given configuration when desired but would handle several elements of validation (noted: this is of course TODO.

Please note that the majority of required behavior *has not* been implemented. This was a quick rough out of the idea with description of intention so that we can all comment on it before doing too much work.

## how you tested the change

TODO: PR is open for design comment. Once initial agreement, if any, is made I will create tests (and will definitely shoot for 100% branch and line coverage) and update at least one usage to this new system.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/decodes/cwms/algo/ResEvapAlgo.java
+++ b/java/opendcs/src/main/java/decodes/cwms/algo/ResEvapAlgo.java
@@ -209,9 +209,20 @@ final public class ResEvapAlgo extends AW_AlgorithmBase
         _awAlgoType = AWAlgoType.AGGREGATING;
         _aggPeriodVarRoleName = DAILY_EVAP_DEPTH;
         //aggPeriodInterval = IntervalCodes.int_one_day;
-        aggUpperBoundClosed = true;
+        // So the below are actually backawards from what's desired.
+        // During the creation of the Property build setup that 
+        // the proprty init flow calls this method, then sets up the bounds
+        // based on either computation configured property or a default.
+        // so in all algorithms, the algorithm configured desired was just
+        // getting ignored.
+        // For ResEvap it would appear that has caused the test data to be
+        // basically "off by one" the whole time as the aggregate bounds weren't as expected.
+        // Might have an affect on the daylight savings processing as well.
+        aggUpperBoundClosed = Property.property("aggUpperBoundClosed", Boolean.class)
+									  .withDefaultValue(false) // true
+									  .build();
         aggLowerBoundClosed = Property.property("aggLowerBoundClosed", Boolean.class)
-									  .withDefaultValue(false)
+									  .withDefaultValue(true) // false
 									  .build();
     }
 
--- a/java/opendcs/src/main/java/decodes/cwms/algo/ResEvapAlgo.java
+++ b/java/opendcs/src/main/java/decodes/cwms/algo/ResEvapAlgo.java
@@ -51,6 +51,7 @@ import org.opendcs.annotations.PropertyRequirements;
 import org.opendcs.database.api.OpenDcsDataException;
 import org.opendcs.model.cwms.CwmsSiteReferenceValue;
 import org.opendcs.utils.logging.OpenDcsLoggerFactory;
+import org.opendcs.utils.properties.Property;
 import org.slf4j.Logger;
 
 
@@ -209,7 +210,9 @@ final public class ResEvapAlgo extends AW_AlgorithmBase
         _aggPeriodVarRoleName = DAILY_EVAP_DEPTH;
         //aggPeriodInterval = IntervalCodes.int_one_day;
         aggUpperBoundClosed = true;
-        aggLowerBoundClosed = false;
+        aggLowerBoundClosed = Property.property("aggLowerBoundClosed", Boolean.class)
+									  .withDefaultValue(false)
+									  .build();
     }
 
     //Initialized hourly water temperature profiles and return double[] of WTP of the previous timeSlice before base.
--- a/java/opendcs/src/main/java/decodes/tsdb/DbAlgorithmExecutive.java
+++ b/java/opendcs/src/main/java/decodes/tsdb/DbAlgorithmExecutive.java
@@ -27,6 +27,7 @@ import org.opendcs.annotations.algorithm.Input;
 import org.opendcs.annotations.algorithm.Output;
 import org.opendcs.utils.AnnotationHelpers;
 import org.opendcs.utils.logging.OpenDcsLoggerFactory;
+import org.opendcs.utils.properties.Property;
 import org.slf4j.Logger;
 import org.slf4j.MDC;
 import org.slf4j.MDC.MDCCloseable;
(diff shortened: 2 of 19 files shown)


java/opendcs/src/main/java/decodes/tsdb/DbAlgorithmExecutive.java
```diff
@@ -151,39 +152,48 @@ public void init( DbComputation comp, TimeSeriesDb tsdb )
...
-			catch(NumberFormatException ex)
-			{
-				this.maxMissingTimeForFill = DecodesSettings.instance().maxMissingTimeForFill;
-				log.atWarn()
-				   .setCause(ex)
-				   .log("Bad maxMissingTimeForFill property '{}' will use default of {}",
-				   		s, maxMissingTimeForFill);
-			}
-		}
+		this.maxMissingValuesForFill = Property.property("maxMissingValuesForFill", Integer.class)
```
ktarbet, line 155: Without Generic type on the builder - it could look something like:

```java
this.maxMissingTimeForFill = Property.property("maxMissingTimeForFill") 
											   .withSources(comp, algo, DecodesSettings.instance())
											   .build()
											   .asInteger(default:0);
```

But what if the 
```java
int maxMissingValuesForFill  
//was instead:
Property <Integer> maxMissingValuesForFill
//  then the type is baked in. 
```
Is that where your going with the Property Class?

java/opendcs/src/main/java/decodes/util/DecodesSettings.java
```diff
@@ -845,4 +847,48 @@ public DecodesSettings asCopy()
         newSettings.loadFromProperties(props);
         return newSettings;
     }
+
+    @Override
+    public void setProperty(String name, String value)
```
ktarbet, line 851: consider if  this would be friendly.

```java
 public void setProperty(String name, int value);
 public void setProperty(String name, double value);
 public void setProperty(String name, String value);
 public void setProperty(Property value);
...
```

ktarbet (commented):
I like how the Property builder will clean up tedious code.

java/opendcs/src/main/java/decodes/tsdb/DbAlgorithmExecutive.java
MikeNeilson: I wasn't thinking that, but does seem like a good idea, most of the code the sets up properties gets things ready that passes it to something else, that would make sense in those contexts.

java/opendcs/src/main/java/decodes/util/DecodesSettings.java
MikeNeilson: This is from the HasProperties interface, at the moment I'm leaving it alone, making a `<T> void setProperty(String name, T value)` and making use of the PropertyConverter may make some sense going forward.

java/opendcs/src/main/java/decodes/tsdb/DbAlgorithmExecutive.java
MikeNeilson: Realized, I don't think we can avoid the passing in of the type. As is rather common in a lot of libraries, it's the only way to get around type erasure, we need an instance of Class<T> so we can perform certain operations.

java/opendcs/src/main/java/decodes/cwms/algo/ResEvapAlgo.java
```diff
@@ -208,8 +209,21 @@ protected void initAWAlgorithm()
...
+        // So the below are actually backawards from what's desired.
+        // During the creation of the Property build setup that 
+        // the proprty init flow calls this method, then sets up the bounds
+        // based on either computation configured property or a default.
+        // so in all algorithms, the algorithm configured desired was just
+        // getting ignored.
+        // For ResEvap it would appear that has caused the test data to be
+        // basically "off by one" the whole time as the aggregate bounds weren't as expected.
+        // Might have an affect on the daylight savings processing as well.
+        aggUpperBoundClosed = Property.property("aggUpperBoundClosed", Boolean.class)
```
MikeNeilson, line 221: @oskarhurst I know there's are backwards, see the not.

If you look at https://github.com/opendcs/opendcs/pull/1861/changes#diff-279132afc9ba923f4a5a449a31ea2b2176ea0d2b2948dd91e920927d0245029dR272-R287, you can see that the Algorithm Init is called and *then* the bounds are determined.

So the ResEvap algorithm, and frankly several others, was never using the configured bounds.


As a seperate PR we should move the _awAlgoType to either static + a method to retrieve, or required constructor value on AW_AlgorithmBase so it's always available then and we can call algo init *after* all the baseline props.

java/opendcs/src/main/java/decodes/tsdb/algo/AW_AlgorithmBase.java
```diff
@@ -279,19 +280,17 @@ protected void initAlgorithm( )
...
-		t_string = comp.getProperty("aggLowerBoundClosed");
-		if (t_string != null) 
-			aggLowerBoundClosed = TextUtil.str2boolean(t_string);
-		else // default is true for regular aggregates, false for running aggregates.
-			aggLowerBoundClosed = 
-				_awAlgoType == AWAlgoType.RUNNING_AGGREGATE ? false : true;
+		aggLowerBoundClosed = Property.property("aggLowerBoundClosed", Boolean.class)
+									  .withSources(comp)
+									  .withExistingPropertyFirst(aggLowerBoundClosed)
+									  .withDefaultValue(_awAlgoType == AWAlgoType.RUNNING_AGGREGATE ? false : true)
```
ktarbet: Sonar is suggesting something like:

```suggestion
									  .withDefaultValue(_awAlgoType != AWAlgoType.RUNNING_AGGREGATE)
```

java/opendcs/src/main/java/decodes/tsdb/algo/AW_AlgorithmBase.java
```diff
@@ -279,19 +280,17 @@ protected void initAlgorithm( )
...
-		t_string = comp.getProperty("aggUpperBoundClosed");
-		if (t_string != null) 
-			aggUpperBoundClosed = TextUtil.str2boolean(t_string);
-		else // default is false for regular aggregates, true for running aggregates.
-			aggUpperBoundClosed = 
-				_awAlgoType == AWAlgoType.RUNNING_AGGREGATE ? true : false;
+		aggUpperBoundClosed = Property.property("aggUpperBoundClosed", Boolean.class)
+									  .withSources(comp)
+									  .withExistingPropertyFirst(aggUpperBoundClosed)
+									  .withDefaultValue(_awAlgoType == AWAlgoType.RUNNING_AGGREGATE ? true: false)
```
ktarbet: same sonar suggestion as above

java/opendcs/src/main/java/decodes/tsdb/algo/AW_AlgorithmBase.java
```diff
@@ -1590,7 +1589,7 @@ public String getEvaluatedProperty(String name)
 		if (name.equalsIgnoreCase("aggUpperBoundClosed"))
 			return "" + aggUpperBoundClosed;
 		else if (name.equalsIgnoreCase("aggLowerBoundClosed"))
-			return "" + aggLowerBoundClosed;
+			return "" + aggLowerBoundClosed.get();
```
ktarbet, line 1592: why isn't  aggUpperBoundClosed  above, also using .get()

java/opendcs/src/main/java/decodes/tsdb/DbAlgorithmExecutive.java
```diff
@@ -91,18 +92,18 @@ public abstract class DbAlgorithmExecutive
...
 
-	private int maxMissingValuesForFill;
+	private Property<Integer> maxMissingValuesForFill;
 	private int maxMissingTimeForFill;
 
 	/** Determines open/closed intervals for aggregate periods.
 	 * The default is [lower,upper)
 	 */
-	protected boolean aggLowerBoundClosed = true;
+	protected Property<Boolean> aggLowerBoundClosed; // = true;
```
ktarbet:
```suggestion
	protected Property<Boolean> aggLowerBoundClosed; 
```

java/opendcs/src/main/java/decodes/tsdb/DbAlgorithmExecutive.java
```diff
@@ -91,18 +92,18 @@ public abstract class DbAlgorithmExecutive
...
 
 	/** Determines open/closed intervals for aggregate periods.
 	 * The default is [lower,upper)
 	 */
-	protected boolean aggUpperBoundClosed = false;
+	protected Property<Boolean> aggUpperBoundClosed; // = false;
```
ktarbet:
```suggestion
	protected Property<Boolean> aggUpperBoundClosed; 
```

java/opendcs/src/main/java/org/opendcs/utils/properties/ExistingPropertySource.java
```diff
@@ -0,0 +1,55 @@
...
+{
+    private final Property<T> property;
+
+    public ExistingPropertySource(Property<T> property)
+    {
+        this.property = property;
+    }
+
+    @Override
+    public void setProperty(String name, String value)
```
ktarbet, line 19: consider if we need an error, or deprecation here.

java/opendcs/src/main/java/org/opendcs/utils/properties/ExistingPropertySource.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    public Enumeration getPropertyNames()
+    {
+        return Collections.enumeration(List.of(property.name()));
+    }
+
+    @Override
+    public void rmProperty(String name)
+    {
+        /* do nothing */
```
ktarbet, line 53: consider : deprecated?

java/opendcs/src/main/java/org/opendcs/utils/properties/Property.java
```diff
@@ -0,0 +1,289 @@
...
+import org.opendcs.utils.properties.conversion.PropertyConverter;
+
+import ilex.util.EnvExpander;
+import ilex.util.HasProperties;
+
+/**
+ * Establish a known Property value that is to be retrieved.
+ *
+ * The get and find methods *always* lookup the value when called. So the created instances
+ * can be stored to retrieve at different times.
```
ktarbet:
```suggestion
 * The get and find methods *always* lookup the value when called. 
 * Values are not cached.
```

ktarbet (commented):
The changes in algo/* and ResEvap -   would feel safer with some specific tests.

MikeNeilson:
> The changes in algo/* and ResEvap - would feel safer with some specific tests.

Yeah. the current tests in ResEvap caught the issue. I'm rather concerned at the moment.

There's only a few built in algorithms that appear to depend on this behavior, so it's time to create tests for them.

However, I'm going to create a new PR for those. I think the solution is one of what I mentioned above, it's definitely an issue that these two properties in particular are reinitialized after the algorithm init call.

java/opendcs/src/main/java/decodes/tsdb/algo/AW_AlgorithmBase.java
MikeNeilson: thanks. missed it.

java/opendcs/src/main/java/org/opendcs/utils/properties/ExistingPropertySource.java
MikeNeilson: It's definitely not deprecated.

I think the HasProperties should be HasProperties + like 3-4 interfaces under. 1 for each operations. like right here we just don't care.

java/opendcs/src/main/java/org/opendcs/utils/properties/ExistingPropertySource.java
MikeNeilson: Not deprecated, just nothing to do.

MikeNeilson:
While this particular change is "ready", we've uncovered a rather disturbing design problem around how various default settings were being handled in regards to aggregate algorithms.


I'm going to leave this in draft until we can put some tests in place for the algorithms whose behavior may have been affected by this change.

PR #1868, opened 2026-04-07
Update tomcat to 11.0.15

ktarbet:


gradle/libs.versions.toml
```diff
@@ -69,7 +69,7 @@ greenmail = "2.1.8"
 mockito = "5.21.0"
 byte-buddy = "1.15.3"
 rest-assured = "6.0.0"
-tomcat = "11.0.14"
+tomcat = "11.0.15"
```
MikeNeilson: Also update the docker file. https://github.com/opendcs/opendcs/blob/0cf0a7af10db57aa4f01c9435af60633ab6167b7/Dockerfile#L91

NOTE: the retrieval of the Tomcat tar.gz release into an alpine container is intentional, drastically reduces the attach surface and can be a smaller container in case upstream hasn't done any cleanup.

MikeNeilson (changes requested):

MikeNeilson (approved):

PR #1871, opened 2026-04-09
Resevap fixes 7.0

oskarhurst:
## Problem Description

add resevap docs fix and exception catch changes

## Solution

Describe your solution.

## how you tested the change

## Where the following done:


java/opendcs/src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
```diff
@@ -817,18 +818,24 @@ public void saveTimeSeries(CTimeSeries ts) throws DbIoException, BadTimeSeriesEx
         }
         catch(SQLException ex)
         {
-            String msg = "Error in cwmsTsJdbc.store for '{}'";
+            String msg = "Error in cwmsTsJdbc.store for '" + path + "'";
```
MikeNeilson: Don't do this, the {} is intentionally. I"ll grant that at Error it would almost already be render but don't break out of the logging frameworks variable substitution.

MikeNeilson (changes requested):

MikeNeilson (approved):

Issue #1062, opened 2025-04-22
USGS NWIS Data retrieval


MikeNeilson:
**Is your feature request related to a problem? Please describe.**
The URLs for retrieving data from the USGS are changing

**Describe the solution you'd like**
Update the existing USGS DataSource to:
1. Verify it actually functions
2. Support the new URL
3. Be better structured to handle future changes to the API (`/nwis` will likely go away in the future in favor of new API designs.)

**Describe alternatives you've considered**
Do nothing and let the users short it out.

**Additional context**

ktarbet:
Here are some screenshots of initial design. 

Goals will be to simplify setup for USGS data retrieval.

usgsApiKey is put in the routing spec.

<img width="1204" height="879" alt="Image" src="https://github.com/user-attachments/assets/a5d05bd9-baf8-4521-9f90-257825cfc92a" />

Config Sensor defines what data to read.
Usgs Standard used to define Data type(usgs parameter code)
USGS Stat code for statistic
property:usgsSublocation used for case where multiple sensors for same parameter exist at a USGS location.



<img width="1209" height="894" alt="Image" src="https://github.com/user-attachments/assets/25489ea1-a42a-40a3-bbe7-982acc9dcd41" />


## Common Statistic codes

| Name | USGS Code |
|------|-----------|
| MAXIMUM | 00001 |
| MINIMUM | 00002 |
| MEAN | 00003 |
| AM | 00004 |
| PM | 00005 |
| SUM | 00006 |
| MODE | 00007 |
| MEDIAN | 00008 |
| STD | 00009 |
| VARIANCE | 00010 |
| INSTANTANEOUS | 00011 |
| EQUIVALENT_MEAN | 00012 |
| SKEWNESS | 00013 |

Reference: https://api.waterdata.usgs.gov/ogcapi/v0/collections/statistic-codes

##  Common Parameter codes.

| Name | USGS Code |
|------|-----------|
| WATER_TEMPERATURE | 00010 |
| AIR_TEMPERATURE | 00021 |
| WIND_SPEED | 00035 |
| WIND_DIRECTION | 00036 |
| PRECIPITATION | 00045 |
| RELATIVE_HUMIDITY | 00052 |
| DISCHARGE | 00060 |
| STAGE | 00065 |
| SPECIFIC_CONDUCTANCE | 00095 |
| SALINITY | 00096 |
| RESERVOIR_STORAGE | 72036 |
| SOLAR_RADIATION | 62608 |
[...]

PR #1872, opened 2026-04-10
Introduce DataMessage

ktarbet:
Initial setup for #1062

### DIFF
--- a/java/opendcs/src/main/java/decodes/routing/RoutingSpecThread.java
+++ b/java/opendcs/src/main/java/decodes/routing/RoutingSpecThread.java
@@ -442,7 +442,13 @@ public class RoutingSpecThread extends Thread
 					continue;
 				}
 				*/
-				rm = (RawMessage) dataMessage;
+				if (dataMessage instanceof RawMessage) {
+					rm = (RawMessage) dataMessage;
+				}
+				else {
+					log.error("Unknown DataMessage type: {}", dataMessage.getClass().getName());
+				}
+
 				if (rm == null)
 				{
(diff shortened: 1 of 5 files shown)


java/opendcs/src/main/java/decodes/routing/RoutingSpecThread.java
```diff
@@ -425,10 +427,22 @@ public void run()
...
 			myExec.setSubsystem("acquire");
+			DataMessage dataMessage = null;
 			RawMessage rm = null;
 			try
 			{
-				rm = source.getRawMessage();
+				dataMessage = source.getDataMessage();
+				/* future note: support other types of DataMessage like the comment below.
+				if (dataMessage instanceof DecodedMessage)
+				{
```
MikeNeilson, line 437: Even if you want to make this a no-op and just log, I wouldn't leave this as always assuming it's a RawMessage. Once merged someone will see the change and may try to use it.

Given our level of testing of the Scheduler/SpecThread it may not get caught earlier enough.

MikeNeilson (changes requested):
Nice trick with the additional method on `DataSourceEx`, save us having to deal with updating all of the data sources that are a raw message.

MikeNeilson (approved):

MikeNeilson:
@ktarbet clicked the rebase on the wrong PR, that's what I get for opening two tabs quickly.

ktarbet:
That sonar hotspot is not related to this PR. 

https://sonarcloud.io/project/security_hotspots?id=opendcs_opendcs_ui&pullRequest=1872&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

Issue #291, opened 2023-08-28
Threaded Compproc


MikeNeilson:
**Is your feature request related to a problem? Please describe.**

The current design of the computation processor requires the user/admin to manually curate and distribute computations
between instances. While this works for our current and some previous use cases it doesn't scale very well.

CWMS for instance will be focusing on the use of docker containers to run everything. These can be sized to minimize cost but scale up under conditions to allow more memory/cpu for spurts of getting things done. Additionally our current cloud platform doesn't led itself to easily creating new instances of comproc with a different "app."

**Describe the solution you'd like**

Instead of sequential going through a set of data compproc should distribute computations to a set of threads, likely using an `ExecutorService`/`Executor` and [CompletableFutures](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.htm). 

While certainly adding complexity it would also allow us to put that complexity in the right spot simplifying the code flow.

e.g. something like
```java
Executor executor = ...;

...

DataCollection dc = getNewData();
List<Computations> comps = getCompsFromDc(dc);
comps.forEach( c- >{
     CompletableFuture.runSync(RunnableFor(c),executor)
       .thenAsync(saveData,executor) // could be same or different executor
      .whenComplete(updateTasklist,executor);
     // there are options for cancelled, execptions, etc
});
```

Users/Admins would be able to configure the level of parallelism by specifying how many threads can be run. It would be a smidge more logic but could also handle a case of wanting a separate executor for saving. Or options for priority. E.g. a given computation (or maybe timeseries) could be run on a higher priority ExecutorService that creates threads with a higher priority.

There is also a ScheduledExecutor service that could help reduce the need for custom time for next run code; which might be somewhat easier to insert now but see additional context.

**Describe alternatives you've considered**
We could just keep what we have, but again, it doesn't easily, and especially automatically, scale.

**Additional context**

I thought of a rough idea on how to do this and wanted to write it down. This requires the Connections to all be in a pool to work due to java.sql.Connection not being practical to share between Threads.

So this will only work for databases that do connection pooling. Currently this is CWMS, but we will be doing so for all of them. Since this can and should be a complete separate "App" I may get started and just mark it as experimental.

PR #1877, opened 2026-04-13
Refactor/comp operations

MikeNeilson:
## Problem Description

Given that the saving of data happens after all computations are run, it's sometimes difficult to correlate log messages of data saved to which computation output the data.

Additionally we know we want to get to ThreadComproc (#291) eventually.

## Solution

Introduce an ExecutorService to allow for use of `CompleteableFuture`, which itself allows chaining operations. ThreadPool size is 1, comp still executes on a separate thread, but for the execution of computations no parallel work occurs.

Alter `DbComputation::apply` and all required to return a `DataCollection` of *only* the output time series.

Update `ComputationApp` to move the Save Time Series from after all computations to a provided handler to be called after every individual computation.

A follow up will allow the save to be on a separate thread, There should be no concurrency issues with saving data for one computation while another is processing.

`TimeSeriesDb::reloadTsIdCache` instances were modified to synchronize the full operation of the block on the class of each implementation after entries of `ConcurrentModificationException` were discovered in the logs.

## how you tested the change

No change in behavior except to timing of operations and log output.
Review logs to verify expected output was still present.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
+++ b/java/opendcs/src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
@@ -1154,58 +1154,60 @@ public class CwmsTimeSeriesDAO extends DaoBase implements TimeSeriesDAI
 
         // Each TSID will need a site, so prefill the site cache to prevent
         // it from doing individual reads for each site.
-
-        String q = cwmsTsidQueryBase + " WHERE upper(a.DB_OFFICE_ID) = upper(?)";
-        int origFetchSize = getFetchSize();
-        try
+        synchronized(CwmsTimeSeriesDAO.class)
         {
-            int tsidFetchSize = DecodesSettings.instance().tsidFetchSize;
-            if (tsidFetchSize > 0)
+            String q = cwmsTsidQueryBase + " WHERE upper(a.DB_OFFICE_ID) = upper(?)";
+            int origFetchSize = getFetchSize();
+            try
             {
-                setFetchSize(tsidFetchSize);
+                int tsidFetchSize = DecodesSettings.instance().tsidFetchSize;
+                if (tsidFetchSize > 0)
+                {
+                    setFetchSize(tsidFetchSize);
+                }
+
+
+                List<TimeSeriesIdentifier> tsidList = getResultsIgnoringNull(q, rs ->
+                {
+                    CwmsTsId retVal = null;
+                    try
+                    {
+                        retVal = rs2TsId(rs, false);
+                    }
+                    catch (DbIoException | NoSuchObjectException ex)
+                    {
+                        log.atWarn()
+                        .setCause(ex)
+                        .log("Error creating Cwms TSID -- skipped.");
+                    }
+                    return retVal;
+                }, dbOfficeId);
+
+                synchronized(cache)
+                {
+                    cache.clear();
+                    for(TimeSeriesIdentifier tsid: tsidList)
+                    {
+                        cache.put(tsid);
+                    }
+                }
+
+                log.debug("After fill, cache has {} TSIDs.", cache.size());
             }
-
-
-            List<TimeSeriesIdentifier> tsidList = getResultsIgnoringNull(q, rs ->
+            catch (SQLException ex)
             {
-                CwmsTsId retVal = null;
-                try
-                {
-                    retVal = rs2TsId(rs, false);
-                }
-                catch (DbIoException | NoSuchObjectException ex)
-                {
-                    log.atWarn()
-                       .setCause(ex)
-                       .log("Error creating Cwms TSID -- skipped.");
-                }
-                return retVal;
-            }, dbOfficeId);
-
-            synchronized(cache)
-            {
-                cache.clear();
-                for(TimeSeriesIdentifier tsid: tsidList)
-                {
-                    cache.put(tsid);
-                }
+                final String msg = "Failed to reload CwmsTimeSeriesDb TsId Cache.";
+                log.atError()
+                .setCause(ex)
+                .log(msg);
+                throw new DbIoException(msg, ex);
             }
-
-            log.debug("After fill, cache has {} TSIDs.", cache.size());
+            finally
+            {
+                setFetchSize(origFetchSize);
+            }
+            lastCacheReload = System.currentTimeMillis();
         }
-        catch (SQLException ex)
-        {
-            final String msg = "Failed to reload CwmsTimeSeriesDb TsId Cache.";
-            log.atError()
-               .setCause(ex)
-               .log(msg);
-            throw new DbIoException(msg, ex);
-        }
-        finally
-        {
-            setFetchSize(origFetchSize);
-        }
-        lastCacheReload = System.currentTimeMillis();
     }
 
     @Override
--- a/java/opendcs/src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
+++ b/java/opendcs/src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
@@ -1,16 +1,16 @@
 /*
 * Where Applicable, Copyright 2025 OpenDCS Consortium and/or its contributors
-* 
+*
 * Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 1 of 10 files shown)


java/opendcs/src/main/java/decodes/tsdb/ComputationExecution.java
```diff
@@ -26,16 +31,39 @@
...
 	public ComputationExecution(OpenDcsDatabase db)
+	{
+		this(db, Executors.newSingleThreadExecutor());
+	}
+
+
+	/**
+	 * Create instance of ComputationExecution using provided ExecutorService.
+	 *
+	 * NOTE: at this time we do not think the operations of DbComputation.apply are thread save.
```
ktarbet:
```suggestion
	 * NOTE: at this time we do not think the operations of DbComputation.apply are thread safe.
```

java/opendcs/src/main/java/decodes/tsdb/ComputationExecution.java
```diff
@@ -26,16 +31,39 @@
...
 	private int computesTried = 0;
+	private final ExecutorService executor;
 
+	/**
+	 * Create instance of ComputationExecution using a new SingleThreadExeuctor
+	 * @param db
+	 */
 	public ComputationExecution(OpenDcsDatabase db)
+	{
+		this(db, Executors.newSingleThreadExecutor());
```
ktarbet, line 55: What will close/shutdown this thread?

java/opendcs/src/main/java/decodes/tsdb/ComputationExecution.java
```diff
@@ -54,54 +82,152 @@ private TimeSeriesDb getTsDb()
...
-				listener.onProgress(String.format("Unexpected error in computation %s", comp.getName()), Level.WARN, ex);
-				numErrors++;
-				for(Integer rn : comp.getTriggeringRecNums())
+					DataCollection ret = new DataCollection();
+					try (var mdcComputation = MDC.putCloseable(COMPUTATION_KEY, comp.getName());
+						var compTimer = MDCTimer.startTimer(comp.getName()))
+					{
+						listener.onProgress(String.format("Executing computation '%s' #trigs=%d",
+								comp.getName(), comp.getTriggeringRecNums().size()), Level.DEBUG, null);
+						computesTried++;
```
ktarbet: Should this regular int (computesTried) be Atomic for safety?

java/opendcs/src/main/java/opendcs/opentsdb/OpenTimeSeriesDAO.java
```diff
@@ -34,6 +34,7 @@
 import org.opendcs.database.ExceptionHelpers;
 import org.opendcs.utils.FailableResult;
 import org.opendcs.utils.logging.OpenDcsLoggerFactory;
+import org.python.modules.synchronize;
```
ktarbet:
```suggestion
```

java/opendcs/src/main/java/org/opendcs/utils/logging/ThreadUtils.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public final class ThreadUtils
+{
+    private ThreadUtils()
+    {
+        /* utility class */
+    }
+
+    /**
+	 * Propgate the MDC context into the given task for things like the Trace ID.
```
ktarbet:
```suggestion
	 * Propagate the MDC context into the given task for things like the Trace ID.
```

java/opendcs/src/main/java/org/opendcs/utils/logging/ThreadUtils.java
```diff
@@ -0,0 +1,39 @@
...
+	 * @param <T>
+	 * @param task
+	 * @return
+	 */
+	public static <T> Supplier<T> propegate(Supplier<T> task)
```
ktarbet:
```suggestion
	public static <T> Supplier<T> propagate(Supplier<T> task)
```

java/opendcs/src/main/java/decodes/tsdb/ComputationExecution.java
MikeNeilson: When the ComputationExecutor is no longer referenced and gets garbage collected.

java/opendcs/src/main/java/decodes/tsdb/ComputationExecution.java
MikeNeilson: Yeah, I just wasn't super worried about it for this, but thanks for the reminder.

java/opendcs/src/main/java/opendcs/opentsdb/OpenTimeSeriesDAO.java
MikeNeilson: Good old code completion trying to be helpful.

MikeNeilson (commented):
thanks forgot to go back and to the atomics

ktarbet (dismissed):

java/opendcs/src/main/java/decodes/tsdb/ComputationApp.java
```diff
@@ -372,26 +372,35 @@ else if (regressionTestModeArg.getValue()
...
 
-						ComputationExecution.CompResults results = execution.execute(List.of(comps), dataCollection);
-						compsTried += results.computesTried();
-						compErrors += results.numErrors();
-
-						action = "Saving results";
-						List<CTimeSeries> tsList = dataCollection.getAllTimeSeries();
-						log.trace("{} {} time series in data.", action, tsList.size());
-						try (var allTsTimer = MDCTimer.startTimer(action))
+						ComputationExecution.CompResults results = execution.execute(List.of(comps), dataCollection,
```
oskarhurst, line 375: Would this work If the Executor you pass in was a multithreaded executer. or would is race when getting all timer series.

java/opendcs/src/main/java/decodes/tsdb/ComputationExecution.java
```diff
@@ -54,54 +83,152 @@ private TimeSeriesDb getTsDb()
...
-			comp.getTriggeringRecNums().clear();
-			listener.onProgress(String.format("End of computation '%s'", comp.getName()), Level.DEBUG, null);
+					listener.logEvent("will save " + dc.size() + " time series from comp", Level.INFO, null);
+					return dc;
+				})
+				.thenApply(afterComp);
+			comps.add(future);
 		}
-		return new CompResults(numErrors, computesTried);
+		CompletableFuture.allOf(comps.toArray(new CompletableFuture[0])).join();
```
oskarhurst: Do we need to catch the CompletionExecption that could be thrown by the allof().Join() here.

java/opendcs/src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
```diff
@@ -1154,58 +1154,60 @@ public void reloadTsIdCache() throws DbIoException
...
+                    catch (DbIoException | NoSuchObjectException ex)
+                    {
+                        log.atWarn()
+                        .setCause(ex)
+                        .log("Error creating Cwms TSID -- skipped.");
+                    }
+                    return retVal;
+                }, dbOfficeId);
+
+                synchronized(cache)
```
oskarhurst: I see this is a nested Synchronized call in a syncronized block. Wont this potentally cause a dead lock?

MikeNeilson: Possibly, will remove.

java/opendcs/src/main/java/decodes/tsdb/ComputationApp.java
MikeNeilson: I *could* work, but likely won't. The default is single thread, the option to add a different executor is to allow us to experiment.

java/opendcs/src/main/java/decodes/tsdb/ComputationExecution.java
MikeNeilson: Did not think of that, good idea.

oskarhurst (approved):

PR #1884, opened 2026-04-17
Fix computation process/group selection

tlelv:
Normalized OpenDCS null-key IDs, so -1's are treated as unset and unmatched values are cleared. Added regression tests for real API data

## Problem Description

Fixes #. 
The computations editor worked against Storybook-style data but was failing against real OpenDCS API payloads. Some computations came back with null-key IDs like appId = -1 and groupId = -1. The UI treated those IDs as valid.

## Solution

Describe your solution.

Updated the computations editor to normalize process/group values before saving.

Made it so only IDs greater than zero are treated as valid.
When the API returns -1, the editor now falls back to matching by applicationName and groupName, and if no valid match exists, the editor clears the process/group fields.

## how you tested the change

Manual verification by launching the website. Previously I had mistakenly only used Storybooks. 
Screenshot 2026-04-17 at 11 10 04 AM

## Where the following done:


MikeNeilson (dismissed):

javascript/opendcs-web-ui-react/src/pages/computations/computations/index.tsx
```diff
@@ -85,25 +85,67 @@ export const Computations: React.FC = () => {
...
-        .catch((e: unknown) => console.error("Failed to save computation", e));
+      try {
+        const saved = await computationApi.postComputation(api.org, {
+          ...computation,
+          computationId,
+        });
+        if (saved.computationId && saved.computationId > 0) {
+          setComputations((current) => {
+            const savedRef: ApiComputationRef = {
+              computationId: saved.computationId,
```
MikeNeilson: shouldn't this map with the spread operator? e.g.

savedRef = {
 ...saved
}
```

MikeNeilson (commented):
Overall like the change, definitely need to chew on it more than the 5 minutes I just spent. One comment that may just be me not seeing something.

tlelv:
Screenshot 2026-04-23 at 12 03 08 PM
Screenshot 2026-04-23 at 12 04 52 PM

The current state. It selects algorithms well and populates. The saving issue was fixed and i reformatted the [+] Button to be below. I'm working for the time series to function right

MikeNeilson:
For the computation run functionality (which FYI great, but you might want to stick and do in a follow up PR just to keep things simpler), I would have it replace the child row card entirely. E.G. hide the computation config and provide a big back arrow at the top. That way you get more screen space to render and the user doesn't have to scroll up or down.

PR #1888, opened 2026-04-20
Expandale data tables

### LINKED ISSUE #1761: [React UI] Organize code for DataTables that handle "child rows" to reduce duplication.
Components that use DataTables and render elements in a "child row"

**Purpose of refactoring**
Currently only the in-progress SitesTable implements this behavior, there are several behaviors that will be shared across all.

**Proposed method**
After the creation of the next page that makes use of DataTables with a child row we should review the design and determine if:

1. It makes sense to create a more refined "DataTable" React component that includes that behavior (but allows additional behavior at each point).
2. Generic Methods can be extracted and shared between usages.
3. It makes sense to keep them all separate.

**Additional context**
@wjonassen had done this within the existing JSP based Web UI. However, there all DataTables usages were roughly the same. For the React UI not all tables would use child rows or all types of behavior. That said I'm already seeing some common patterns looking at the new SiteNameList and PropertiesTable implementation.

oskarhurst:
## Problem Description

Fixes #1761

## Solution

Created a Wrapper components around the Datatables to handle shared and boiler plat functionality

## how you tested the change

## Where the following done:


MikeNeilson (dismissed):
I like it.

Though I think I agree, the inline and child row, at least at the usage level, should be separate components.

MikeNeilson (dismissed):

javascript/opendcs-web-ui-react/src/components/data-table/AppDataTable.tsx
```diff
@@ -690,7 +690,7 @@ export function AppDataTable<T, TId extends string | number, TSave = T>(
       // DataTables binds `this` to the Api on each callback — read it once
       // and hand everything else off to plain helpers so this component body
       // stays free of `this`-using flows.
-      const dt = this.api();
+      const dt = this.api(); // NOSONAR — DataTables binds `this` to the Api instance in drawCallback
```
MikeNeilson, line 693: Thanks, that also kept driving me up the wall.

MikeNeilson (approved):

PR #1898, opened 2026-04-27
Devops/docker build improvements

MikeNeilson:
## Problem Description

Nightly build from Main was failing.
So far as we could tell it was an issue with insufficient outbound ports from the runner.

## Solution

- Modify release action to user docker bake tool to better organize the docker build.
- Extract the build artifacts from the docker builder image so they aren't rebuild locally
- Rearrange the release yaml to deal with the new flow.

## how you tested the change

Tested locally with the nektos/act tool up to the point of singing. method of signing was not changed, just the building of the artifacts.

## Where the following done:


docker-bake.hcl
ktarbet, line 1:
```
  dockefile = "Dockerfile"
```
should be:
```
  dockerfile = "Dockerfile"
```

in a few places

MikeNeilson: huh, honestly surprised it worked without issue. thanks for catching that.

ktarbet: It must  default to 'Dockerfile'.

ktarbet (approved):

PR #1905, opened 2026-04-30
Reorganize DAOs into own subprojects.

### LINKED ISSUE #1886: Reorganize new DAOs into subprojects
All new Stateless DAO implementations

**Purpose of refactoring**
Making small changes to a DAO causes the build system to decide it needs to run the base tests again. Since most, if not all, of the changes need to be checked against a running database this just adds extra time.

Additionally, it will provide a technical enforcement to keep us from falling into the previous trap of mixing systems.

**Proposed method**
Have DAOs be in isolated projects so that the build changes are isolated.

**Additional context**
There may be some value in lifting more of the test fixtures into the integration testing fixtures project so that the DAO tests can be in the implementation specific projects.

That said, there's also some value in having one DAO test that all implementations have to pass regardless of their internals.

MikeNeilson:
## Problem Description

Fixes #1886. 

## Solution

- Pull Database/DAO implementation details, as much as possible, into isolated subprojects.
- As "OpenDCS" stands as the reference implementation the CWMS-Implementation does depend on it for some elements.
- Pulled the previously EnumDao into it's own class, due to the original way I did that the old does depend on the new. Adjustments are... awkward, but work.
- Much of the JDBI elements remained in place. Some of that is fairly baseline work though I do think it shouldn't be in the core "opendcs" directory. Unfortunately some of the required interfaces/implementations are in that project and it will take some time to extract them.

## how you tested the change

Existing integration tests. No changes in behavior, just reorganization.

## Where the following done:


java/opendcs-implementation/src/main/java/org/opendcs/database/impl/opendcs/dao/EnumDaoImpl.java
```diff
@@ -0,0 +1,192 @@
...
+    public Optional<DbEnum> getEnum(DataTransaction tx, String enumName) throws OpenDcsDataException
+    {
+        var handle = tx.connection(Handle.class)
+                       .orElseThrow(() -> new OpenDcsDataException(SqlErrorMessages.NO_JDBI_HANDLE));
+        try (var query = handle.createQuery("select id from enum where upper(name) = upper(:name)"))
+        {
+            var id = query.bind(GenericColumns.NAME, enumName)
+                          .mapTo(DbKey.class)
+                          .findOne();
+            return getEnum(tx, id.orElse(DbKey.NullKey));
```
ktarbet, line 83: Should this be Optional.empty() instead of DbKey.NullKey

java/opendcs/src/main/java/opendcs/dao/EnumSqlDao.java
```diff
@@ -491,7 +483,7 @@ public void writeEnum(DbEnum dbenum) throws DbIoException
     {
         try (DataTransaction tx = this.getTransaction())
         {
-            var written = this.writeEnum(tx, dbenum);
+            var written = newDao.writeEnum(tx, dbenum);
```
ktarbet, line 486: what about adding or clearing the cache for this enum.

ktarbet (dismissed):

java/opendcs-implementation/src/main/java/org/opendcs/database/impl/opendcs/dao/EnumDaoImpl.java
MikeNeilson: it could be, but looking up by -1 will also return empty. Based on previous lessons this can be simplified but I didn't want to do that work now so I could focus on just the file/project organization.

java/opendcs/src/main/java/opendcs/dao/EnumSqlDao.java
MikeNeilson: Hmm, good point, I believe I just deleted it from the new DAO but it should remain here.

ktarbet (approved):

PR #1907, opened 2026-04-30
Refresh DECODES caches during routing

### LINKED ISSUE #1463: Routing scheduler needs restarted when a DECODES change is made
**Describe the bug**
For at least some DECODES changes, the routing scheduler doesn't pick up the change until it is restarted.  We were able to confirm this is the case by changing a parameter name from "Stage" to "Stage-Orifice" and adding a new sensor (and thus time series).  Neither change was seen until the routing scheduler was restarted 9 hours later.

**To Reproduce**
1. DECODES
2. Open a config
3. Change a parameter name from Stage to Stage-Orifice
4. See that the change is not reflected in CWMSVue until after a routing scheduler restart

**Expected behavior**
Expected the changes to be picked up right away.

 - Client side 7.0.15 on Windows 11
 - Server side 7.0.16-RC06

krowvin:
## Problem Description

Fixes #1463.

As discussed in the OpenDCS call AI Tools were used to complete this PR

## Solution

Updated the routing scheduler refresh path so DECODES edits are picked up without restarting the scheduler.

The periodic platform-list refresh now reloads platform configs before refreshing platforms. SQL and CWMS platform-list readers now update existing cached platform entries instead of skipping them, including transport media and nullable platform fields. Config reloads also clear stale sensors and decoding scripts before loading the fresh definition.

## how you tested the change

Ran:

`.\gradlew.bat :opendcs:test '-Pno.docs=true'`

Result: build successful. Also added a unit test for clearing a cached platform config decoding definition before reload.

## Where the following done:


java/opendcs/src/main/java/decodes/db/PlatformConfig.java
```diff
@@ -495,6 +495,12 @@ public void removeAllSensors()
 		configSensors.removeAllElements();
 	}
 
+	public void clearDecodingDefinition()
```
MikeNeilson: This seems excessive.

java/opendcs/src/test/java/decodes/db/PlatformConfigTest.java
MikeNeilson, line 1: Certainly tests the function that was created, but doesn't test the underlying issue which is "if I reload the platforms am I also getting any updated configs"

The test definitely needs to be an integration test for that level of check.

MikeNeilson (changes requested):

MikeNeilson (changes requested):
The changes to PlatformCOnfig and RoutingSpecThread a good.

The test needlessly duplicates things though. If you move that test, basically as-is, to the integration tests you don't need to duplicate the schema creation.

The actual contents of the test is valid, would just be easy to maintain as an integration test (database is already setup). NOTE: will need the `@EnableIfTsdb` annotation.

integrationtesting/opendcs-tests/src/test/java/decodes/sql/PlatformListIOTestIT.java
```diff
@@ -0,0 +1,192 @@
...
+		throws Exception
+	{
+		oldDb = Database.getDb();
+		Database.setDb(decodesDb);
+		decodesDb.platformList.clear();
+		decodesDb.platformConfigList.clear();
+
+		try (DataTransaction tx = openDcsDatabase.newTransaction())
+		{
+			Connection conn = tx.connection(Connection.class).orElseThrow();
```
MikeNeilson, line 51: while valid to yank the connection out, and it is needed for the key generator, down in insertPlatformWithConfig it should use a Jdbi3 Handle ( you pass `Handle.class` into `tx.connection`) and use the JDBI 3 binding. Wwe don't have a new DAO for platforms yet, so doing the raw sql is fine here, but wrapping with JDBI 3 will get us better error messages.

MikeNeilson: it'll also handle the date conversion, that's causing an issue below, correctly.

integrationtesting/opendcs-tests/src/test/java/decodes/sql/PlatformListIOTestIT.java
```diff
@@ -0,0 +1,192 @@
...
+				+ "(ID, Agency, IsProduction, SiteId, ConfigId, Description, LastModifyTime, Expiration, "
+				+ "PlatformDesignator) "
+				+ "VALUES (?, ?, ?, NULL, ?, ?, ?, NULL, NULL)"))
+		{
+			stmt.setLong(1, platformId.getValue());
+			stmt.setString(2, "TEST");
+			stmt.setString(3, "true");
+			stmt.setLong(4, configId.getValue());
+			stmt.setString(5, "platform list io test platform");
+			stmt.setTimestamp(6, Timestamp.from(Instant.now()));
```
MikeNeilson: so, in OpenDCS-Postgres or -Oracle this is an integer, in CWMS-Oracle it's a Date column.

Which is not logic to handle here. Use JDBI3 for the main branch, we'll switch the query to using our custom handlers in the 7.0 branch. In either case it handles the date conversion correctly by implementation so the point of usage, e.g. here, doesn't have to care.

MikeNeilson (changes requested):

MikeNeilson:
@krowvin looks like some sql got mixed up.

MikeNeilson (approved):

PR #1908, opened 2026-05-04
AI Assisted - PythonAlgorithm cwms rating performance improvement

msweier:
Ok so I misremembered....it was a bit more than a couple line fix, but the problem probably deserves a good look. I probably need to convert some of my algos to java anyway though. 

This PR was assisted w/ AI.

## Problem Description

`PythonAlgorithm` was acquiring a new database connection from the pool for every `rating()` call during time slice processing. For algorithms processing many time slices, this caused significant overhead — each `rating()` invocation would trigger a full connection acquisition and return, leading to extra comp time of of unnecessary pool contention per algorithm execution.

**Test results**: A 1-week computation for two gates improved from 52 seconds to 36 seconds (~31% improvement) using the windows GUI. Tested a non-rating python algo with windows GUI and it worked as expected.

## Solution

Reuse a single database connection throughout the algorithm execution:

1. Acquire the connection once in `beforeTimeSlices()` and fail fast if unavailable
2. Pass it to all `rating()` calls during time slice processing via a new `CwmsTimeSeriesDb.rating()` overload
3. Release it in `alwaysAfterTimeSlices()` to guarantee cleanup even on exception

### Files Modified

**PythonAlgorithm.java**
- Added `private Connection algoConnection` field
- `beforeTimeSlices()`: acquire connection once, throw `DbCompException` on failure
- `rating()`: use `algoConnection` when available (only for CWMS databases via `instanceof` guard)
- `alwaysAfterTimeSlices()`: close connection in finally block

**CwmsTimeSeriesDb.java**
- Added `rating(String, Date, Connection, double...)` overload that accepts an existing connection

### Design Notes

- **CWMS-only**: Connection is only acquired for `CwmsTimeSeriesDb` instances. HDB databases and algorithms that never call `rating()` are unaffected.
- **Fail fast**: If the pool is unavailable at start, the exception propagates immediately — no silent fallback to per-call connections.
- **Backward compatible**: Original `rating()` method unchanged; non-CWMS paths fall back to the original implementation.
- **Thread safe**: Each algorithm instance has an isolated connection field.


java/opendcs/src/main/java/decodes/tsdb/algo/PythonAlgorithm.java
```diff
@@ -248,6 +249,19 @@ protected void beforeTimeSlices()
 			}
 		}
 
+		if (tsdb instanceof CwmsTimeSeriesDb)
```
MikeNeilson: skip the database type check, this will benefit all the implementations.

java/opendcs/src/main/java/decodes/tsdb/algo/PythonAlgorithm.java
```diff
@@ -1250,7 +1285,10 @@ public double rating(String specId, double... indeps)
 
 		try
 		{
-			return tsdb.rating(specId, _timeSliceBaseTime, indeps);
+			if (algoConnection != null && tsdb instanceof CwmsTimeSeriesDb)
```
MikeNeilson: Same as above, skip the database type check.

MikeNeilson (changes requested):
The database type check isn't required, that will be useful for any implementation.

That said, move the connection retrieval into "alwaysBefore", the counterpart to alwaysAfter.

If you can work out an integration test, great, if not let us know and we'll see about creating one to merge into your PR.

MikeNeilson:
do us a favor, within the PythonAlgorithm class, beyond the changes requested. Go find the getConnection method and have it return

```
   return new WrappedConnection(algoConnection);
```

There are other algorithms out there that also retrieve a connection this will make them more performant and cause less other problems beyond just the use of rating.

java/opendcs/src/main/java/decodes/tsdb/algo/PythonAlgorithm.java
msweier: Done — removed the instanceof CwmsTimeSeriesDb guard.

java/opendcs/src/main/java/decodes/tsdb/algo/PythonAlgorithm.java
msweier: Done — removed the cast and instance of check.

msweier:
> do us a favor, within the PythonAlgorithm class, beyond the changes requested. Go find the getConnection method and have it return
> 
> ```
>    return new WrappedConnection(algoConnection);
> ```

Sure, I didn't notice any performance improvement from this, but makes sense to reuse a connection if you can.

java/opendcs/src/main/java/decodes/tsdb/algo/PythonAlgorithm.java
```diff
@@ -1341,6 +1377,8 @@ public double tabrating(String tabfile, double indep)
 	}
 
 	public Connection getConnection(){
+		if (algoConnection != null)
```
MikeNeilson, line 1381:
```suggestion
		if (algoConnection != null)
		{
 		    return new WrappedConnection(algoConnection);
 		}
 		else
 		{
 		    return tsdb.getConnection()
 		}
```

MikeNeilson: Note, careful excepting that change in the Github UI, not sure it's picking up the multiline suggestions correctly.

MikeNeilson (changes requested):
Overall looks good, the change to TimeSeriesDb::rating definitely an important one. We've had to give a few other elements that treatment as well.

A minor nitpick on formatting in the getConnection, but once that's fixed and we've made some tests (I've been wanting a PythonAlgo test for a while) we can merge in.


I'll work on the test itself, however if you are willing to provide the python algo text of your algorithm and some range of input/output data, that would be the easiest and best (you know your covered).

This PR at present is in a good state so back it up first, but a test shouldn't be too difficult for the AI tool to create given the existing examples... granting it might get confused on the XML formatting for the python algo code itself. Perhaps you could run compexport and say "use algo XYZ and a starting template".... I don't use the AI tools, I have no idea if that's reasonable or not, but I'd probably just do the compexport myself. Let us know if you want to try that with the AI tool, or if you'd rather pass the test work to us - just at least provide some good sample data/config.

java/opendcs/src/main/java/decodes/tsdb/algo/PythonAlgorithm.java
msweier: I'll give the test a shot and let you know if I get stumped. I have a reservoir inflow algo that follows the example in the read the docs fairly closely that uses a rating lookup for storage.

msweier: Added python inflow integration test based on Res Evap files.

MikeNeilson (approved):

MikeNeilson:
Good work with the test.

MikeNeilson:
@Oskar Sonar analysis failure I suspect is due to some differences in the 7.0 branch target, likely too old of sonar cloud plug that's still triggering compile, and the fact that the react-ui doesn't exist in that branch.


Going to ignore for this, visually it all looks fine, but we should definitely get the figured out at some point.

PR #1909, opened 2026-05-04
PlatformConfig/Decodes Script DAO

### LINKED ISSUE #1225: PlatformConfig DAO

None

**Additional context**
NOTE: May be part of Platforms, may make sense to have a separate decoding script DAO. Decision on such should be made after additional investigation.

MikeNeilson:
## Problem Description

Fixes #1225. 

## Solution

Create DAO in the new stateless style

Opening as draft is this is a bit... nuts. But the alternative is a bunch of totally separate DAOs and we'd have to do the join in memory.

Given the overall complexity, I think for this one I'll make some extra methods for the REST API "ref" retrievals. Don't really need to do the full join for those. (previously haven't worried about it because they were small but there's like 2-3 more join tables to do on this one.)

NOTE: This is, currently, intentionally duplicating a rather large about of column names in the SQL. I don't think that's unnecessary duplication, but I'm considering better ways to handle mapping out the columns, most likely from the DAOs, but an additional method on the PrefixRowMapper may be better, basically a "retrieve list of required columns with prefix." The mappers are generally what would change with newer versions anyways.... but suggestions welcome. This is already large enough I won't do it in this PR, but I think before the next one of these.

## how you tested the change

- Manually testing query

Otherwise todo.

## Where the following done:

### DIFF
--- a/integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/DecodesConfigDaoTestIT.java
+++ b/integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/DecodesConfigDaoTestIT.java
@@ -11,7 +11,9 @@ import java.util.ArrayList;
 import java.util.Random;
 
 import org.junit.jupiter.api.Test;
+import org.opendcs.database.api.DataTransaction;
 import org.opendcs.database.api.OpenDcsDatabase;
+import org.opendcs.database.dai.DataTypeDao;
 import org.opendcs.database.dai.DecodesConfigDao;
 import org.opendcs.fixtures.AppTestBase;
 import org.opendcs.fixtures.annotations.ConfiguredField;
@@ -20,7 +22,6 @@ import org.opendcs.fixtures.annotations.EnableIfTsDb;
 
 import decodes.db.ConfigSensor;
 import decodes.db.Constants;
-import decodes.db.DataType;
 import decodes.db.DecodesScript;
 import decodes.db.FormatStatement;
 import decodes.db.PlatformConfig;
@@ -35,6 +36,10 @@ import decodes.db.UnitConverterDb;
 })
 class DecodesConfigDaoTestIT extends AppTestBase
 {
+
+    private static final String[] SENSORS = new String[]{"Stage", "Elev", "Flow"};
+    private static final String[] UNTIS = new String[]{"in", "ft", "cfs"};
+
     @ConfiguredField
     OpenDcsDatabase db;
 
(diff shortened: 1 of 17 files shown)


integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/DecodesConfigDaoTestIT.java
ktarbet, line 1: consider naming this class DecodesConfigDaoTestIT

java/opendcs-implementation/src/main/java/org/opendcs/models/PlatformConfigBuilder.java
```diff
@@ -0,0 +1,131 @@
...
+        if (cs != null)
+        {
+         cs.addDataType(dataType);
+        }
+        return this;
+    }
+
+    public PlatformConfigBuilder withEquipmentModel(EquipmentModel equipmentModel)
+    {
+        if (this.equipmentModel != null)
```
ktarbet, line 99:
```suggestion
        if (equipmentModel != null)
```

ktarbet (commented):
checkpoint

java/opendcs-implementation/src/main/java/org/opendcs/models/PlatformConfigBuilder.java
MikeNeilson: I should put a comment here, this actually is intentionally checking if it's already set. The Joins end up making a lot of duplicate columns so this is done this way so that the flow in the Accumulator can be more direct.

MikeNeilson:
going to do rest api usage in a follow up, this one already seems a bit big.

ktarbet (commented):
I'm having  issues with github inline comments; 
manually putting suggestions here:


DecodesConfigDaoImpl.java:114: 

private static final String DELETE_UNITCONVERETER = "delete from unitconverter where id = :id";

spelling: DELETE_UNITCONVERTER


DecodesConfigDaoImpl.java:418

var deleteScript = handle.createUpdate(DELETE_FORMATSTATEMENTS);   // should be DELETE_DECODESSCRIPT


DecodesConfigDaoTestIT.java:41: UNTIS (spelling)


java/opendcs-implementation/src/main/java/org/opendcs/database/impl/opendcs/jdbi/mapper/decodes/configs/DecodesConfigAccumulator.java


spelling: change volumous to  voluminous

ktarbet:
integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/DecodesConfigDaoTestIT.java:186

Sensor is not added to pc, with pc.add(...)


line 213:  

 final var sensorName = pc.getSensor(k + 1);
 
should be
 final var sensor = pc.getSensor(k + 1);
 
 and
 if (SENSORS[sensorIdx].equals(sensor.getName??())

MikeNeilson:
Will also add some additional test assertions to make sure some of those sensor things are actually caught.

ktarbet (approved):

PR #1919, opened 2026-05-11
Add tasklistDebounceSeconds to fix CompProc partial-input race

Enovotny:
## Problem Description

Downstream comps that depend on outputs of upstream comps occasionally
ran with only a partial input set when all upstream tasklist rows
landed in a single ComputationApp poll. The comp completed with 1
(or 0) values instead of 24, and because no exception was thrown the
existing retryFailedComputations / FAIL_TIME path didn't fire. Result:
recurring 24-hour data gaps in batch outputs (observed on Barren
Flow-Outflow.Inst.1Hour and Flow-Inflow.Ave.1Hour timeseries at LRL).

The trigger pathway is CP_COMP_TASKLIST polled by getNewData()

Adds a tasklistDebounceSeconds DecodesSettings (default 0 = no
behavior change) that filters CP_COMP_TASKLIST.DATE_TIME_LOADED so
only entries older than the configured window are returned. Skipped
entries remain in the queue and are picked up on the next poll. This
gives cascaded writes within a comp cycle time to settle before the
dependent comp resolves.

Applied to OpenTimeSeriesDAO (Oracle + Postgres dialects),
CwmsTimeSeriesDAO, and HdbTimeSeriesDAO. ComputationApp logs the
active value at startup. TasklistDebounceTestIT added (currently
@Disabled, mirroring CompProcTestIT.test_bad_recs_cleared until the
cp_comp_tasklist sourceid constraint is addressed for cross-DB tests).

Recommended setting for production: tasklistDebounceSeconds=5

created using Claude Opus 4.7


java/opendcs/src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
```diff
@@ -1325,13 +1325,19 @@ public DataCollection getNewData(DbKey applicationId, int maxTake)
                     ? " and (a.FAIL_TIME is null OR SYSDATE - a.FAIL_TIME >= 1/24)"
                     : "";
 
+        String debounceClause = "";
+        int debounceSec = DecodesSettings.instance().tasklistDebounceSeconds;
+        if (debounceSec > 0)
+            debounceClause = " and a.DATE_TIME_LOADED <= SYSDATE - " + debounceSec + "/86400";
```
MikeNeilson: This needs to be a bind variable. For the other implementations don't worry about it, they haven't gotten the proper prepared statement treatment yet.

integrationtesting/opendcs-tests/src/test/java/org/opendcs/regression_tests/TasklistDebounceTestIT.java
```diff
@@ -0,0 +1,115 @@
...
+ * from returning entries that were just inserted by the same compproc cycle.
+ * See DecodesSettings.tasklistDebounceSeconds and the getNewData() implementations
+ * in OpenTimeSeriesDAO, CwmsTimeSeriesDAO, HdbTimeSeriesDAO.
+ */
+@DecodesConfigurationRequired({
+    "shared/test-sites.xml",
+    "shared/presgrp-regtest.xml"
+})
+@ComputationConfigurationRequired({"shared/loading-apps.xml"})
+@Disabled("Mirrors CompProcTestIT.test_bad_recs_cleared — disabled until "
```
MikeNeilson: This need to be enabled for at least one implementation to demonstrate behavior directly. You can use `@EnableIfTsDb("OpenDCS-Postgres")` to limit to the OpenDCS-Postgres implementation, for example.

MikeNeilson (changes requested):
Based on the provided description I do not believe this will actually correct the given issue as there appears to be some misconceptions.

1. Only already pulled tasklist records are then deleted from the tasklist after computations.
2. After determining which computations need to be run, required data is pulled from the database (e.g. only one input is required to trigger the full computation will all of the data.)
3. Data is written after a computations runs, but at this time no new tasklist records are retrieved until after all computations of the batch are run.

If, and big iff there, I'm reading the description it's claiming that entries are being retrieved and removed in the middle of a current batch of computations.


That said, I don't believe this will hurt anything and will at least isolate one possiblity without breaking the system. Once the requested changes are addressed we can merge in.

Additionally, while we do not have our formal AI policy ready, if AI was used, please indicate the agent and version with an assist-by clause in either the commits or the PR description as that will be a part of the policy.

java/opendcs/src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
Enovotny: change made

integrationtesting/opendcs-tests/src/test/java/org/opendcs/regression_tests/TasklistDebounceTestIT.java
Enovotny: change made

MikeNeilson (dismissed):

Enovotny:
@MikeNeilson I think I got the tests to pass now. also add a CWMS database test.

MikeNeilson (dismissed):

MikeNeilson (commented):
Once @Enovotny adds self to contributors.md will merge in.

PR #1921, opened 2026-05-11
Update TomcatServer to use seed data

oskarhurst:
## Problem Description

Manual testing is very hard without real data to test against.

## Solution

Add an environment variable to tell the tomcat server where to find seed data for it to load.

## Where the following done:


oskarhurst:
@MikeNeilson I also updated the sonar.yaml so that we can manually trigger sonarqube if it ever get stuck so we dont always have to re run the tests

.github/workflows/sonar.yaml
```diff
@@ -3,25 +3,76 @@ on:
...
     steps:
+      - name: Resolve source build context
+        id: ctx
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+        run: |
+          set -euo pipefail
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            run_id="${{ inputs.build_run_id }}"
```
MikeNeilson: move the run ID up into the `env:` block, matter of how the input is processed.

.github/workflows/sonar.yaml
```diff
@@ -3,25 +3,76 @@ on:
...
+            base_owner=$(jq -r '.repository.owner.login' <<<"$run_json")
+            event=$(jq -r '.event' <<<"$run_json")
+            conclusion=$(jq -r '.conclusion' <<<"$run_json")
+            if [[ "$conclusion" != "success" ]]; then
+              echo "::warning::Build run $run_id concluded as '$conclusion' (not 'success'). Continuing because this is a manual dispatch."
+            fi
+          else
+            run_id="${{ github.event.workflow_run.id }}"
+            head_sha="${{ github.event.workflow_run.head_sha }}"
+            head_branch="${{ github.event.workflow_run.head_branch }}"
```
MikeNeilson: same as above.

integrationtesting/opendcs-tests/src/test/java/org/opendcs/fixtures/TomcatServer.java
```diff
@@ -319,6 +328,88 @@ private static void startDbContainer(Configuration config, String dbType) throws
...
+			DecodesSettings settings = DecodesSettings.fromProfile(srcProfile);
+			settings.DbAuthFile = "java-auth-source:password=DCS_PASS,username=DCS_USER";
+			settings.saveToProfile(tmpProfile);
+
+			if (!decodesFiles.isEmpty())
+			{
+				log.info("Loading {} seed DECODES file(s) from {}", decodesFiles.size(), decodesDir.getAbsolutePath());
+				List<String> fileNames = decodesFiles.stream()
+													  .map(File::getAbsolutePath)
+													  .collect(Collectors.toList());
```
MikeNeilson:
```suggestion
													  .toList();
```

integrationtesting/opendcs-tests/src/test/java/org/opendcs/fixtures/TomcatServer.java
```diff
@@ -319,6 +328,88 @@ private static void startDbContainer(Configuration config, String dbType) throws
...
+				new DbImport(tmpProfile, null, false, false, false, false, true,
+							 null, null, null, null, fileNames).importDatabase();
+			}
+
+			if (!compFiles.isEmpty())
+			{
+				log.info("Loading {} seed computation file(s) from {}", compFiles.size(), compsDir.getAbsolutePath());
+				List<String> fileNames = compFiles.stream()
+												  .map(File::getAbsolutePath)
+												  .collect(Collectors.toList());
```
MikeNeilson:
```suggestion
												  .toList();
```

MikeNeilson (changes requested):
Seems reasonable, needs a few tweaks for safety on the sonar.yaml file.

MikeNeilson:
I believe you pulled NWDM data for the test. In this or a separate PR we should include that data as a test set to use. coordinate with @jbatterman on anything that shouldn't be included (beyond passwords and LRGS targets... just change the LRGS to `lrgs.opendcs.org` and username/password as anonymous)

@jbatterman FYI: data not included yet, but your district is well, fairly comprehensive so good first sample set, just let us know if something shouldn't be included either because sensitive or outside users just couldn't get it anyways.

.github/workflows/build.yml
```diff
@@ -65,65 +65,160 @@ jobs:
...
+            task: testOpenDCSPostgres
+          - name: opendcs-oracle
+            task: testOpenDCSOracle
+          - name: cwms-oracle
+            task: testCWMSOracle
+    steps:
+      - name: Clean up disk space, so we don't run out.
         run: |
-          ./gradlew :testing:opendcs-tests:testOpenDCSOracle  --info
-      - name: integration tests (CWMS-Oracle)
```
MikeNeilson, line 93: The job "host" isn't actually shared, so this each of these are building the project again.

If they `need: tests` then tests could bundle up and output the build and we can extract into each environment, should speed things up a bit as well as harassing external artifact services, or the cache for the matter, a bit less.

MikeNeilson (changes requested):
I like it, the time to run has also been bugging me.

The current solution is duplicating compilation, but I don't think that's a difficult fix. 

Can either depend on the root build first (tests) or split into yet-another job that just does the compiling system to pass off to everything else.

MikeNeilson (dismissed):

MikeNeilson (approved):

jbatterman:
> I believe you pulled NWDM data for the test. In this or a separate PR we should include that data as a test set to use. coordinate with @jbatterman on anything that shouldn't be included (beyond passwords and LRGS targets... just change the LRGS to `lrgs.opendcs.org` and username/password as anonymous)
> 
> @jbatterman FYI: data not included yet, but your district is well, fairly comprehensive so good first sample set, just let us know if something shouldn't be included either because sensitive or outside users just couldn't get it anyways.

@MikeNeilson there is some data that we need to keep out of any tests.  Are you looking at our whole database?  I can send a separate email with the pathnames that should not be included.

MikeNeilson:
> there is some data that we need to keep out of any tests. Are you looking at our whole database? I can send a separate email with the pathnames that should not be included.

Yeah, that'll work.

PR #1922, opened 2026-05-12
Make computationExecutor autocloseable

oskarhurst:
## Problem Description

I think this is causing the flaky test failure we have been seeing. When you change the comp processor to be multi threaded we needed to make sure the comp executor was closing the thread correctly.

## how you tested the change

## Where the following done:


MikeNeilson (approved):
Though really we should probably just required the the executor always be provided. The creating one in the constructor was more of a stop gap. But I like that this accounts for that eventually by determining ownership of the pool.

PR #1927, opened 2026-05-12
fix to exit routing scheduler

### LINKED ISSUE #1807: Routing scheduler process does not exit when one thread dies
**Describe the bug**
Potentially we lost one thread on our LRGS routing spec - out of 816 locations, 56 of them stopped reporting yesterday after a patch.  There is no mention of them in the log after the patch.  Grepping in the log for missing DCP IDs or pathnames shows no information.  We discovered the missing data this morning, and I spot checked some of the gages and all the information was still there in DECODES.  Manually running a routing spec stored the data with no errors, so it didn't give me any indicator of a problem.  Restarting the routing scheduler process appears to have fixed the issue and those gages are reporting data in real-time now.  

But logging never stopped so our monitoring script never rebooted the process.

**To Reproduce**
1. Start routing scheduler process & add LRGS routing spec to run continuously
2. Monitor data after restart to potentially see thread die

**Expected behavior**
Expected the routing scheduler process to exit (and logging to stop) once one thread dies.  This would trigger the monitoring script to reboot the process and fix the issue.

 - OS: Windows 11, running 7.0.16 client
 - Server: 7.0.16 client

**Additional context**
For this specific case, this was using the NWO T7 on the NWDM database.  Routing scheduler process is called "scheduler-lrgs" and the log the day of the patch is called "scheduler-lrgs.log.2026.02.26.log".

This is an issue because in this instance a monitoring script is unable to trigger a reboot of the process.

Enovotny:
Title: Fix #1807 — RoutingScheduler should exit when a worker thread dies                                       
                                                                                                                    
  Summary                                                   
                                                                                                                  
  - Issue #1807: when a worker thread inside RoutingScheduler died from an uncaught exception, the parent JVM kept
  running. Monitoring (which restarts on process exit / log cessation) had no signal, so 56 of 816 LRGS-routed
  locations stopped reporting silently.
  - Adds two layered defenses: a JVM-default Thread.UncaughtExceptionHandler that hard-exits on any thread death,
  and a catch(Throwable) around the previously-unguarded post-acquisition pipeline in RoutingSpecThread.run() so    
  individual spec failures terminate cleanly and let the existing executive restart logic recover.
                                                                                                                    
  Changes                                                   
                                                                                                                  
  RoutingScheduler.java
  - New static newFatalHandler(IntConsumer exitAction) — returns an UncaughtExceptionHandler that logs the dead
  thread + cause at ERROR and calls the exit action with code 1. The IntConsumer seam exists so tests can verify
  behavior without killing the JVM.
  - oneTimeInit() calls Thread.setDefaultUncaughtExceptionHandler(newFatalHandler(System::exit)) before any worker
  thread is spawned. Covers RoutingSpecThread, StatusWriteThread, the anonymous lock-check thread,
  PurgeOldFilesThread, and any future workers.

  RoutingSpecThread.java
  - Added catch (Throwable ex) to the try-with-resources block at lines 521-601 (decode → format/output →
  platform-status DAO write). Previously any unchecked exception or Error from attemptDecode,                       
  formatAndOutputMessage, or DAO close() escaped run() and silently killed the thread.
  - On catch: increments error counters, logs at ERROR with platform name as context, sets done = true and          
  currentStatus = "ERROR-system" — the same termination signal the existing data-source error path uses, so
  ScheduleEntryExecutive.check() transitions to RunState.shutdown and the existing 60-second restart logic kicks in
  for continuous specs.

  RoutingSchedulerTest.java (new)
  - newFatalHandler_invokesExitWithCodeOne — direct unit test of the handler factory.
  - uncaughtExceptionInWorkerThread_triggersFatalHandler — spawns bad-worker thread that throws, verifies the
  JVM-default handler fires the captured exit action with code 1 and surfaces the thread name + cause. Original
  default handler restored in finally.

  How the two defenses interact

  1. Targeted catch (preferred path): a per-message decode/output/DAO failure → spec terminates cleanly → executive
  restarts it after 60s. The routing spec recovers without process restart.
  2. JVM-default handler (backstop): any other thread death (StatusWriteThread, lock-check, PurgeOldFilesThread,
  future workers, or anything we missed) → process exits → external monitoring restarts the JVM and surfaces the    
  dead thread + cause in the ERROR log.

  Out of scope

  - Restarting dead worker threads (issue explicitly requested exit, not restart).
  - Equivalent handlers in ComputationApp / other TsdbAppTemplate subclasses — same latent risk, separate issue.

Generated with Claude Code 4.7

### DIFF
--- a/java/opendcs/src/main/java/decodes/routing/RoutingScheduler.java
+++ b/java/opendcs/src/main/java/decodes/routing/RoutingScheduler.java
@@ -161,18 +161,28 @@ public class RoutingScheduler extends TsdbAppTemplate implements RoutingSchedule
 	 */
 	static Thread.UncaughtExceptionHandler newFatalHandler(IntConsumer exitAction)
 	{
-		return (t, e) ->
+		return (t, ex) ->
 		{
-			log.atError().setCause(e)
+			log.atError().setCause(ex)
 				.log("Uncaught exception in thread '{}' — exiting RoutingScheduler.", t.getName());
 			exitAction.accept(1);
 		};
 	}
 
+	/**
+	 * Installs the fatal-thread handler as the JVM default. Extracted so tests
+	 * can exercise the real install path with a captured exit action instead
+	 * of asserting only on the factory's output.
+	 */
+	static void installFatalHandler(IntConsumer exitAction)
+	{
+		Thread.setDefaultUncaughtExceptionHandler(newFatalHandler(exitAction));
+	}
+
 	@Override
 	protected void oneTimeInit()
 	{
-		Thread.setDefaultUncaughtExceptionHandler(newFatalHandler(System::exit));
+		installFatalHandler(System::exit);
 
 		/**
 		 * Using lock files as an IPC mechanism (for status GUI) is unreliable in windoze.
--- a/java/opendcs/src/main/java/decodes/routing/RoutingScheduler.java
+++ b/java/opendcs/src/main/java/decodes/routing/RoutingScheduler.java
@@ -33,6 +33,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.function.IntConsumer;
 
 import org.opendcs.utils.logging.OpenDcsLoggerFactory;
 import org.slf4j.Logger;
@@ -153,9 +154,26 @@ public class RoutingScheduler extends TsdbAppTemplate implements RoutingSchedule
 		appNameArg.setDefaultValue("RoutingScheduler");
 	}
 
+	/**
+	 * Builds a {@link Thread.UncaughtExceptionHandler} that logs the dead thread and
+	 * terminates the JVM via {@code exitAction}. Issue #1807: a silently dying worker
+	 * thread must take the process down so external monitoring restarts it.
+	 */
+	static Thread.UncaughtExceptionHandler newFatalHandler(IntConsumer exitAction)
+	{
+		return (t, e) ->
+		{
+			log.atError().setCause(e)
+				.log("Uncaught exception in thread '{}' — exiting RoutingScheduler.", t.getName());
+			exitAction.accept(1);
+		};
+	}
+
 	@Override
 	protected void oneTimeInit()
 	{
+		Thread.setDefaultUncaughtExceptionHandler(newFatalHandler(System::exit));
+
 		/**
 		 * Using lock files as an IPC mechanism (for status GUI) is unreliable in windoze.
 		 * Tell server lock never to exit as a result of lock file I/O error.
--- a/java/opendcs/src/main/java/decodes/routing/RoutingSpecThread.java
+++ b/java/opendcs/src/main/java/decodes/routing/RoutingSpecThread.java
@@ -599,12 +599,16 @@ public class RoutingSpecThread extends Thread
 
 				currentStatus = "Running";
 			}
-			catch (Throwable ex)
+			catch (Exception ex)
 			{
-				// Issue #1807: any unchecked exception/error from decode, output, or DAO
+				// Issue #1807: any unchecked exception from decode, output, or DAO
 				// access used to escape run() and silently kill this thread. Log with
 				// platform context, terminate the spec cleanly, and let
 				// ScheduleEntryExecutive restart it per its existing recovery logic.
+				// Errors (OOME, StackOverflow, etc.) are intentionally allowed to
+				// escape and reach the JVM-default UncaughtExceptionHandler installed
+				// by RoutingScheduler.oneTimeInit(), which exits the process so
+				// external monitoring restarts it.
 				numErrsRun++;
 				numErrsToday++;
 				log.atError().setCause(ex)
--- a/java/opendcs/src/main/java/decodes/routing/RoutingSpecThread.java
+++ b/java/opendcs/src/main/java/decodes/routing/RoutingSpecThread.java
@@ -599,6 +599,20 @@ public class RoutingSpecThread extends Thread
 
 				currentStatus = "Running";
 			}
+			catch (Throwable ex)
+			{
+				// Issue #1807: any unchecked exception/error from decode, output, or DAO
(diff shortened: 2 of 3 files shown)


java/opendcs/src/main/java/decodes/routing/RoutingScheduler.java
```diff
@@ -153,9 +154,26 @@ protected void addCustomArgs(CmdLineArgs cmdLineArgs)
...
 	}
 
+	/**
+	 * Builds a {@link Thread.UncaughtExceptionHandler} that logs the dead thread and
+	 * terminates the JVM via {@code exitAction}. Issue #1807: a silently dying worker
+	 * thread must take the process down so external monitoring restarts it.
+	 */
+	static Thread.UncaughtExceptionHandler newFatalHandler(IntConsumer exitAction)
+	{
+		return (t, e) ->
```
MikeNeilson:
```suggestion
		return (t, ex) ->
```

We use ex for exceptions. Need to find a good way to have automation deal with that (also a an issue within non-AI automation.)

java/opendcs/src/main/java/decodes/routing/RoutingScheduler.java
```diff
@@ -153,9 +154,26 @@ protected void addCustomArgs(CmdLineArgs cmdLineArgs)
...
+			log.atError().setCause(e)
+				.log("Uncaught exception in thread '{}' — exiting RoutingScheduler.", t.getName());
+			exitAction.accept(1);
+		};
+	}
+
 	@Override
 	protected void oneTimeInit()
 	{
+		Thread.setDefaultUncaughtExceptionHandler(newFatalHandler(System::exit));
```
MikeNeilson: Aggressive, probably the best way at the moment.

java/opendcs/src/main/java/decodes/routing/RoutingSpecThread.java
```diff
@@ -599,6 +599,20 @@ public void run()
 
 				currentStatus = "Running";
 			}
+			catch (Throwable ex)
```
MikeNeilson: Should probably catch Exception and not Throwable or RuntimeException and let the error handler above deal with it.

Then again could be valid here. 

What was the reasoning here? this looks like it would replace/negate the Thread exception handler.

java/opendcs/src/test/java/decodes/routing/RoutingSchedulerTest.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * Verifies the fatal-uncaught-exception handler added for issue #1807.
+ * A worker thread that throws must take the JVM down (here, signalled via the
+ * injected exit action) so external monitoring restarts RoutingScheduler.
+ */
+final class RoutingSchedulerTest
```
MikeNeilson, line 25: This tests that the `UncaughtExceptionHandler` works but not that the RoutingScheduler would fail correctly. So you've basically confirmed java works.

MikeNeilson (changes requested):

java/opendcs/src/main/java/decodes/routing/RoutingSpecThread.java
Enovotny: - The catch(Throwable) inside run() only wraps the post-acquisition decode/format/output block. On catch it sets done=true and currentStatus="ERROR-system", which signals ScheduleEntryExecutive.check() to transition to shutdown — and the existing 60-second restart logic (ScheduleEntryExecutive.java:149-161) brings the spec back. Net effect: one bad message in one routing spec terminates that spec cleanly without taking the JVM down. Important because losing the JVM means losing 815 working specs to recover one. - The JVM-default handler is the backstop for everything we didn't catch: StatusWriteThread, the anonymous lock-check thread (RoutingSpecThread.java:331-363), PurgeOldFilesThread in RawArchive, and any future workers. None of those have try/catch around their run(). They die silently today, exactly the failure mode #1807 reported. The handler turns those silent deaths into a process exit so monitoring restarts the JVM.
 
Throwable (vs Exception) is deliberate: an OutOfMemoryError or StackOverflowError from an oversized message should also drop the spec gracefully rather than escape and take the whole process down. Open to switching to Exception if you'd rather Errors propagate to the JVM-default handler — that's a tradeoff between "lose one spec on weird Errors" vs "lose the whole process on weird Errors."

java/opendcs/src/main/java/decodes/routing/RoutingScheduler.java
Enovotny: Agreed it's aggressive. The issue (#1807) explicitly asks for hard exit so external monitoring sees a dead process and restarts. We can revisit if we ever want a more targeted "restart-the-thread" recovery story.

java/opendcs/src/main/java/decodes/routing/RoutingScheduler.java
Enovotny: done

java/opendcs/src/test/java/decodes/routing/RoutingSchedulerTest.java
Enovotny: updated to: Extract the install into a small installFatalHandler(IntConsumer exitAction) helper that oneTimeInit() calls; the test then calls the helper directly and verifies the JVM default handler post-install fires the captured exit action when a thread throws. Tests the actual install path, not just the factory.

MikeNeilson: This is still just testing that Java works, not the the routing scheduler will "fail correctly" with the given conditions.

MikeNeilson (changes requested):
NOTE: this is still just testing java's thread error handling, not the RoutingScheduler itself.

Some additional tests have been added since that excercise Routing components. They don't fully setup a scheduler but might be a sufficient hint to whatever tools you are using to get something going.

PR #1917, opened 2026-05-09
Tasks/1225 rest api usage

MikeNeilson:
## Problem Description

Addition Fixes #1225. 

Describe the problem you are trying to solve.

## Solution

use new dao in rest api

## how you tested the change

update test to use realistic data and required mocks.

## Where the following done:

- [ ] Tests. Check all that apply:
   - [ ] Unit tests created or modified that run during ant test.
   - [ ] Integration tests created or modified that run during integration testing
         (Formerly called regression tests.)
   - [ ] Test procedure descriptions for manual testing
- [ ] Was relevant documentation updated?
- [ ] Were relevant config element (e.g. XML data) updated as appropriate

If you aren't sure leave unchecked and we will help guide you to want needs changing where.

PR #1928, opened 2026-05-13
DecodesConfigDaoTestIT CI failures

tlelv:
## Problem Description

DecodesConfigDaoTestIT was failing for the SQL-backed targets (OpenDCS-Postgres, OpenDCS-Oracle, CWMS-Oracle). I believe it is from:

Bug 1: Wrong data type standard
test_basic_operations and createConfig helper used CWMS like CWMS:Stage, CWMS:Elev, CWMS:Flow as the data type standard. The OKVI4 test seeds SHEF-PE data types (HG, PC, VB). The lookups for the CWMS types returned Optional.empty(), causing tests to fail with "DataType … doesn't exist."

Bug 2: Wrong pagination offset
 the original code used offset 1 when test_pagination called getAll(tx, 1, 0) to retrieve the first config. It then should assert "OKVI4" but it may skip OKVI4 (the only config present before the test creates its own). The first page assertion would return nothing or a test-generated config, not "OKVI4".

## Solution

Describe your solution.

changed DATATYPE_STANDARD constant from "CWMS" to "SHEF-PE"
changed SENSORS constant from {"Stage","Elev","Flow"} to {"HG","PC","VB"}
test_pagination: changed getAll(tx, 1, 1) → getAll(tx, 1, 0) so the first-page assertion correctly retrieves OKVI4

## how you tested the change

XML — all 3 tests skipped as expected
Postgres — all 3 tests passed
Oracle — all 3 tests passed

## Where the following done:

### DIFF
--- a/integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/DecodesConfigDaoTestIT.java
+++ b/integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/DecodesConfigDaoTestIT.java
@@ -37,7 +37,8 @@ import decodes.db.UnitConverterDb;
 class DecodesConfigDaoTestIT extends AppTestBase
 {
 
-    private static final String[] SENSORS = new String[]{"Stage", "Elev", "Flow"};
+    private static final String DATATYPE_STANDARD = "SHEF-PE";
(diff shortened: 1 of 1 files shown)


MikeNeilson (changes requested):
Reopen this as a PR to #1917

PR #1940, opened 2026-05-15
Lrl debounce app property

Enovotny:
## Problem Description        
                                                                                          
  Follow-up to #1919.                                                                                                                                                                  
                                                                                                                                                                                       
  `tasklistDebounceSeconds` was added in #1919 as a global VM-wide setting on `DecodesSettings`. That scope is wrong: the value tunes one specific app — the CompProc tasklist polling 
  loop — and operators have to set it in `decodes.properties` where it applies uniformly to every CompProc instance running on the JVM.                                                
                                                                                                                                                                                     
  Other CompProc tunables (`reclaimTasklistSec`, `checkTimedCompsSec`, `compRunWaitTime`) already live as per-app `CompAppInfo` properties on the `REF_LOADING_APPLICATION_PROP` row   
  for the loading application. `tasklistDebounceSeconds` should follow the same pattern so each CompProc instance can be tuned independently from its own DB row.
                                                                                                                                                                                       
  ## Solution                                                                                                                                                                        

  Move `tasklistDebounceSeconds` from `DecodesSettings` to a per-app `CompAppInfo` property, mirroring the existing `reclaimTasklistSec` wiring:

  - Removed the field and `PropertySpec` entry from `DecodesSettings`.                                                                                                                 
  - Added a `protected int tasklistDebounceSeconds` field plus public `getTasklistDebounceSeconds()` / `setTasklistDebounceSeconds(int)` to `TimeSeriesDb` (next to
  `reclaimTasklistSec`).                                                                                                                                                               
  - Added a `PropertySpec` entry to `ComputationApp.myProps[]` so the property is visible in the app-properties editor.                                                              
  - In `ComputationApp.runAppInit()`, read `appInfo.getProperty("tasklistDebounceSeconds")` with the same parse-with-default pattern as `reclaimTasklistSec`, then assign via
  `theDb.setTasklistDebounceSeconds(...)`. Moved the startup log line so it shows the resolved per-app value instead of the global default.                                            
  - Updated the three DAOs (`CwmsTimeSeriesDAO`, `HdbTimeSeriesDAO`, `OpenTimeSeriesDAO`) to read `((TimeSeriesDb) db).getTasklistDebounceSeconds()` instead of
  `DecodesSettings.instance().tasklistDebounceSeconds`.                                                                                                                                
                                                                                                                                                                                     
  **Operator migration:** any deployment with `tasklistDebounceSeconds=N` in `decodes.properties` will silently revert to `0`. The value must be moved to the CompProc app's           
  `REF_LOADING_APPLICATION_PROP` row before deploying this build (GUI: launcher → Processes → edit compproc app → Properties tab; or direct SQL).                                    

  ## how you tested the change

  - `./gradlew :opendcs:compileJava` — clean.
  - `./gradlew :testing:opendcs-tests:compileTestJava` — clean.
  - Updated `TasklistDebounceTestIT` to set the value via `db.setTasklistDebounceSeconds(N)` instead of mutating `DecodesSettings`. Existing assertions unchanged; the test still
  exercises the same fresh-row-excluded / aged-row-admitted behavior end-to-end through `getNewData` + `releaseNewData`.
  - `CwmsTasklistDebounceTestIT` is unchanged — it asserts the SQL predicate by hand-rolling the WHERE clause, so it was unaffected by where the value comes from.
  - Grep sweep confirms zero references to the removed `DecodesSettings.tasklistDebounceSeconds` field.
                                                                                                                                                                                       
  ## Where the following done:
                                                                                                                                                                                       
  - [x] Was relevant documentation updated? (PropertySpec description on the `tasklistDebounceSeconds` entry in `ComputationApp.myProps[]` carries the same operator-facing description
   that was previously on the `DecodesSettings` PropertySpec — surfaces in the app-properties editor.)

developed with Claude opus 4.7

### DIFF
--- a/java/opendcs/src/main/java/opendcs/opentsdb/OpenTimeSeriesDAO.java
+++ b/java/opendcs/src/main/java/opendcs/opentsdb/OpenTimeSeriesDAO.java
@@ -1721,7 +1721,7 @@ public class OpenTimeSeriesDAO extends DaoBase implements TimeSeriesDAI
 			failTimeClause = " and (a.FAIL_TIME is null OR "
 				+ System.currentTimeMillis() + " -  a.FAIL_TIME >= 3600000)";
 		String debounceClause = "";
-		int debounceSec = DecodesSettings.instance().tasklistDebounceSeconds;
+		int debounceSec = ((TimeSeriesDb)db).getTasklistDebounceSeconds();
 		if (debounceSec > 0)
 		{
 			// DATE_TIME_LOADED is BIGINT epoch millis (see V6.8__opendcs.sql),
--- a/java/opendcs/src/main/java/opendcs/opentsdb/OpenTimeSeriesDAO.java
+++ b/java/opendcs/src/main/java/opendcs/opentsdb/OpenTimeSeriesDAO.java
@@ -1696,7 +1696,7 @@ public class OpenTimeSeriesDAO extends DaoBase implements TimeSeriesDAI
 
 
 	@Override
-	public DataCollection getNewData(DbKey applicationId, int maxTake)
+	public DataCollection getNewData(DbKey applicationId, int maxTake, int tasklistDebounceSeconds)
 		throws DbIoException
 	{
 		// Reload the TSID cache every hour.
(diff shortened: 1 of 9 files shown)


java/opendcs/src/main/java/opendcs/opentsdb/OpenTimeSeriesDAO.java
```diff
@@ -1721,7 +1721,7 @@ public DataCollection getNewData(DbKey applicationId, int maxTake)
 			failTimeClause = " and (a.FAIL_TIME is null OR "
 				+ System.currentTimeMillis() + " -  a.FAIL_TIME >= 3600000)";
 		String debounceClause = "";
-		int debounceSec = DecodesSettings.instance().tasklistDebounceSeconds;
+		int debounceSec = ((TimeSeriesDb)db).getTasklistDebounceSeconds();
```
MikeNeilson: pass it in as a parameter, e.g. change the signature of the getNewData, we did that with the maxTake variable. Ends up cleaner and keeps everything in the ComputationApp for this particular property.

Having it retrieved this way is mixing concerns, which we try to avoid now.

MikeNeilson (changes requested):

java/opendcs/src/main/java/opendcs/opentsdb/OpenTimeSeriesDAO.java
Enovotny: change made

MikeNeilson (approved):

MikeNeilson:
You should sync and rebase to upstream main, I think Oskar has "fixed" the issue causing macos failure in storybook.... though if that is synced possible that hasn't been merged in yet. Will rerun manually after all the other jobs complete if you haven't changed anything.

MikeNeilson:
> You should sync and rebase to upstream main, I think Oskar has "fixed" the issue causing macos failure in storybook.... though if that is synced possible that hasn't been merged in yet. Will rerun manually after all the other jobs complete if you haven't changed anything.

Nevermind, that's hasn't been merged in yet.

PR #1943, opened 2026-05-18
change key to e

oskarhurst:
## Problem Description

replace em column name with the correct e

## how you tested the change

## Where the following done:


MikeNeilson (approved):
Yep, apparently I didn't have any equipment model info in the tests. arguably the query should be changed to em, but given no other use of e<something> this is valid here as well.

PR #1959, opened 2026-05-26
Modify ts store handler to account for possible difference in time pr…

### LINKED ISSUE #743: Comp Processor Processing Unchanged Data between Edits?
**Describe the bug**
It appears that CP is recomputing unchanged data in between two data points that have been edited. I thought this was a CWMS-VUE behavior, but using cwms-python/CDA, I'm thinking it's an OpenDCS behavior. 

**To Reproduce**
I wrote null values to two data spikes about 1 month apart. 
image
And it triggered 9,100 comps. 
image

**Expected behavior**
I would expect maybe 50 computations at most for two data points considering daily average etc.... I would expect it would only queue the data that was corrected....not the data in between too. 

OpenDCS 7.0.12

**Additional context**

MikeNeilson:
…ecision.

## Problem Description

Fixes #743. 

## Solution

Modifies how it is determined that values go into to the cp_comp_tasklist.

Primary change is the addition of the time series value data_entry_date to the message l_store_time, which would match the data_entry_date for any data changed.

Additionally remove the enqueue time check.

The use of the data_entry_date comparison allow data managers to edit bulk windows of data, but only reprocess data that was actually changed. 

NOTE: this does have the side affect where if data in between *should* be recomputed, it may not be. I believe the performance gain achieved by this warrants that issue; which needs to have it's own workaround anyways as a computation triggered by both regular and irregular data has the same problem.

It was also discovered that the data_entry_date values *may* include sub millisecond data. While this is not supposed to happen, the column is defined as `timestamp (6)` (microseconds) and data as been see in live databases that contain values to the microsecond.

## how you tested the change

- existing integration tests that make use of the queue
- MVP has been testing this for more than a year now
  - EXCEPT: remove of enqueue check, change of comparison to explicitly be millis.

## Where the following done:

### DIFF
--- a/schema/opendcs-schemas-cwms-oracle/src/main/resources/db/CWMS-Oracle/packages/R__cwms_ccp.sql
+++ b/schema/opendcs-schemas-cwms-oracle/src/main/resources/db/CWMS-Oracle/packages/R__cwms_ccp.sql
@@ -308,36 +308,29 @@ create or replace package body cwms_ccp as
     for r2 in
       (select distinct cc.loading_application_id
          from cp_comp_depends cd, cp_computation cc
-         where cd.ts_id = p_ts_code
+         where cd.ts_id = p_ts_code and cc.enabled = 'Y'
            and cc.loading_application_id is not null and cd.computation_id = cc.computation_id
       )
     loop
-      insert into cp_comp_tasklist
-      (  record_num,
-         loading_application_id,
-         site_datatype_id,
-         date_time_loaded,
-         start_date_time,
-         value,
-         unit_id,
-         delete_flag,
-         quality_code,
-         flags)
-      select
-         cp_comp_tasklistidseq.nextval,
-         r2.loading_application_id,
-         p_ts_code,
-         sysdate,
-         r1.date_time,
-         nvl(r1.value,0),
-         l_unit_id,
-         decode (r1.quality_code,5,'Y','N') delete_flag,
-         r1.quality_code,
-         r1.quality_code
-      from cwms_v_tsv r1
-      where r1.ts_code = p_ts_code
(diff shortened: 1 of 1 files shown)


schema/opendcs-schemas-cwms-oracle/src/main/resources/db/CWMS-Oracle/packages/R__cwms_ccp.sql
```diff
@@ -308,36 +308,29 @@ create or replace package body cwms_ccp as
...
-        and r1.data_entry_date <= p_enqueue_time;
+      for r1 in
+        (select date_time,value,quality_code from cwms_v_tsv
+           where ts_code = p_ts_code and date_time between l_start_time and l_end_time
+            -- and p_enqueue_time >= data_entry_date
+             and (cwms_util.to_millis(p_store_time) = cwms_util.to_millis(data_entry_date) OR p_store_time is NULL)
+        )
+      loop
+        l_delete_flag := 'N';
+        if (r1.quality_code = 5) then
```
ktarbet: anything we can do for this magic  '5'   in SQL here?

ktarbet (dismissed):

schema/opendcs-schemas-cwms-oracle/src/main/resources/db/CWMS-Oracle/packages/R__cwms_ccp.sql
MikeNeilson: Yeah, should be a constant defined somewhere within CWMS_20. If not it can be created here for preference. It's the quality code for "missing value"

MikeNeilson:
I think think this should be merged in, however it did not solve the problem at hand.

ktarbet (approved):

PR #1960, opened 2026-05-26
add transport media and platform sensors

oskarhurst:
## Problem Description

Added the transport media table and platform sensor to the platform details. Also fixed configs api bug to my knowledge  dao.getAll(tx, -1, -1) call in ConfigResources.getConfigRefs() is extremely expensive because it is joining all ten tables just to throw aways everything but the configs this giant request would fail and time out or OOM. solution I replaced it with a simple query.

## how you tested the change

## Where the following done:


java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ConfigResources.java
```diff
@@ -121,12 +122,32 @@ public Response getConfigRefs() throws WebAppException
...
-								   .map(pc -> mapRef(pc))
-								   .toList();
-			return Response.ok().entity(configs).build();
+			final Handle handle = tx.connection(Handle.class)
+									.orElseThrow(() -> UNABLE_TO_GET_CONFIG_DAO);
+			// Lightweight refs-only query. The full DAO.getAll loads every joined
+			// sensor / script / format-statement / unit-converter row per config,
+			// which OOMs / times out on databases with many configs. The chooser
+			// only needs id / name / description / # platforms.
+			final String sql =
```
MikeNeilson: Not the place to put this.

Move it into the DecodesConfigDaoImpl in one of the following ways:
1. Additional method
2. A parameter that allows specifying if all the tables should be joined or not.


I appreciate the speed and memory concern, though I'm a bit surprised it would OOM will have to test that out later, but this is a complete lack of separation of concerns.

javascript/opendcs-web-ui-react/src/pages/platforms/TransportMedium.tsx
```diff
@@ -0,0 +1,550 @@
...
+import { useTranslation } from "react-i18next";
+import type { ApiTransportMedium } from "opendcs-api";
+import { DetailFade } from "../../components/data-table";
+import type { CancelAction } from "../../util/Actions";
+import { TransportMediumReducer } from "./TransportMediumReducer";
+
+const INPUT_H = { height: "2.25rem" };
+const LABEL_H = { height: "1rem" };
+
+const MEDIUM_TYPES = [
```
MikeNeilson: These should be pulled from a ref list enum.

javascript/opendcs-web-ui-react/src/pages/platforms/TransportMedium.tsx
```diff
@@ -0,0 +1,550 @@
...
+  "GOES-RANDOM",
+  "IRIDIUM",
+  "POLLED-MODEM",
+  "POLLED-TCP",
+  "INCOMING-TCP",
+  "DATA-LOGGER",
+  "OTHER",
+] as const;
+
+const BAUD_RATES = [300, 1200, 2400, 4800, 9600, 19200, 38400, 57600, 115200];
```
MikeNeilson: these should probably be pulled from a ref list enum as well, but to be fair it's unlikely they will change any time soon.

javascript/opendcs-web-ui-react/src/pages/platforms/TransportMedium.tsx
MikeNeilson, line 1: Consider what happens when a new type of Transport Medium gets added.

While we don't need to account for a totally dynamic new type now, we want to be reasonably prepared with what we do have.

The detail options for each transport media type should be their own component that gets loaded in. Lazy loading should help here.

MikeNeilson (changes requested):
Conceptually good. A few things are implemented in the wrong spot.

MikeNeilson (approved):

PR #1961, opened 2026-05-26
Add live process status to Loading Apps

tlelv:
## Problem Description

The Loading Apps table had no way to tell whether an app process was actually
running or not. 

## Solution

Added a live Status column to the Loading Apps table. It polls the
/appstat endpoint and renders a green when the app has an active PID, or gray otherwise.

## how you tested the change

- Status column appears in the Loading Apps table
- Badge updates automatically after 30ish seconds when a process starts or stops
- Confirmed the page still loads normally when `/appstat` is unavailable
- All existing unit and storybook tests pass 

- Inserted a fake running record into cp_comp_proc_lock for compproc (appId 4) directly in Postgres to simulate a live process. Confirmed compproc showed a green Running badge in the table while all other apps showed gray Inactive. 

Screenshot 2026-05-26 at 4 49 13 PM

## Where the following done:

### DIFF
--- a/javascript/opendcs-web-ui-react/public/locales/de-DE/loadingapps.json
+++ b/javascript/opendcs-web-ui-react/public/locales/de-DE/loadingapps.json
@@ -10,5 +10,8 @@
   "edit_app": "Ladeanwendung {{id}} bearbeiten",
   "save_app": "Ladeanwendung {{id}} speichern",
   "cancel_for": "Bearbeitung von Ladeanwendung {{id}} abbrechen",
-  "delete_for": "Ladeanwendung {{id}} löschen"
+  "delete_for": "Ladeanwendung {{id}} löschen",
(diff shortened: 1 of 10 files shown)


MikeNeilson (dismissed):

MikeNeilson:
Nice change, the original UI doesn't have this, at least directly, and it's a really cool at a glance feature.

Don't worry about the failed test, I'll rerun it after all the builds finished. Unrelated to this work and will need to be investigated separately.

MikeNeilson:
You'll have to mock the HTTP request/response for the application status, but it shouldn't be too difficult to do. That'll bring the test coverage up and help keep regression out. 

Change otherwise excellent. I'm just being a stickler about tests.

MikeNeilson (approved):

PR #1962, opened 2026-05-27
add configs page

oskarhurst:
## Problem Description

update properties to have custom titles and add a configs table with sensors and decodes script table.

## how you tested the change

## Where the following done:


javascript/opendcs-web-ui-react/src/pages/decodes/configs/ConfigSensorsTable.tsx
```diff
@@ -0,0 +1,149 @@
...
+    [t],
+  );
+
+  const rowActions = useMemo<RowAction<ApiConfigSensor>[]>(
+    () =>
+      edit
+        ? [
+            {
+              key: "edit",
+              icon: "bi-pencil",
```
MikeNeilson, line 90: These are showing up so small that it's just yellow and read rounded rectangle.

Icons need to be large enough to be seen. Reasonably easy to click.

javascript/opendcs-web-ui-react/src/pages/decodes/configs/ConfigScriptsTable.stories.tsx
```diff
@@ -0,0 +1,161 @@
...
+  { sensorNumber: 2, sensorName: "Battery" },
+];
+
+const scripts: ApiConfigScript[] = [
+  {
+    name: "Main",
+    headerType: "goes",
+    dataOrder: ApiConfigScriptDataOrderEnum.D,
+    formatStatements: [{ sequenceNum: 1, label: "Stage", format: "/,4f(s,a,6D' ',1)" }],
+    scriptSensors: [{ sensorNumber: 1, unitConverter: { ucId: 1 } }],
```
MikeNeilson, line 23: The actual units value is not showing - just `Loading`, can't remember if the unitConverter type has a unit (I think perhaps) but that value, in read only should be the unit set for this script, or if not set `raw`.

MikeNeilson (changes requested):
Overall seems reasonable. the flow of scrolling for the script isn't ideal, but it's not as drastically bad as I was expecting.

RE the button sizing. Consider two people:
1. My not great sighted self... and others who have it worse.
2. Someone coming in cold. 

For 2, it's *possible* that by the time they get here, and see other controls that the orange and read will register as edit and delete; however, you are increasing cognitive load whether that relationship has been established or not. The use of icon over text is a reasonable conceit for us as it is *unlikely* that someone using OpenDCS is coming into computing in general cold.


A mouse over with text could also work; however, again additional cognitive load, and an additional barrier to understanding the button.

However, perhaps we can reach a compromise... a compressed and expanded mode. If someone is comfortable with the simple color boxes great, but I'd rather start by default with very explicit.

MikeNeilson (approved):

PR #1964, opened 2026-05-28
add the routing page to react ui

oskarhurst:
## Problem Description

Add the Routing page and update the routing api to correcting return the consume in the routingref

## how you tested the change

## Where the following done:


javascript/opendcs-web-ui-react/src/components/data-table/MultiSelectorModal.tsx
MikeNeilson, line 1: Needs test. Even if no assertions, just something that can be interacted with within storybook independently for future testing and changes.

javascript/opendcs-web-ui-react/src/pages/decodes/routing/Routing.tsx
```diff
@@ -0,0 +1,517 @@
...
+                      <option key={opt} value={opt}>
+                        {opt}
+                      </option>
+                    ))}
+                  </Form.Select>
+                </Col>
+              </FormGroup>
+            </Col>
+            <Col md={6}>
+              {flagCheck("ascendingTime", t("routing:ascending_time"))}
```
MikeNeilson: If it's not crazy difficult, it would be good if this set of data was it's own Component.

I don't want to try and "fix" it now, but this particular portion hard codes way too much knowledge of specific systems such that it's difficult to expand to searches within other data source types. Something we'd like to improve in the future and starting with some good UI component isolation should help with that thinking.

javascript/opendcs-web-ui-react/src/components/controls/SinceUntilEditor.tsx
MikeNeilson, line 1: This should be moved to the components directory. The Computations, and several other elements, also use this style of relative time.

MikeNeilson (changes requested):
Looks good. Couple of maintainability nitpicks but that's it.

MikeNeilson (approved):

PR #1965, opened 2026-05-28
Refactor/split up user management

MikeNeilson:
## Problem Description

Need to do some work on roles and how they're handled.

the UserManagementDao was one of the first created and does not contain any lessoned learned since.
Lesson 1, too many eggs in one basket.
Lesson 2, single "save" method vs add and update.

This PR addresses only lesson 1, 2 will be dealth with in a followup. Already a bit large.

## Solution

Split User Management operations into 3 distinct DAOs: Users, Roles, IdentityProviders.

Take existing combined implementation and break out implementations as needed.

## how you tested the change

- existing integration tests. No functional changes.

## Where the following done:


MikeNeilson:
Duplication will be reduced in follow up PRs, this is an initial reorg.

May even be able to completely remove the "CWMS" only variants. This was first "modernized" DAO written and does not include any of the lessons learned in organizing the queries and methods.

ktarbet (approved):

PR #1970, opened 2026-05-29
add the schedule entries page to the react app

oskarhurst:
## Problem Description

Add the schedule entries page to the react UI and set Vite test to only print on failed tests

## how you tested the change

## Where the following done:


javascript/opendcs-web-ui-react/src/pages/schedule/SchedulesTable.tsx
MikeNeilson, line 1: Don't do this now, but it would be good to give this the same treatment Treyson did to the "Loading Apps" page and show the current status.... assuming we have an API endpoint that gives us said status.

MikeNeilson (approved):

PR #1982, opened 2026-06-03
Add timezone to Sites and since/until time window to Computations

tlelv:
## Problem Description

The timezone field for sites, and the  `effectiveStartType/Date/Interval` and `effectiveEndType/Date/Interval` for computations were present in the legacy Java Swing editor were missing from the web UI.

## Solution

Describe your solution.

Added them to the respective sites 

## how you tested the change

Screenshot 2026-06-03 at 1 53 21 PM
Screenshot 2026-06-03 at 1 53 00 PM

## Where the following done:


javascript/opendcs-web-ui-react/src/pages/computations/computations/Computation.tsx
```diff
@@ -165,6 +165,18 @@ const requiredParmsFromAlgorithm = (algorithm?: ApiAlgorithm): ApiCompParm[] =>
       algoParmType: parm.parmType,
     }));
 
+const START_TYPES = ["No Limit", "Now -", "Calendar"] as const;
```
MikeNeilson: Good change, but consider if the new logic can be added here: https://github.com/opendcs/opendcs/blob/510458e87c09812961d7ad462e5aab9c6cac67e7/javascript/opendcs-web-ui-react/src/components/controls/SinceUntilEditor.tsx and Computations updated to use it.

MikeNeilson (changes requested):

javascript/opendcs-web-ui-react/src/pages/computations/computations/Computation.tsx
```diff
@@ -451,91 +547,49 @@ export const Computation: React.FC<ComputationProperties> = ({
                   </Col>
                 </FormGroup>
                 <FormGroup as={Row} className="mb-3">
-                  <Form.Label column sm={3} htmlFor="effectiveStartType">
```
MikeNeilson: While the control is called SinceUntil, The name here of "Effect..." is still correct in the context of computations.

MikeNeilson (changes requested):
Keep the naming of "Effective" in computation, but otherwise looks fine.

MikeNeilson (dismissed):

MikeNeilson (approved):

PR #1987, opened 2026-06-04
Task/1227 interval dao

### LINKED ISSUE #1227: Interval DAO

None

**Additional context**

*MAY* make sense to include within Time Series DAO instead of separate.

MikeNeilson:
## Problem Description

Fixes #1227. 

## Solution

- Create new IntervalDurationDao.
- Separate Methods for intervals and durations.

- OpenDCS Impl uses one list, allowed additons.
- CWMS Impl uses separate lists, no modifications allowed.

## how you tested the change

- New Integration Tests.

## Where the following done:

### DIFF
--- a/java/cwms-implementation/src/main/java/org/opendcs/database/impl/cwms/dao/CwmsIntervalDurationDao.java
+++ b/java/cwms-implementation/src/main/java/org/opendcs/database/impl/cwms/dao/CwmsIntervalDurationDao.java
@@ -1,0 +1,220 @@
+package org.opendcs.database.impl.cwms.dao;
+
+import java.util.ArrayList;
+import java.util.Calendar;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+import org.opendcs.database.api.DataTransaction;
+import org.opendcs.database.api.OpenDcsDataException;
+import org.opendcs.database.dai.IntervalDurationDao;
+import org.openide.util.lookup.ServiceProvider;
+
+import decodes.sql.DbKey;
+import decodes.tsdb.IntervalCodes;
+import opendcs.opentsdb.Interval;
+
+/**
+ * While this class currently does no actual Database access, that is planned for the future.
+ * Such updates will require sorting out how to get the cwms encoding of durations and intervals
+ * into the Interval structure of OpenDCS. In the mean time, we continue to just hard code things.
+ */
+@ServiceProvider(service = IntervalDurationDao.class, path = "dao/CWMS-Oracle")
+public class CwmsIntervalDurationDao implements IntervalDurationDao
+{
+    private static final List<Interval> intervals = new ArrayList<>();
+    private static final List<Interval> durations = new ArrayList<>();
+    private static final OpenDcsDataException INTERVALS_CONTROLLED = new OpenDcsDataException("CWMS Intervals are controlled by the CWMS Database.");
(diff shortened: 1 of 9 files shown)


java/opendcs/src/main/java/org/opendcs/database/dai/IntervalDurationDao.java
```diff
@@ -0,0 +1,66 @@
...
+ * An Interval is the time between two samples.
+ * The duration is a window over which data was aggregated.
+ * 
+ * Thus this interace is called IntervalDurationDao, as it is rare the usage
+ * would ever need one but not the other.
+ * 
+ * Unfortunately, some implementations do no distinguish between the two.
+ * This interface does, if you're system does not, share logic appropriately.
+ */
+public interface IntervalDurationDao extends OpenDcsDao
```
ktarbet, line 28:
```suggestion
 * DAO to retrieve Intervals or Durations valid within this database.
 * 
 * Intervals and Durations, while related, are not the same concept.
 * 
 * An Interval is the time between two samples.
 * The duration is a window over which data was aggregated.
 * 
 * Thus this interface is called IntervalDurationDao, as it is rare the usage
 * would ever need one but not the other.
 * 
 * Unfortunately, some implementations do not distinguish between the two.
 * This interface does, if your system does not, share logic appropriately.
 */
public interface IntervalDurationDao extends OpenDcsDao
```

java/opendcs-implementation/src/main/java/org/opendcs/database/impl/opendcs/dao/OpenDcsIntervalDurationDaoImpl.java
```diff
@@ -0,0 +1,187 @@
...
+                        .orElseThrow(() -> new OpenDcsDataException("No key generator configured."));
+        
+        try (var query = handle.createUpdate(MERGE_SQL)
+                               .define("dual", ctx.getDatabase() == DatabaseEngine.ORACLE ? " from dual " : ""))
+        {
+            DbKey id = interval.getKey();
+            var existing = findIntervalByName(tx, interval.getName());
+            if (existing.isPresent())
+            {
+                // If there's an existing app with this name, we'll just assume the provided id, if any, was in error
```
ktarbet:
```suggestion
                // If there's an existing interval with this name, we'll just assume the provided id, if any, was in error
```

java/cwms-implementation/src/main/java/org/opendcs/database/impl/cwms/dao/CwmsIntervalDurationDao.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    @Override
+    public List<Interval> getAllDurations(DataTransaction tx) throws OpenDcsDataException
+    {
+        return Collections.unmodifiableList(durations);
+    }
+
+    
+    private static void loadDurations()
```
ktarbet, line 106: It would be nice if each database could have a resource file that defined durations and intervals; somehow,

ktarbet (dismissed):

java/cwms-implementation/src/main/java/org/opendcs/database/impl/cwms/dao/CwmsIntervalDurationDao.java
MikeNeilson: These are actually in the database and should be looked up. Unfortunately for CWMS it's all "minutes" between samples vs a calendar constant and multiplier.... which may be fine, just needs more testing and possibly logic in place before going that far. So right now I'm just copying the hand jam over so it works.

PR #1973, opened 2026-06-01
Initial interface design of TimeSeriesIdentifierDao

MikeNeilson:
## Problem Description

Fixes #650. 

NOTE: No implementation yet, opening PR with interface design to get feedback as soon as possible.

## Solution

- Implement independent TimeSeriesIdentifierDao as new stateless DAO.
- Extract all "operate on/provide/etc" TimeSeriesIdentifier operations from TimeSeriesDAI, TimeSeriesDb, and others as found.

NOTE: implementation will not be removed from those places at this time. Duplication warnings are expected.

## how you tested the change

Will create new Dao Tests for this DAO.

## Where the following done:

- [ ] Tests. Check all that apply:
   - [ ] Unit tests created or modified that run during ant test.
   - [ ] Integration tests created or modified that run during integration testing
         (Formerly called regression tests.)
   - [ ] Test procedure descriptions for manual testing
- [ ] Was relevant documentation updated?
- [ ] Were relevant config element (e.g. XML data) updated as appropriate

If you aren't sure leave unchecked and we will help guide you to want needs changing where.

PR #1990, opened 2026-06-06
Use a map for setting instances so we don't keep expanding input para…

MikeNeilson:
…meters.

## Problem Description

During #1973 realized we needed yet-another-opendcs settings object to provide some key data required.

## Solution

Instead of adding another parameter, simplify to a `Map<Class<? extends OpenDcsSettings>, OpenDcsSettings>`

Also simplified the return requested settings logic.

## how you tested the change

Existing integration and other tests. No functional change, pure refactor.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/org/opendcs/database/SimpleOpenDcsDatabaseWrapper.java
+++ b/java/opendcs/src/main/java/org/opendcs/database/SimpleOpenDcsDatabaseWrapper.java
@@ -59,23 +59,24 @@ import decodes.util.DecodesSettings;
 public class SimpleOpenDcsDatabaseWrapper implements OpenDcsDatabase
 {
     private static final Logger log = OpenDcsLoggerFactory.getLogger();
-    protected final DecodesSettings settings;
+    //protected final DecodesSettings settings;
     private final Database decodesDb;
     private final TimeSeriesDb timeSeriesDb;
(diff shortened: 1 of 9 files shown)


java/opendcs/src/main/java/org/opendcs/database/SimpleOpenDcsDatabaseWrapper.java
```diff
@@ -98,13 +99,14 @@ public SimpleOpenDcsDatabaseWrapper(DecodesSettings settings, Database decodesDb
             throw new IllegalStateException("Unable to determine database type", ex);
         }
 
+        DecodesSettings decodesSettings = (DecodesSettings)settingsMap.get(DecodesSettings.class);
```
ktarbet, line 102: consider null check; get(DecodesSettings.class) might return null

ktarbet (approved):

java/opendcs/src/main/java/org/opendcs/database/SimpleOpenDcsDatabaseWrapper.java
MikeNeilson: I added it to #1996 ... without thinking about which PR this was. But that one is already getting a bunch of other changes so I think it makes sense there.

PR #1996, opened 2026-06-08
Refactor/1991 database wrapper

### LINKED ISSUE #1991: Reorganize SimpleOpenDcsDatabaseWrapper to something that makes more conceptual sense
`org.opendcs.database.SimpleOpenDcsDatabaseWrapper`

**Purpose of refactoring**
It is no long "simple", nor primarily a wrapper anymore.

**Proposed method**
1. Rename to `AbstractOpenDcsDatabaseWrapper`, or perhaps `AbstractOpenDcsJdbiDatabaseWrapper` just to make the Jdbi expectation more clear
2. Determine if any methods actually should be abstract (I'm thinking some sort of initial setup.
3. Adjust downstream classes as required.

**Additional context**
I think we might be done adding "major" features, we already have the dao and settings lookup, but It's definitely no longer simple.

MikeNeilson:
## Problem Description

Fixes #1991. 

NOTE: depends on #1990

## Solution

- Rename `SimpleOpenDcsDatabaseWrapper` to `AbstractJdbiOpenDcsDatabaseWrapper` to reflect 1) it's not actually simple, 2) the general intent.
- Rename `getDatabase` to `getDatabaseEngine` as we aren't actually getting a database instance.
- Move definition of `getSettings` to interface as the definition is shared by `OpenDcsDatabase` and `TransactionContext`
- Added abstract `initialSetup` method to `AbstractJdbiOpenDcsDatabaseWrapper` to handle any future 1 time init. (Can be used for things like defines that are engine specific to only do once, or registering specific ArgumentFactories or Mappers.)

## how you tested the change

Existing integration tests. No functional change. Just renaming and reorganization of things.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/org/opendcs/database/AbstractJdbiOpenDcsDatabaseWrapper.java
+++ b/java/opendcs/src/main/java/org/opendcs/database/AbstractJdbiOpenDcsDatabaseWrapper.java
@@ -108,6 +108,7 @@ public abstract class AbstractJdbiOpenDcsDatabaseWrapper implements OpenDcsDatab
         {
             throw new IllegalStateException("Unable to create key generator of type '" + decodesSettings.sqlKeyGenerator + "'", ex);
         }
+        initialSetup();
     }
 
     /**
--- a/java/opendcs/src/main/java/org/opendcs/database/AbstractJdbiOpenDcsDatabaseWrapper.java
+++ b/java/opendcs/src/main/java/org/opendcs/database/AbstractJdbiOpenDcsDatabaseWrapper.java
@@ -59,18 +59,16 @@ import decodes.util.DecodesSettings;
 public abstract class AbstractJdbiOpenDcsDatabaseWrapper implements OpenDcsDatabase
 {
     private static final Logger log = OpenDcsLoggerFactory.getLogger();
-    //protected final DecodesSettings settings;
     private final Database decodesDb;
     private final TimeSeriesDb timeSeriesDb;
     protected final DataSource dataSource;
     protected final Jdbi jdbi;
     protected final DatabaseEngine dbEngine;
     protected final KeyGenerator keyGenerator;
-    //protected final DatabaseQuerySettings querySettings;
     protected final Map<Class<? extends OpenDcsSettings>, OpenDcsSettings> settingsMap = new HashMap<>();
     private final Map<Class<? extends OpenDcsDao>, DaoWrapper<? extends OpenDcsDao>> daoMap = new HashMap<>();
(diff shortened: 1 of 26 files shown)


java/opendcs/src/main/java/org/opendcs/database/AbstractJdbiOpenDcsDatabaseWrapper.java
```diff
@@ -98,16 +97,22 @@ public SimpleOpenDcsDatabaseWrapper(DecodesSettings settings, Database decodesDb
...
         {
-            throw new IllegalStateException("Unable to create key generator of type '" + settings.sqlKeyGenerator + "'", ex);
+            throw new IllegalStateException("Unable to create key generator of type '" + decodesSettings.sqlKeyGenerator + "'", ex);
         }
     }
 
+    /**
+     * Handle any additional setup or overrides to what was previously done.
+     */
+    protected abstract void initialSetup();
```
ktarbet, line 118: consider calling initialSetup()  in the ctor.

ktarbet (dismissed):

java/opendcs/src/main/java/org/opendcs/database/AbstractJdbiOpenDcsDatabaseWrapper.java
ktarbet, line 157: CwmsLocationLevelDAO  -- doesn't feel abstract.

java/opendcs/src/main/java/org/opendcs/database/AbstractJdbiOpenDcsDatabaseWrapper.java
MikeNeilson: ... duh. Thank you, actually did mean to do that.

java/opendcs/src/main/java/org/opendcs/database/AbstractJdbiOpenDcsDatabaseWrapper.java
MikeNeilson: It's not, I basically let @oskarhurst get away with that one to get going faster. Those DAOs do need the update treatment, but they're currently used in algorithms so it's a bit of a rock and an hardplace... or maybe chicken and egg issue to update.

ktarbet: I've created an issue to track this: 
https://github.com/opendcs/opendcs/issues/1999

ktarbet (approved):

PR #1997, opened 2026-06-08
Bugfix/lrl data gap

MikeNeilson:
## Problem Description

We're seeing an odd gap in data for a group computation using ScalerAdder for a set of projects.
Data is inserted in batches of several hours worth at a time and data saves quickly.

However, in-situ, at least one project has no data compute, nor does it see a record of the data. We have confirmed that data has at least made it to the queue and is processed by the queue handler.

## Solution

Create test setup of ScalerAdder with the entire groups worth of data. Additionally, we add several additional random time series that the queue handler has to deal with.
However, the issue is not present in the test.

Add this time the data is saved sequentially. Will attempt randomization of data storage to better mimic real world conditions.

## how you tested the change

No changes, added integration test.

## Where the following done:


integrationtesting/opendcs-tests/src/test/resources/data/CompProc/ScalerAdder/input.tsimport
ktarbet, line 1: It seems odd that all the time-series input data has the same values.  Consider having the first time series have all 1,1,1,1 
the second 2,2,2,2    -- may be helpful for debugging.

integrationtesting/opendcs-tests/src/test/java/org/opendcs/regression_tests/CompProcTestIT.java
ktarbet, line 187: Considering the "new Date()"  , what if sample_time occurred after the date_time_loaded, would that confuse anything.

integrationtesting/opendcs-tests/src/test/java/org/opendcs/regression_tests/CompProcTestIT.java
```diff
@@ -195,4 +195,92 @@ public void test_bad_recs_cleared() throws Exception
...
+                                     new File(configuration.getUserDir(),"compproctest-adder-update.log"),
+                                     environment, 60, TimeUnit.SECONDS, "-O");
+
+        // First import is to create time series so compdepends can be updated to use them.
+        // This import is to actually trigger the computation.
+        Programs.ImportTs(
+            new File(config.getUserDir(),"/importTs.log"),
+            config.getPropertiesFile(),
+            environment, exit,
+            getResource(config, "CompProc/ScalerAdder/input.tsimport"));
```
ktarbet, line 254: If we are not reproducing yet -- consider importing a single time-series at a time to get more gears moving; possibly more realistic -- perhaps generate the file on the fly with simple data.

ktarbet (approved):

integrationtesting/opendcs-tests/src/test/java/org/opendcs/regression_tests/CompProcTestIT.java
MikeNeilson: In the context of this test it doesn't matter. Though could be something to worry about in the future. Current change is just the cleanup of not needing the method to be public.

integrationtesting/opendcs-tests/src/test/java/org/opendcs/regression_tests/CompProcTestIT.java
MikeNeilson: First test was with a single time series. The data itself is loaded in batches.

Attempting to load a single time series at a time is *less* likely to recreate the issue.

I suspect something with data volume and tasklist churn. But there has yet to be a smoking gun indicator.

This current design is currently as class to "what is happening" as I am able to determine.

integrationtesting/opendcs-tests/src/test/resources/data/CompProc/ScalerAdder/input.tsimport
MikeNeilson: I may later, this was simply easier to setup. I'm also not considered about the exact data for this test as there are no outlier values present into the real data sets.

PR #1998, opened 2026-06-08
Add additional helper and one sample usage.

MikeNeilson:
## Problem Description

It's annoying to constantly type out the `prefix + "<Column>"`

@ktarbet provided a suggestion on #1973 and this is what I came up with.

## Solution

- Create an interface that Enums of columns can implement to return the column name.
- Create these enums for every mapper.
- Update Mappers and DAOs for the change.
- PrefixRowMapper handles given an enum, provide column name with the required prefix.

NOTE: While the DAOs usage in binds *SHOULD* be directly using the Specific mapper provided Enum, that seems like too much change to look at in this PR which is now already quite large. 

I will go back through and do that in a follow up PR.

## how you tested the change

Existing integration tests. No functional change.

## Where the following done:


ktarbet (dismissed):

ktarbet (dismissed):

java/opendcs/src/main/java/org/opendcs/database/model/mappers/engineeringunit/EngineeringUnitMapper.java
```diff
@@ -23,12 +24,39 @@ public static EngineeringUnitMapper withPrefix(String prefix)
 
     @Override
     public EngineeringUnit map(ResultSet rs, StatementContext ctx) throws SQLException
-    {   
+    {
```
ktarbet, line 27: Still using 'legacy' prefix +"name"  here.

java/opendcs/src/main/java/org/opendcs/utils/sql/GenericColumns.java
ktarbet, line 1: could a  .toString() method here,    help avoid needing to call .column() >

java/opendcs/src/main/java/org/opendcs/database/model/mappers/PrefixRowMapper.java
```diff
@@ -1,19 +1,38 @@
...
 
+import java.sql.SQLException;
+import java.util.EnumSet;
+
 import org.jdbi.v3.core.mapper.RowMapper;
+import org.opendcs.database.sql.TableColumnDefinition;
 
 /**
  * Helper class for row mappers to take a prefix.
  * If the provided prefix does not start end with and underscore (_)
```
ktarbet:
```suggestion
 * If the provided prefix does not start end with an underscore (_)
```

java/opendcs/src/main/java/org/opendcs/database/model/mappers/PrefixRowMapper.java
```diff
@@ -1,19 +1,38 @@
...
  * one will be added.
+ * 
+ * A enum of defined columns must also be provided. At this time only
+ * the column name is handled. Future work will allow building a
+ * <pre> table.column prefix_column</pre> list for joins.
+ * 
+ * @param <T> The type that the mapper will return
+ * @param <E> A enum containing all columns. This enum must implement the {@see TableColumnDefinition} interface.
```
ktarbet:
```suggestion
 * @param <E> A enum containing all columns. This enum must implement the {@link TableColumnDefinition} interface.
```

ktarbet:
I'm seeing matches still for this regex.

`
rs\..*Prefix.*\+"
`

java/opendcs/src/main/java/org/opendcs/database/model/mappers/engineeringunit/EngineeringUnitMapper.java
MikeNeilson: oops, thanks

java/opendcs/src/main/java/org/opendcs/utils/sql/GenericColumns.java
MikeNeilson: sadly no, tried it already. no c++ style implicit casting in Java.

MikeNeilson:
> I'm seeing matches still for this regex.
> 
> `rs\..*Prefix.*\+"`

thanks for catching that. Also updated some other prefix usages that I realized we can get away from now.

java/opendcs/src/main/java/org/opendcs/utils/sql/GenericColumns.java
ktarbet: // consider
  @Override
    public String toString()
    {
        return this.column;   
    }

// so we get 'name'  instead of 'NAME', for cases we are not yet calling .column()

MikeNeilson: I can, but where are you seeing 'NAME', it seemed I had to update every usage of the now enum to .column() so I'm not sure exactly where it would show up.

ktarbet: If the databases are not case sensitive, perhaps doesn't matter.

This regex finds some that may apply.

`
\+\s*GenericColumns
`

ktarbet (approved):

java/opendcs/src/main/java/org/opendcs/utils/sql/GenericColumns.java
MikeNeilson: ah now I see it, those will get updated in another way. and I'd like to reserve the toString to be more informational, like in the future we'll include type info.

PR #2003, opened 2026-06-10
Rename OpenTsdbSetting, setup system to allow Implementation specific…

MikeNeilson:
… settings to be loaded and set appropriate instances.

## Problem Description

While working on #1973 realized there were some settings we needed access to that we couldn't readily get.

## Solution

- Setup Implementation Specific Setting object and method to load values from the tsdb_property table.
- Also did some additional cleanup. Renaming class, using AutoService for meta-inf entries.

## how you tested the change

Existing integration tests. No functional change... wait should probably verify we can get the object. will add the test in a bit.

## Where the following done:

### DIFF
--- a/java/cwms-implementation/src/main/java/org/opendcs/cwms/CwmsSettings.java
+++ b/java/cwms-implementation/src/main/java/org/opendcs/cwms/CwmsSettings.java
@@ -1,0 +1,36 @@
+package org.opendcs.cwms;
+
+import java.util.Properties;
+
+import org.opendcs.settings.api.OpenDcsSettings;
+
+import decodes.util.PropertiesOwner;
+import decodes.util.PropertySpec;
+
+/**
+ * Hold any CWMS Specific operational settings.
+ * 
+ * NOTE: Inentionally empty. Placehodler for future work.
+ */
+public final class CwmsSettings implements PropertiesOwner, OpenDcsSettings
+{
+
+    @SuppressWarnings("java:S1172") // future work
+    public CwmsSettings(Properties props)
+    {
+        // Future work.
+    }
+
+    @Override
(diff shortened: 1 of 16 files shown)


java/cwms-implementation/src/main/java/org/opendcs/cwms/CwmsSettings.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.opendcs.settings.api.OpenDcsSettings;
+
+import decodes.util.PropertiesOwner;
+import decodes.util.PropertySpec;
+
+/**
+ * Hold any CWMS Specific operational settings.
+ * 
+ * NOTE: Inentionally empty. Placehodler for future work.
```
ktarbet:
```suggestion
 * NOTE: Intentionally empty. Placeholder for future work.
```

java/cwms-implementation/src/main/java/org/opendcs/cwms/CwmsSettings.java
ktarbet, line 1: What would a Cwms Specific setting be?     OfficeID ?

ktarbet (dismissed):

java/cwms-implementation/src/main/java/org/opendcs/cwms/CwmsSettings.java
MikeNeilson: Maybe, this is more "system wide" settings. I considering adding it in for testing, but it already exists in DecodesSettings and what to think about it a it.

Will likely do something along the lines of "<office>.<property>" to handle an office specific override so it's available at usage. But now wasn't the time to sort that out.

ktarbet (dismissed):

ktarbet (approved):

PR #2017, opened 2026-06-11
Fix Loading Apps status column

tlelv:
## Problem Description

Showing inactive for running processes

## Solution

Describe your solution.

The lock query and the app-name lookup were wrapped in a single try-catch. If the name lookup failed, the catch block discarded all valid lock data silently, returning an empty list. I think this caused every process to appear inactive regardless of actual state. Fixed by splitting into two independent try-catches 

## how you tested the change

AppResourcesTest (4 unit tests) — all pass
Frontend Vitest suite (14 tests across 3 files including LoadingApps) — all pass
TypeScript type check (tsc --noEmit) — no errors
opendcs:compileJava — compiles clean

## Where the following done:


MikeNeilson (dismissed):

ktarbet (dismissed):
I tested manually.  Looks good!

As part of testing I forced a row  (cp-comp_proc_lock) to be active as shown below:

Image

MikeNeilson (approved):

PR #2044, opened 2026-06-18
Presentation Groups Adding Presentation Element Bugfix

### LINKED ISSUE #2031: Web UI - Presentation Groups - Adding Presentation Element deleted my existing row and no new rows added
**Describe the bug**
When attempting to add a row to the Presentation Elements inside a Presentation Group, it appeared to add an empty row to fill out.  Once typing in information and clicking save, both my new row and the existing row disappeared.  Now there are no rows in the group, and I cannot add new ones.

**To Reproduce**
1. Open Web UI Presentation Groups page
2. Edit a presentation group
3. Attempt to add a presentation element
4. See rows disappear on save

**Expected behavior**
Expected a second row to save successfully, or an error to pop-up if I had an issue with what I typed in.

tlelv:
## Problem Description

Fixes #2031 

## Solution

Describe your solution.

From backend:
Null elements list would NPE before processing.
Unknown dataTypeStd/dataTypeCode gave a silent 500

Frontend:
Save/cancel acted on wrong state
save/cancel was unreachable sometimes due to DataTables' "responsive: true"
savePresentation sent an invalid Date for lastModified, i think causing the backend to wipe the element list on save

## how you tested the change
[Issue2031Fix.zip](https://github.com/user-attachments/files/29071341/Issue2031Fix.zip)

## Where the following done:


java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PresentationResources.java
```diff
@@ -295,13 +297,15 @@ static Vector<DataPresentation> map(DataTransaction tx, DataTypeDao dtDao, List<
         {
             DataPresentation dataPres = new DataPresentation();
             dataPres.setUnitsAbbr(ape.getUnits());
-            final var dt = dtDao.lookup(tx, ape.getDataTypeStd(), ape.getDataTypeCode())
-                                .orElseGet(() -> new DataType(ape.getDataTypeStd(), ape.getDataTypeCode()));
+            final var existingDt = dtDao.lookup(tx, ape.getDataTypeStd(), ape.getDataTypeCode());
+            final var dt = existingDt.isPresent()
+                    ? existingDt.get()
```
MikeNeilson: This would be cleaner if kept in the `.orElseGet(() -> dtDao.save(...` syntax.

Functionally equivalent, less lines of code.

MikeNeilson (changes requested):

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PresentationResources.java
MikeNeilson: unused import according to sonarcloud.

java/opendcs/src/main/java/org/opendcs/database/dai/DataTypeDao.java
```diff
@@ -59,6 +59,12 @@ public interface DataTypeDao extends OpenDcsDao
       */
      Optional<DataType> lookup(DataTransaction tx, String standard, String dataTypeCode) throws OpenDcsDataException;
 
+    default DataType lookupOrCreate(DataTransaction tx, String standard, String code) throws OpenDcsDataException
```
MikeNeilson, line 62: that works too. though I'd still use the Option.orElseGet method, but that's just me.

MikeNeilson (changes requested):

java/opendcs/src/main/java/org/opendcs/database/dai/DataTypeDao.java
tlelv: I thought orElseGet wouldnt work since it uses Supplier<T>, which wont throw checked exceptions and save declares throws OpenDcsDataException.

MikeNeilson: ... oh, good point. always forget out those checked exceptions in those handlers.

MikeNeilson (dismissed):

MikeNeilson (approved):

PR #2053, opened 2026-06-26
Run Computation Button

### LINKED ISSUE #2037: Web UI - Computations - No "Run Computations" button
**Describe the bug**
No `Run Computations` button is available on the new Web UI.  A way to test computations is something that is critical before enabling them so that an entire process doesn't crash from a user error.

**To Reproduce**
1. Open Web UI Computations page
2. Open or Edit a computation
3. See there is no button

**Expected behavior**
Expected the button to be there.

tlelv:
## Problem Description

Fixes #2037. 

## Solution

Describe your solution.

Adds a Run Computation button (▶) to each row in the Computations table. Clicking it opens a modal where the user selects a start/end time window and runs the computation against the live database.

How it works:
* Streams output from the backend via Server-Sent Events and implemented with fetch + ReadableStream.
* Fetches the computed output values for each output time series that has a valid database key.
* Displays a pivot table of the results (timestamps × output series, with storage units in column headers)
* Falls back to a "no values written in this time window" message when the output time series exist but have no data in the requested window

Screenshot 2026-06-26 at 12 47 48 PM
Screenshot 2026-06-26 at 12 47 55 PM
Screenshot 2026-06-26 at 12 48 01 PM

## how you tested the change

Ran the computation against the live Docker backend and confirmed SSE log messages stream in real-time
Verified the "no values written" fallback displays correctly for a computation with unresolved output parameters (..null. identifier / #trigs=0)

(Need help verifying a successful timeseries with the real backend)

## Where the following done:

### DIFF
--- a/javascript/opendcs-web-ui-react/public/locales/de-DE/computations.json
+++ b/javascript/opendcs-web-ui-react/public/locales/de-DE/computations.json
@@ -81,5 +81,22 @@
     "none_found": "No algorithms available.",
     "add_selected": "Add {{count}} Computation(s)",
     "adding": "Adding..."
+  },
+  "run": {
+    "title": "Berechnung ausführen: {{name}}",
+    "start": "Start",
+    "end": "Ende",
+    "run": "Ausführen",
+    "run_for": "Berechnung {{id}} ausführen",
+    "stop": "Stoppen",
+    "running": "Wird ausgeführt...",
+    "aborted": "Ausführung abgebrochen.",
+    "log_label": "Berechnungsprotokoll",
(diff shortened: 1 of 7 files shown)


MikeNeilson (approved):
This doesn't appear to handle group computations. You have to select which input(s) will be used for the executation.

E.g. if you have a stage/flow computation and the group is state, you have to select the individual time series to execute (desktop version is multi select, then sequential, doing parallel would be nice, but browser limits will be an issue.)

However, that should be handled in a follow up. This is a good start to testing a computation with the web ui.

PR #2054, opened 2026-06-27
Fix Copy/Create new computation

### LINKED ISSUE #2036: Web UI - Computations - Unable to Copy or Create a new computation
**Describe the bug**
Using the Copy button or the + icon to add a new computation disappears after clicking Save.  No new computation is created and no error is thrown.

**To Reproduce**
1. Open Web UI
2. Open Computations page
3. Create a new computation (Copy or Add)
4. Save, see row disappear

**Expected behavior**
Expected a new computation to be created or an error to pop up telling the user why the save failed.

tlelv:
## Problem Description

Fixes #2036. 

## Solution

Describe your solution.
I believed AppDataTable's detail-save handler caught all errors from `onSave` and had removed the local draft item from the list.

AppDataTable.tsx re-throws the error when the failing row is a new
local item. New drafts stay in edit mode so the user
doesn't lose their work.

saveComputation now awaits actions.save and
catches the re-thrown error. 

## how you tested the change

Able to add a new computation and it populates the list. 
The copy button copies an existing computation and adds a new entry, leaving just the name slot open. Adding from algorithm also worked. 

## Where the following done:


MikeNeilson (dismissed):

tlelv:
added fix for #2060

MikeNeilson (approved):

PR #2056, opened 2026-06-29
Run Group Computations

tlelv:
## Problem Description

Could not select time series when running a group computation from the web UI. Let alone the group area saving properly. Unable to choose which inputs to execute the computation against. 

I believe this came from: the /computationrefs endpoint never included groupId or groupName in its response (DTOMappers.mapRef did not populate those fields). The Run button therefore always received undefined as the group ID, regardless of whether the computation had a group assigned.

## Solution

Describe your solution.

mapRef(DbComputation) now populates groupId and groupName on the returned ApiComputationRef when the computation has a group assigned). When no group is assigned the fields are left null, so the frontend correctly treats the computation as non-group.

## how you tested the change

* Assigned a TS group to an existing computation and saved it
* Confirmed the Run modal now shows the Time Series checklist
* Clicked Run on a computation with and without a group

Group computation run:
image

Single computation run:
image

## Where the following done:

### DIFF
--- a/java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/TimeSeriesResources.java
+++ b/java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/TimeSeriesResources.java
@@ -764,6 +764,56 @@ public final class TimeSeriesResources extends OpenDcsResource
 		return ret;
 	}
 
+	@GET
+	@Path("expandgroup")
+	@Produces(MediaType.APPLICATION_JSON)
+	@RolesAllowed({ApiConstants.ODCS_API_USER, ApiConstants.ODCS_API_ADMIN})
+	@Operation(
+			summary = "Expand a time series group to its full list of member time series identifiers.",
+			description = "Returns the fully-expanded list of time series identifiers that belong to the "
+					+ "specified group, including members from sub-groups, attribute-based filters, "
+					+ "and include/exclude/intersect relationships.  \n\n"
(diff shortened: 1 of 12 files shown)


tlelv:
Adding fix for #2061

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/TimeSeriesResources.java
```diff
@@ -764,6 +769,123 @@ static ApiTsGroup map(TsGroup group)
 		return ret;
 	}
 
+	@GET
+	@Path("expandgroup")
+	@Produces(MediaType.APPLICATION_JSON)
+	@RolesAllowed({ApiConstants.ODCS_API_USER, ApiConstants.ODCS_API_ADMIN})
+	@Operation(
+			summary = "Expand a time series group to its full list of member time series identifiers.",
+			description = "Returns the fully-expanded list of time series identifiers that belong to the "
```
MikeNeilson: we're on java 21, listen to Sonar and use the """ text """ technique, way easier than all the String concats

MikeNeilson (changes requested):
Otherwise looks good. but we should start using the """ java strings in those description fields.

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/TimeSeriesResources.java
tlelv: ah okay you're very right it is a lot cleaner

MikeNeilson (approved):

PR #2066, opened 2026-06-30
Behavior when deleting

### LINKED ISSUE #2026: Web UI - Behavior when deleting
**Describe the bug**
When deleting a config, some delete successfully and others do not work.  There is no indication from the UI on a success or failure.  I assume the ones that deleted are the ones that weren't connected to a platform but can't verify that since they are gone now :)

**To Reproduce**
1. Open Web UI Configs Page
2. Click the trashcan next to several configs
3. See that some delete and some don't with no indicator

**Expected behavior**
Expect that on a delete, the user will see a pop-up indicating a successful deletion or a failure with a reason why.

**Additional context**
Related to https://github.com/opendcs/opendcs/issues/1598.

tlelv:
## Problem Description

Fixes #2026. 

Deleting in the Web UI gave no indication of success or failure. Some deletes silently succeeded or silently failed (like an algorithm in use by a computation), and there was no confirmation prompt before a big destructive action.

## Solution

* Constraint violations return a 409 instead of a silent failure or 500
* SqlConstraintTranslator: classifies the SQLException and wraps it into OpenDcsDataConstraintException. Wired into the delete paths of the Config, Site, and Loading App DAOs.
* ToastBus + ToastStack component wired into TanStack Query's MutationCache via mutationToasts.ts. Failures show the API's Status.message in a red toast and successes show a green toast.
* New confirmation modal in AppDataTable: clicking delete now prompts "Are you sure?" for Configs, Sites, Network Lists, Loading Apps, Algorithms, Computations, and Schedule Entries.

## how you tested the change

* Ran the full stack locally and verified on all of the 7 wired tables that delete now shows a confirm dialog and a success/error toast.

## Where the following done:


java/opendcs-api/src/main/java/org/opendcs/database/api/OpenDcsDataConstraintException.java
```diff
@@ -0,0 +1,34 @@
...
+* the License.
+*/
+package org.opendcs.database.api;
+
+/**
+ * Thrown when a delete or save is rejected because the record is still
+ * referenced by another record in the database (foreign-key constraint
+ * violation).  Maps to HTTP 409 Conflict at the REST layer.
+ */
+public class OpenDcsDataConstraintException extends OpenDcsDataException
```
MikeNeilson: already exits here: org.opendcs.database.api.exceptions.data.OpenDcsContraintException, along with some specific exception under it.

java/opendcs-implementation/src/main/java/org/opendcs/database/impl/opendcs/SqlConstraintTranslator.java
MikeNeilson, line 1: See the jdbi3 SqlExceptionHandlers we already have started.

(Good logic, it just already has a place to go.)

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppExceptionMapper.java
MikeNeilson, line 1: See comment on OpenDcsDataContraintException, logic good, just needs to use the existing contraint exceptions.

java/opendcs-rest-api/src/test/java/org/opendcs/odcsapi/res/AppExceptionMapperTest.java
MikeNeilson, line 1: Good idea, I should've done this when I first made the exception we did a few weeks ago.

java/opendcs/src/main/java/opendcs/dao/AlgorithmDAO.java
```diff
@@ -453,6 +453,10 @@ public void deleteAlgorithm(DbKey id) throws DbIoException, ConstraintException
                 doModify(q, id);
             });
         }
+        catch(ConstraintException ex)
```
MikeNeilson: This ContraintException is different and follows a different change of exception, I would leave the original wrapped exception in place and add the additional "getCause()" check to the resource code.

(if only to minimize changes to the old DAOs. This is is probably fairly innocuous but it could easily get out of hand.)

MikeNeilson (changes requested):
Okay, so might fault here. I need to write up some stuff in the dev docs about how we're now dealing with Exceptions/Contraint errors/etc in the new DAOs

java/opendcs-api/src/main/java/org/opendcs/database/api/OpenDcsDataConstraintException.java
tlelv: cleaned it out

java/opendcs-rest-api/src/test/java/org/opendcs/odcsapi/res/AppExceptionMapperTest.java
tlelv: AppExceptionMapperTest coverage stays, minus the two cases that referenced the exception type I removed

java/opendcs/src/main/java/opendcs/dao/AlgorithmDAO.java
tlelv: switched it back to its original behavior

java/opendcs-implementation/src/main/java/org/opendcs/database/impl/opendcs/dao/DecodesConfigDaoImpl.java
```diff
@@ -441,10 +441,13 @@ public void delete(DataTransaction tx, DbKey id) throws OpenDcsDataException
...
         }
+        catch (OpenDcsConstraintException ex)
+        {
+            throw ex;
+        }
         catch (RuntimeException ex)
         {
-            var engine = tx.getContext().getDatabaseEngine();
-            throw SqlConstraintTranslator.translate("Config " + id + " is still used by one or more platforms and cannot be deleted", engine, ex);
+            throw new OpenDcsDataException("Config " + id + " is still used by one or more platforms and cannot be deleted", ex);
```
MikeNeilson: So, something is wrong here if the contraint exception above isn't capturing the condition above. Additionally we don't want to assume any Runtime exception would be this.

java/opendcs/src/main/java/org/opendcs/database/impl/opendcs/jdbi/plugins/OpenDcsConstraintSqlExceptionHandler.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public OpenDcsConstraintSqlExceptionHandler(DatabaseEngine engine)
+    {
+        this.engine = engine;
+    }
+
+    @Override
+    public void handle(SQLException ex, StatementContext ctx)
+    {
+        String sqlState = ex.getSQLState();
```
MikeNeilson, line 48: These generally need to be per implementation that way we aren't encoding too much.

Postgres codes are different from Oracle, and CWMS uses some user defined error codes in place of the standard ones.

There will be some duplication. That's where a class like the SqlTranslator comes in to lump in some of that logic. ALternatively, just stuff it in the existing ones, possible making a different OpenDcsPostgres and OpenDcsOracle implementation.


There is a specific Cwms Implementation. Though the generic one *should* work for any OpenDCs tables, it's the CWMS tables/data that would likely throw different ones.

MikeNeilson (changes requested):
Closer. FYI (the general logic and ideas are good, I'm just really nitpicking where exactly they go.)

Read through the existing handlers and how they're setup and go from there. They aren't complete but are intended to be filled in.

PR #2067, opened 2026-07-01
Sites location info

### LINKED ISSUE #2024: Web UI - Sites - Location Information
**Feature Requests.**
1. No specified format on Latitude/Longitude can cause confusion. Recommend "Latitude (DD)" and "Longitude (DD)". This is currently not specified in desktop client either.
2. Alphabetize the drop-down for Elevation Units (see screenshot below).
3. Make State a dropdown menu to constrain it.  Currently not a dropdown in the desktop client but is constrained in the CWMS-Vue Locations tab
4. Is Region supposed to be the same as County in the CWMS-Vue Locations tab?  Some of the filled-out Regions look similar to Counties that haven't been specified yet in CWMS-Vue in the NWDM database.  However, on our local T7 Region is blank when the County is specified for BIL in CWMS-Vue.  Maybe a tooltip would be nice here or renaming it to County if it's supposed to match.
5. Country is named the same as it is in the desktop client but it would be nice to match that to the CWMS-Vue Locations tab, too, to make sure people know those are the same property.  Also, this is constrained to a dropdown menu in CWMS-Vue which would be nice.

**Additional context**
Screenshot of Elev units dropdown:
Image

tlelv:
## Problem Description

Fixes #2024. 

## Solution

Describe your solution.

* Reorganizing for elevation units.
* Fixed the condition controlling whether the Actions column renders. 
* fixed where DataTables visibleRows.every() used without a proper type could cause subtle binding bugs.

## how you tested the change
image

image

## Where the following done:


javascript/opendcs-web-ui-react/src/pages/sites/SiteNameList.tsx
```diff
@@ -102,6 +102,7 @@ export const SiteNameList: React.FC<SiteNameListProperties> = ({
         return data;
       }
     },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
MikeNeilson: are we sure the linter isn't correct? If it's not, we should explain why.

tlelv: You're right, the linter was right here. I fixed it properly instead of just satisfying the rule.

MikeNeilson (approved):

PR #2069, opened 2026-07-06
Bugfix/add presentation element

tlelv:
## Problem Description

Re-Fixes #2044. 

## Solution

Describe your solution.
Cause of the reopened bug: JdbiTransaction/SimpleTransaction.close() used to unconditionally commit. PresentationGroupDaoImpl does delete-then-insert in one transaction when saving a group's elements. If the insert failed after the delete succeeded, the
  exception went to close().

Fix was that close() now rolls back unless commit() was explicitly called.

## how you tested the change

## Where the following done:


java/opendcs/src/main/java/org/opendcs/database/SimpleTransaction.java
```diff
@@ -60,6 +60,17 @@ public void close() throws OpenDcsDataException
     {
         try
         {
+            try
```
MikeNeilson: Doing this is Java is terrible idea. Throwing an exception incurs a rather large performance penalty and in theory we're going to be in the "happy path" far more often than not. Making that penalty add up a lot over time.

The original intention behind this design was that barring error, commit and close.
The commit it self will throw and error, in which case rollback should then be called.


The change to manually call commit in all usages is also just... awkward, especially if we want it to be the default.


It would appear that part of the issue is that the Jdbi commit failure itself is not getting propagated correctly.

This is the Transaction code we are most likely using: https://github.com/jdbi/jdbi/blob/78c2fa4012633b86190a9227a4f05cb7268f241a/core/src/main/java/org/jdbi/v3/core/transaction/LocalTransactionHandler.java#L136 which as you can see already rollsback on error.

TransactionError extends JdbiException which extends RuntimeException. Most likely we just want to capture that locally so we can propagate the error up correctly.

MikeNeilson (changes requested):

java/opendcs/src/main/java/org/opendcs/database/SimpleTransaction.java
tlelv: Oh duh, sorry to include the bad practice. Let me try

MikeNeilson: definitely not a "duh" moment. Using try for code flow is rather normal in some languages (especially python). And the performance hit isn't super obvious... at least until you try to process a few thousand elements and see it being fairly obvious in the measurements.

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/EquipmentResources.java
```diff
@@ -171,8 +187,8 @@ public Response postEquipment(ApiEquipmentModel apiEquipmentModel) throws WebApp
         }
         catch (OpenDcsDataException ex)
         {
-            throw new WebAppException(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(),
-                                      "Unable to save equipment model.", ex);
+            rollbackQuietly(tx, ex);
```
MikeNeilson: instead of these nested methods, perhaps in the main handler of each, create the web app exception then call rollbackQuietly? and throw the created exception.

Just seems odd to have a "doX" method for "X" which is already a method.

MikeNeilson (commented):
@oskarhurst thoughts?

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/EquipmentResources.java
MikeNeilson: The JdbiTransactionTest is also a good place to create some various assertions and force behavior to see how things look.

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/OpenDcsResource.java
```diff
@@ -91,4 +93,22 @@ protected TimeSeriesDb getLegacyTimeseriesDB()
...
 				.orElseThrow(() -> new UnsupportedOperationException(UNSUPPORTED_OPERATION_MESSAGE));
 	}
+
+	/**
+	 * Rolls back a transaction after a failure, preserving the original exception rather than
+	 * letting a rollback failure replace it. Call from the catch block that handles a failed
+	 * commit/DAO call within a {@code try (var tx = db.newTransaction())}; the caller is still
+	 * responsible for rethrowing {@code cause} afterward.
+	 */
+	protected static void rollbackQuietly(DataTransaction tx, Exception cause)
```
MikeNeilson, line 106: Something about this was bugging but and so I went digging, It appears that, according to https://docs.oracle.com/javase/specs/jls/se17/html/jls-14.html#jls-14.20.3.1, java already does this for us.

java/opendcs/src/main/java/org/opendcs/database/JdbiTransaction.java
```diff
@@ -48,7 +49,14 @@ public void commit() throws OpenDcsDataException
     {
         if (jdbiHandle.isInTransaction())
         {
-            jdbiHandle.commit();
+            try
+            {
+                jdbiHandle.commit();
+            }
+            catch (TransactionException ex)
+            {
```
MikeNeilson, line 57: I think if we call rollback here, it'll start doing what we expect. Though we may want to create an internal handler that also takes an exception in so we can add the original as a suppress exception or something.

See https://stackoverflow.com/questions/51736424/why-do-i-have-to-use-rollback-in-jdbc for some more discussion. But appearently that is on the application, and we would prefer to have it handled more naturally than having tx.commit and tx.rollback every time it's used. (Some sort of wrapper helper could also deal with that, but I think the `JdbiTransaction`, and also `SimpleTransaction` can be responsible for that behavior.

integrationtesting/fixtures/src/main/java/org/opendcs/fixtures/lrgs/LrgsTestInstance.java
```diff
@@ -12,6 +12,8 @@
 import org.opendcs.tls.TlsMode;
 import org.opendcs.utils.logging.LoggingEventBuffer;
 
+import ilex.util.PasswordFile;
```
MikeNeilson, line 16: These changes were merged in and shouldn't be showing up again in a PR.

You may need to sync your fork and rebase... or something more drastic (hopefully not)

MikeNeilson (changes requested):
Your PR has gotten into an awkward state and is showing changes that are already merged in, which is making it difficult to see you're actual changes.

integrationtesting/fixtures/src/main/java/org/opendcs/fixtures/lrgs/LrgsTestInstance.java
tlelv: Ah shoot okay let me try and rebase things

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/sec/user/UserResources.java
```diff
@@ -75,38 +75,46 @@ public Response updatePassword(ApiPasswordChange passwordChange) throws WebAppEx
...
+                {
+                    throw new WebAppException(Response.Status.FORBIDDEN.getStatusCode(), "Unable to update password");
+                }
             }
-            else
+            catch (OpenDcsDataException | OpenDcsAuthException ex)
             {
-                throw new WebAppException(Response.Status.FORBIDDEN.getStatusCode(), "Unable to update password");
+                final var wex = new WebAppException(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(), "Unable to process password change.", ex);
+                rollbackQuietly(tx, wex);
```
MikeNeilson, line 109: not entirely sure we want to keep that rollbackQuietly here, but this is definitely a bit more nested than the tx.wrap would likely allow. I'll try to dig into it a big next week.

MikeNeilson: I haven't had time to look, so don't worry about it. It can be address later.

MikeNeilson (approved):

PR #2072, opened 2026-07-07
initial work on handling Jdbi/SQL errors.

MikeNeilson:
@tlelv and I have been going back and forth on #2069, and how we handle certain errors.

There's a lot of boiler plate that's been added that I'd like to avoid if at all possible.

So I attempted to boil down the issue to to a single test that forces a constraint error (generally our most common issue) so we can focus on the transaction handler and immediate logic vs making larger changes to every usage of the new DAOs.


MikeNeilson:
NOTE: Test (PresentationResourceIT delete) is intentionally failing, I set the expected code to something that *should* expect but currently does not. Thought is to map to a specific  (OpenDcsConstraintException... note: not-yet-implemented) and then from that with a proper handler to the appropriate HTTP code.


At this point though, thoughts on proposed design/behavior/boilerplate changes?

@opendcs/opendcs-core-devs 

But I think especially @adamkorynta , you tend to have an good eye/brain for these things.

MikeNeilson:
The handling could likely also be cleaned up by just declaring a throws for OpenDcsDataException and including it in the JAX-RS exception handler. Will expirement with that in followup PRs.

Additional note: I have intentionally *not* mapped every SQL Error, and the foreign key constraint doesn't yet distinguish between current data is used, or target data doesn't exist. We will add these over time so as not to overload this PR.

java/opendcs-api/src/main/java/org/opendcs/database/api/OpenDcsDataException.java
```diff
@@ -22,6 +22,11 @@ public OpenDcsDataException(String msg)
...
     }
 
+    public OpenDcsDataException(Throwable cause)
+    {
+        super(cause);
+    }
+
     public OpenDcsDataException(String msg, Throwable cause)
     {
         super(msg);
```
ktarbet:
```suggestion
        super(msg, cause);
```

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppExceptionMapper.java
```diff
@@ -37,147 +42,163 @@
...
+      Status status = new Status("Bad Request.  There was an issue with the request, please try again or contact your system administrator.");
+      return Response.status(Response.Status.INTERNAL_SERVER_ERROR)
+            .entity(status)
+            .type(MediaType.APPLICATION_JSON)
+            .build();
+   }
+
+   private static Response handle(OpenDcsConstraintException ex)
+   {
+      String msg = "Cannot peform operation cause ";
```
ktarbet:
```suggestion
      String msg = "Cannot perform operation because ";
```

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppExceptionMapper.java
```diff
@@ -37,147 +42,163 @@
...
+      switch (ex)
+      {
+         case RelatedDataConstraintException related -> msg = msg + " data is use.";
```
ktarbet:
```suggestion
         case RelatedDataConstraintException related -> msg = msg + " data in use.";
```

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppExceptionMapper.java
```diff
@@ -37,147 +42,163 @@
...
+         case UniqueConstraintViolationException unique -> msg = msg + " data already exists.";
+         case null, default -> msg = msg + " cannot perform operation due to constraint.";
+      }
+
```
ktarbet:
```suggestion
log.atInfo().setCause(ex).log(msg);
```

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppExceptionMapper.java
```diff
@@ -37,147 +42,163 @@
...
+      return Response.status(Response.Status.CONFLICT)
+                     .entity(msg)
+                  .build();
+   }
```
ktarbet, line 92:
```suggestion
      return Response.status(Response.Status.CONFLICT)
                   .entity(new Status(msg))
                   .type(MediaType.APPLICATION_JSON)
                  .build();
   }
```

java/opendcs/src/test/java/org/opendcs/database/JdbiTransactionTest.java
```diff
@@ -49,6 +56,16 @@ public DatabaseEngine getDatabaseEngine()
     void create_db()
     {
         jdbi = Jdbi.create("jdbc:derby:memory:db;create=true");
+        // Idea is to move this to the Wrapper Implementations, default to this
+        // allow/exepect implementations to refine. Deriving things like
```
ktarbet:
```suggestion
        // allow/expect implementations to refine. Deriving things like
```

java/opendcs/src/test/java/org/opendcs/database/JdbiTransactionTest.java
```diff
@@ -49,6 +56,16 @@ public DatabaseEngine getDatabaseEngine()
...
+        // constraint errors from Runtime exceptoin and passing them on.
```
ktarbet:
```suggestion
        // constraint errors from Runtime exception and passing them on.
```

ktarbet (dismissed):

ktarbet (approved):

PR #2083, opened 2026-07-14
Move PDT load attempt to be performed once. Reduce wait.

### LINKED ISSUE #1986: Missing  PDTs_compress.txt file causing excessive retries and tanking performance.
This may be related to the issue reported in https://github.com/opendcs/opendcs/issues/255

We recently updated our environment from running OPENDCS 6.2 RC04 to OpenDCS 7.0.16 RC16.

Since the upgrade, we have observed that when we call rs from terminal it is taking a long time to run. 

When I call opendcs 6 with this configuration, the file is decoded and accessable within a minute, when I use opendcs 7 it is taking over 45 minutes before timing out. 

We are using the output format of hydromet dms3, I have observed using other formats causes this to operate as expected.

We have both opendcs 6 and opendcs 7 installed on the server. 

I have run it in both opendcs 6 and opendcs 7 with the exact same configuration pointing to a single site and a single 24 hour period. 

I have also observed that opendcs7 is adding a GPS to the decode that is not defined within the decoding commands. 

I am running the command
rs -d3 -R -l test.log testdecodes

I am not sure if this should be submitted as a bug report, so putting it in as a blank issue.

MikeNeilson:
## Problem Description

Fixes #1986. 

## Solution

Move Pdt load attempt from format to constructor, reduce wait time.

## how you tested the change

- TODO. no tests setup of any formatter, should be reasonably straight forward and will work on later.

## Where the following done:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,17 +26,17 @@ jobs:
           sudo rm -rf /usr/share/dotnet
           sudo rm -rf /usr/local/lib/android
           sudo rm -rf /opt/ghc
-      - uses: actions/checkout@v4.1.1
+      - uses: actions/checkout@v6.0.2
       - name: Set up JDK
-        uses: actions/setup-java@v4.4.0
+        uses: actions/setup-java@v5.2.0
         with:
           java-version: ${{matrix.jdk}}
           distribution: temurin
-      - uses: actions/setup-python@v5.1.0
+      - uses: actions/setup-python@v6.2.0
         with:
           python-version: "3.8"
       - name: Setup Gradle
-        uses: gradle/actions/setup-gradle@v4.1.0
+        uses: gradle/actions/setup-gradle@v6.0.1
         with:
           dependency-graph: generate-and-submit
       - name: Change temp dir in windows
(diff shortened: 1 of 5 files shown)


ktarbet (dismissed):

MikeNeilson:
The windows tests are failing due to an unrelated issue with how certificates are handled in the LRGS tests. Will fix in a dedicated follow up.

agilmore2:
That's funny. Pretty sure the original issue was unrecognized certs
blocking the PDT downloads.

MikeNeilson:
> That's funny. Pretty sure the original issue was unrecognized certs blocking the PDT downloads.


Well, I stand by this fix in general, but will go look at that.

MikeNeilson:
Well, good thing I looked, it was the new test failing. Issue with with the test setup (\n is always used with the given Consumer) and the test was setup to use the system line separator.

If the tests pass I will merge in and make the final 7.0.17 release with this change in.

MikeNeilson:
Fixing the sonar analysis is very likely to be a rather obnoxious effort... that may well be worth it, just not for this PR at this time.

PR #2097, opened 2026-07-27
Strip out the vast majority of the Swing UI. Remaining elements are s…

MikeNeilson:
…till required, at least partially, by the LRGS UIs.

## Problem Description

Fixes #2096 (mostly). 

Describe the problem you are trying to solve.

## Solution

Remove bulk of Swing CODE.
Some elements were still used by the LRGS UI elements that have no "web" replacement at this time.

## how you tested the change

Existing integration tests. LRGS components were not verified to still operate during this PR.
May go do so later.

## Where the following done:

- [ ] Tests. Check all that apply:
   - [ ] Unit tests created or modified that run during ant test.
   - [ ] Integration tests created or modified that run during integration testing
         (Formerly called regression tests.)
   - [ ] Test procedure descriptions for manual testing
- [ ] Was relevant documentation updated?
- [ ] Were relevant config element (e.g. XML data) updated as appropriate

If you aren't sure leave unchecked and we will help guide you to want needs changing where.

PR #2098, opened 2026-07-27
Feature/inflow calc nwp integration tests

### LINKED ISSUE #1044: Add support for NWP's Inflow calculations
@ArthurArmour has requested that an algorithm be created to support the inflow calculations performed at NWP.

Need some more feedback from Art to determine all of the requirements. Creating this ticket now in order to not lose the request.

N/A

**Additional context**
Depending on the complexity, the work may be accomplished under the project: https://github.com/orgs/opendcs/projects/39

rma-kayla:
## Problem Description

Fixes #1044 

## Solution

Updates to inflow estimation algorithm to incorporate integration tests (#1044)
 
 - reworked input variable typing
 - fixed tailwater tsid variable mistmatch
 - added getTimeSeries and setMissingActionIgnore to allow missing/optional values without throwing null errors.
 - added matchesInflowIntervalOffset to fix UTC offset errors due to missing initial setup
 - fixed writing each 5 minute increment and only allowing the 15 minute interval to write to output
 - added non DB validation unit tests
 - minor cleanup to bypass logging errors and allow scroll behaviors in edit comp window

## how you tested the change

## Where the following done:

   - [x ] Unit tests created or modified that run during ant test.
   - [x ] Integration tests created or modified that run during integration testing
- [x ] Was relevant documentation updated?
- [x ] Were relevant config element (e.g. XML data) updated as appropriate


java/opendcs/src/main/java/decodes/tsdb/algoedit/AlgorithmWizard.java
```diff
@@ -890,6 +890,7 @@ public void runApp()
 
 		TopFrame algowizFrame = initTopFrame();
 		algowizFrame.setVisible(true);
+		algowizFrame.toFront();
```
MikeNeilson: FYI, this class was recently deleted ([or at least will be soon.](https://github.com/opendcs/opendcs/pull/2097))

java/opendcs/src/main/java/org/opendcs/algorithms/TimeSeriesUtil.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Date;
+
+import decodes.tsdb.BadTimeSeriesException;
+import decodes.tsdb.CTimeSeries;
+import decodes.tsdb.DbCompException;
+import decodes.tsdb.DbIoException;
+import decodes.tsdb.IntervalCodes;
+import opendcs.dai.TimeSeriesDAI;
+
+public final class TimeSeriesUtil
```
MikeNeilson: I believe there is an existing Time Series Utility class somewhere.

I will grant I don't remember the exact name, or if I'm thinking of test code.

@oskarhurst may remember.

java/opendcs/src/main/java/decodes/cwms/algo/InflowEstimationAlgo.java
```diff
@@ -0,0 +1,505 @@
...
+	@Override
+	protected void beforeTimeSlices() throws DbCompException
+	{
+		log.atDebug().log("InflowEstimationAlgo::beforeTimeSlices");
+	}
+
+	@Override
+	protected void doAWTimeSlice() throws DbCompException
+	{
+		log.atDebug().log("InflowEstimationAlgo::doAWTimeSlice");
```
MikeNeilson: Don't log these occurrences.

Or at least do it at Trace.

java/opendcs/src/main/java/decodes/cwms/algo/InflowEstimationAlgo.java
```diff
@@ -0,0 +1,505 @@
...
+		catch(NotEnoughDataException ex)
+		{
+			log.atDebug().setCause(ex).log("Not enough data found to perform inflow calculation");
+		}
+	}
+
+	@Override
+	public void afterAllTimeSlices()
+	{
+		tsdb.freeConnection(conn);
```
MikeNeilson:
```suggestion
		conn.close();
```


The connection object returned by getConnection will be a WrappedConnection that correctly handles if the instances needs to be closed or now.

the SQLException will need to be handled locally.

java/opendcs/src/main/java/decodes/cwms/algo/IntervalOffsetUtil.java
```diff
@@ -0,0 +1,176 @@
...
+import java.util.Date;
+import java.util.TimeZone;
+
+import decodes.cwms.CwmsTsId;
+import decodes.tsdb.CTimeSeries;
+import decodes.tsdb.IntervalCodes;
+import decodes.tsdb.TimeSeriesIdentifier;
+import opendcs.opentsdb.Interval;
+
+final class IntervalOffsetUtil
```
MikeNeilson: This seems like it should be in either TimeSeriesUtil, the existing Intervals class,
 or least not specific to the CWMS namespace.

java/opendcs/src/main/java/decodes/cwms/algo/InflowEstimationAlgo.java
```diff
@@ -0,0 +1,505 @@
...
+		ParmRef outflow2 = getParmRef("outflow_2");
+		ParmRef outflow3 = getParmRef("outflow_3");
+		ParmRef outflow4 = getParmRef("outflow_4");
+		ParmRef outflow5 = getParmRef("outflow_5");
+		ParmRef outflow6 = getParmRef("outflow_6");
+		ParmRef releaseTsid = getParmRef("release_tsid");
+		ParmRef tailwaterTsid = getParmRef("tailwater_tsid");
+		ParmRef stagePoolTsid = getParmRef("stage_pool_tsid");
+		ParmRef storageTsid = getParmRef("storage_tsid");
+		outflow1.setMissingAction(MissingAction.IGNORE);
```
MikeNeilson: this should get set by the user configuring the computation, not a validation method. Or even the comp itself; perhaps with an exception for an algorithm were a given input is always required or the math is invalid.

java/opendcs/src/main/java/org/opendcs/algorithms/TimeSeriesUtil.java
MikeNeilson: Just remembered, it's TSUtil, same one used to do the unitConversion in the algorithm.

Please move this method there.

java/opendcs/src/main/java/decodes/cwms/algo/InflowEstimationAlgo.java
rma-kayla: moved missing ignore setup to the comp.xml file

java/opendcs/src/main/java/decodes/cwms/algo/InflowEstimationAlgo.java
rma-kayla: removed log

java/opendcs/src/main/java/org/opendcs/algorithms/TimeSeriesUtil.java
rma-kayla: Removed TimeSeriesUtil and utilized TSUtil instead

java/opendcs/src/main/java/decodes/tsdb/algoedit/AlgorithmWizard.java
rma-kayla: noted

java/opendcs/src/main/java/decodes/cwms/algo/InflowEstimationAlgo.java
rma-kayla: Updated to use conn.close() instead of tsdb.freeConnection(conn).
Moved the resource cleanup to alwaysAfterTimeSlices() since AW_AlgorithmBase runs it even when the compute fails and afterAllTimeSlices can get skipped.

java/opendcs/src/main/java/decodes/cwms/algo/IntervalOffsetUtil.java
rma-kayla: leaving this as is for now as part of it is CWMS specific and only this class is relying on it for now.  Will discuss with Adam

java/opendcs/src/main/java/decodes/cwms/algo/IntervalOffsetUtil.java
```diff
@@ -0,0 +1,176 @@
...
+			{
+				// In monthly value, offset is seconds since start of month.
+				// It may span a DST change.
+				if(allowDstVariation &&
+						(offsetError == -3600 || offsetError == 3600))
+				{
+					violation = false;
+				}
+
+				//TODO Consider use case where end of month is stored.
```
MikeNeilson: In these cases, you may want to consider converting the Dates to ZonedDateTimes.

The math functions behave better than the Calendar methods for these calculations.

java/opendcs/src/main/java/decodes/cwms/algo/IntervalOffsetUtil.java
MikeNeilson: The only part that I think would cause an issue is the isAllowDst flag, the UTC offset may also with HDB, but that seems like a method to include in the baseline interface and require.

I would argue that it still gets moved, but the limitations be documented, it already does "appropriate" (defensive?) instanceof chances so it doesn't crash but is still otherwise useful.

That said, I've said my piece and will defer to Adam as to whether or not it remains here. However, definitely document any implementation specific requirements in the javadoc so it's blindingly obvious to future developers, ourselves included.

MikeNeilson (commented):
Before I forget, the Copyright years are wrong, it's 2026 now.

NOTE: existing fails change to have `<year that was present> - 2026`, new fails can be `2026` or `2026 -`

rma-kayla:
I updated the property/param naming to better reflect the values.  I removed the IntervalOffsetUtil class and moved it into the TSUtil class.  I kept the CWMS specifics in the algo code so that it remains universal.  I updated the copyright years error and the algo's documentation.  I verified that integration tests pass and this PR is ready for @adamkorynta to review before merging.

java/opendcs/src/main/java/decodes/tsdb/algoedit/AlgorithmWizard.java
adamkorynta: resolved via merge with the latest codebase

java/opendcs/src/main/java/org/opendcs/gui/models/LoggingEventListModel.java
```diff
@@ -22,14 +23,27 @@ public LoggingEventListModel()
...
+        previousSize = events.size();
         final Thread t = new Thread(() -> 
         {
             while (true)
             {
                 try
                 {
-                    final int size = getSize();
-                    SwingUtilities.invokeLater(() -> this.fireIntervalAdded(this, 0, size));
+                    final int currentSize = getSize();
```
adamkorynta, line 33: Given the swing GUI is gone, this change can be removed?

adamkorynta (changes requested):

java/opendcs/src/main/java/org/opendcs/gui/models/LoggingEventListModel.java
adamkorynta: actually, the revert will still have SwingUtilities so better to not remove

java/opendcs/src/main/java/decodes/cwms/algo/IntervalOffsetUtil.java
rma-kayla: removed this class and moved code into TSUtil

java/opendcs/src/main/java/decodes/cwms/algo/IntervalOffsetUtil.java
rma-kayla: I left the function as is and kept this todo note in the TSUtil class to be resolved in a future ticket.

adamkorynta (dismissed):
I have no further feedback on the updates. Thanks, Kayla

adamkorynta (approved):

PR #2105, opened 2026-07-30
fix/Create New Routing Spec

### LINKED ISSUE #2102: Web UI - Unable to create new Routing Spec in UI
**Describe the bug**
Attempted to create a new Routing Spec to verify how some database elements were configured, was unable to do so, first got 400 then 500 error

**To Reproduce**

(Note 400 error was previously due to correct sqlKeyGenerator value not being set in the database properties and may have affected some other elements)

1. Go to routing
2. Create new
3. Fill in details
4. See error (in console, not response to the user in the GUI, edit page just stays open.)

Relevant log portion:
```
  |   | 2026-07-28 09:57:58.227 | at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103) |  
  |   | 2026-07-28 09:57:58.227 | at org.opendcs.odcsapi.res.RoutingResources.postRouting(RoutingResources.java:376) |  
  |   | 2026-07-28 09:57:58.227 | at decodes.sql.SqlDatabaseIO.writeRoutingSpec(SqlDatabaseIO.java:1569) |  
  |   | 2026-07-28 09:57:58.227 | at decodes.sql.RoutingSpecListIO.write(RoutingSpecListIO.java:531) |  
  |   | 2026-07-28 09:57:58.227 | at decodes.sql.RoutingSpecListIO.insert(RoutingSpecListIO.java:613) |  
  |   | 2026-07-28 09:57:58.227 | java.lang.NullPointerException: Cannot invoke "decodes.db.DataSource.getId()" because "rs.dataSource" is null |  
  |   | 2026-07-28 09:57:58.227 | WARNING: Unknown Error |  
  |   | 2026-07-28 09:57:56.175 | 10.7.210.143 - - [28/Jul/2026:16:57:56 +0000] "GET /odcsapi/health/live HTTP/1.1" 200 41

```

It appears when the web UI creates the ApiRouting it doesn't include the known ID (at least it should be known) of the DataSource and when the REST API maps the ApiRouting to a RoutingSpec it doesn't lookup the name and create a DataSource for the RoutingSpecListIO to use to create the entry.

**Expected behavior**
Routing Spec saves

**Screenshots**

 - OS: N/A
 - Version: Version 99.main-SNAPSHOT (445df7c)

**Additional context**

tlelv:
## Problem Description

Fixes #2102.

## Solution

Describe your solution.
The web UI only sends dataSourceName when saving a routing spec. Added RoutingResources.resolveDataSource(RoutingSpec, ApiRouting, DatabaseIO), called from postRouting after map(routing). 

## how you tested the change

## Where the following done:


java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/RoutingResources.java
```diff
@@ -386,6 +388,33 @@ public Response postRouting(ApiRouting routing)
 		}
 	}
 
+	/**
+	 * the XML DAO's routing spec parser.
```
MikeNeilson: Don't worry about the XML database in the Rest API, use the new DAO for any operations like this.

Though I think the thing to check might be why the DataSourceID isn't already present in the ApiRouting object that gets sent. The selector *should* be providing the whole ApiDataSource object which should have the id contained in it.

MikeNeilson (changes requested):
Sorry, forgot to actually submit this yesterday.

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/RoutingResources.java
tlelv: Good call, fixed DataSourceSelect and it definitely was less wiring

MikeNeilson (dismissed):

MikeNeilson (approved):

Issue #2109, opened 2026-08-02
LRGS DDS Server - Support of PROXY Protocol in


MikeNeilson:
**Is your feature request related to a problem? Please describe.**
We are starting to have instances of the LRGS behind proxies. This strips the original client IP address due to the nature of how a simple TCP proxy works.

**Describe the solution you'd like**
Implement support the HA Proxy's PROXY Protocol

**Describe alternatives you've considered**
Don't worry about it. Wait for DDS-over-HTTP. However, the protocol itself is simple enough seems reasonable.

**Additional context**
https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt

We already pull in netty (https://netty.io) as a transitive dependency, it may also be worth converting the existing implementation to use netty vs our custom client/server setup. A separate still will be opened for that.

PR #2111, opened 2026-08-03
Feature/2109 netty dds server impl

MikeNeilson:
## Problem Description

partial for #2109

## Solution

Implement baseline netty channel handlers/decoders/encoders/adapters to handle DdsServer network code

Implementation is just the basic handling, with the "server" setup still in test code. During implementation the generic handling code was the easy part; unfortunately, it was impossible to hack the original command handling code in as it depends on a Socket all through the chain.

A few tricks were attempt; however, they all proved impractical. I believe the solution is to either extract the response generation code from the LddsCommand implementations, or at least have execute return the LddsMessage vs taking the LddsThread. However, the LddsThread is used to hold several things.
It is mostly static information so we should be able to extract appropriate interfaces/etc and get the actual "send" logic to more appropriate places.

The provided implementation is intentionally small, to help get the idea across and to not have one giant PR that redesigns a dozen interactions.

NOTE: part of the purpose of going this route, beyond just general organization and modernization, is #2109, Netty already has a handler to do so, While the PROXY PROTOCOL *is* very simple, this is a good time to start cleaning up some of the older designs that are difficult to follow and maintain.

## how you tested the change

New test that uses the origin LddsClient against the new "server". 

## Where the following done:


ktarbet:
With Java 21 can we do this with built-in features (and some coding), and not have to include netty? 

ref: https://www.reddit.com/r/java/comments/1d43rbh/how_can_java_21_virtual_thread_replace_reactive/

MikeNeilson:
> With Java 21 can we do this with built-in features (and some coding), and not have to include netty?
> 
> ref: https://www.reddit.com/r/java/comments/1d43rbh/how_can_java_21_virtual_thread_replace_reactive/

No, this change has little to do with Threading itself, and more to do with how the low level socket code is organized. This gives a structure with which to ignore most of the super low level socket I/O details and focus on the DDS protocol itself while at the same time forcing better design of the client/archive handling because of that structure.

Due to the design of Netty it *do* look a lot like reactive programming, that is the style, but there's not an equivalent built in library that handles the socket I/O data mapping like that BytesToMessageDecoder you see. 

For main we are using the built-in reactive programming for the logs, but as-of-yet a similar tool to netty is not built-in for this low level IO. The DDS client and Server, HRIT, DRGS, Iridium, and Network DCP sources are what would use the library as they all (except maybe the Network DCP sources in a few cases) are explicit low level protocols (e.g. they aren't HTTP). Everything else new would just use whatever the HTTP App server we're using would be.

MikeNeilson:
> Re: the virtual thread comment again

The benefits may also not be super clear in this initial setup. A follow up PR once this is merged will show how behavior is easily adapted based on various elements like requested DDS version and such.

MikeNeilson:
> > With Java 21 can we do this with built-in features (and some coding), and not have to include netty?
> > ref: https://www.reddit.com/r/java/comments/1d43rbh/how_can_java_21_virtual_thread_replace_reactive/
> 
> No, this change has little to do with Threading itself, and more to do with how the low level socket code is organized. This gives a structure with which to ignore most of the super low level socket I/O details and focus on the DDS protocol itself while at the same time forcing better design of the client/archive handling because of that structure.
> 
> Due to the design of Netty it _do_ look a lot like reactive programming, that is the style, but there's not an equivalent built in library that handles the socket I/O data mapping like that BytesToMessageDecoder you see.
> 
> For main we are using the built-in reactive programming for the logs, but as-of-yet a similar tool to netty is not built-in for this low level IO. The DDS client and Server, HRIT, DRGS, Iridium, and Network DCP sources are what would use the library as they all (except maybe the Network DCP sources in a few cases) are explicit low level protocols (e.g. they aren't HTTP). Everything else new would just use whatever the HTTP App server we're using would be.

Now that I reread the question, the answer is technically yes. But the reason for Netty was to deal with the low level Socket IO, that I would like to avoid doing as it is rather difficult to get correct. Netty provided a nice flow of "I know I have something valid so I can build the next chuck and pass it on" which kept the code well organized and focused. Could I do the same without netty... probably, but then aren't we just recreating netty? (I mean sure, not all of it, but arguably a lot.)

java/opendcs/src/main/java/org/opendcs/lrgs/dds/LddsCommandHandler.java
```diff
@@ -0,0 +1,126 @@
...
+import lrgs.ldds.LddsMessage;
+
+/**
+ *
+ * LddsCommandHandler
+ *
+ * Given an {@link lrgs.ldds.LddsCommand instance} process the message appropriate.
+ *
+ * NOTE: This also where things like establishing session specific objects, like a MessageArchiveRetriever
+ * or an implementation similar to {@link lrgs.ldds.LddsThread} (though far more generic) that wraps those instanced.
```
ktarbet:
```suggestion
 * or an implementation similar to {@link lrgs.ldds.LddsThread} (though far more generic) that wraps those instances.
```

java/opendcs/src/main/java/org/opendcs/lrgs/dds/LddsMessageDecoder.java
```diff
@@ -0,0 +1,74 @@
...
+            hdr[2] != validSync[2] ||
+            hdr[3] != validSync[3])
+		{
+			throw new ProtocolError("Could not read valid sync pattern ("+new String(hdr, StandardCharsets.UTF_8) +")");
+		}
+
+        var msg = new LddsMessage(hdr);
+        var length = Integer.parseInt(new String(ArrayUtil.getField(hdr, 5, 5)));
+
+        if (in.readableBytes() < length)
```
ktarbet, line 66: If we have just the header (no message body yet) , will this code drop that message before it comes in?   Or does a whole message always come together (header+body)

java/opendcs/src/main/java/org/opendcs/lrgs/dds/LddsCommandDecoder.java
```diff
@@ -0,0 +1,41 @@
...
+ * appropriate {@link lrgs.ldds.LddsCommand} instance
+ */
+public class LddsCommandDecoder extends MessageToMessageDecoder<LddsMessage>
+{
+    private static final CmdFactory factory = new CmdFactory();
+
+    @Override
+    protected void decode(ChannelHandlerContext ctx, LddsMessage msg, List<Object> out) throws Exception
+    {
+        out.add(factory.makeCommand(msg));
```
ktarbet, line 39: makeCommand(msg) can return null.  Should null's be dropped instead of added.

java/opendcs/src/main/java/org/opendcs/lrgs/dds/LddsCommandHandler.java
```diff
@@ -0,0 +1,126 @@
...
+        else if (msg instanceof CmdGoodbye)
+        {
+            log.atInfo().log("User Sent Goodbye");
+            var res = new LddsMessage(LddsMessage.IdGoodbye, "");
+            ctx.writeAndFlush(res).addListener(ChannelFutureListener.CLOSE);
+        }
+        else if (msg instanceof LddsCommand cmd)
+        {
+            log.atInfo().log("Sending back response. Command was {}", cmd.cmdType());
+            ctx.writeAndFlush(cmd);
```
ktarbet, line 111: Does cmd need to be converted to LddsMessage? -- The other if/else above all  writeAndFlush(LddsMessage )

integrationtesting/lrgs/src/test/java/org/opendcs/lrgs/NettyLrgsTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+class NettyLrgsTest
+{
+    private static LrgsTestInstance lrgs = null;
+    private static EventLoopGroup boss = new NioEventLoopGroup();
+    private static EventLoopGroup worker = new NioEventLoopGroup();
+    private static ChannelFuture channel;
+    private static int port;
+
+    @BeforeAll
```
ktarbet, line 38: Do we need cleanup via  @AfterAll

java/opendcs/src/main/java/org/opendcs/lrgs/dds/LddsMessageDecoder.java
ktarbet, line 66: consider limiting length  to LddsMessage.ValidMaxDataLength

java/opendcs/src/main/java/org/opendcs/lrgs/dds/LddsCommandDecoder.java
```diff
@@ -0,0 +1,41 @@
...
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.MessageToMessageDecoder;
+import lrgs.ldds.CmdFactory;
+import lrgs.ldds.LddsMessage;
+
+/**
+ *
+ * LddsCommandDecoder
+ *
+ * Given the LddsMessage that should've been previosly created, create an
```
ktarbet:
```suggestion
 * Given the LddsMessage that should've been previously created, create an
```

java/opendcs/src/main/java/org/opendcs/lrgs/dds/LddsCommandHandler.java
```diff
@@ -0,0 +1,126 @@
...
+import lrgs.ldds.CmdHello;
+import lrgs.ldds.DdsVersion;
+import lrgs.ldds.LddsCommand;
+import lrgs.ldds.LddsMessage;
+
+/**
+ *
+ * LddsCommandHandler
+ *
+ * Given an {@link lrgs.ldds.LddsCommand instance} process the message appropriate.
```
ktarbet:
```suggestion
 * Given an {@link lrgs.ldds.LddsCommand instance} process the message appropriately.
```

java/opendcs/src/main/java/org/opendcs/lrgs/dds/LddsCommandHandler.java
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * NOTE: This also where things like establishing session specific objects, like a MessageArchiveRetriever
```
ktarbet:
```suggestion
 * NOTE: This is also where things like establishing session specific objects, like a MessageArchiveRetriever
```

java/opendcs/src/main/java/org/opendcs/lrgs/dds/LddsCommandHandler.java
```diff
@@ -0,0 +1,126 @@
...
+
+    @Override
+    public void channelInactive(ChannelHandlerContext ctx) throws Exception
+    {
+        log.atInfo().log("Channel deactivated. Client disconnected");
+    }
+
+    /**
+     * The command reponse implementation details here present as placeholders to verify the generic request/response
+     * handling with the origin LddsClient works. The idea is not to implement the entire functionally here.
```
ktarbet:
```suggestion
     * handling with the original LddsClient works. The idea is not to implement the entire functionality here.
```

java/opendcs/src/main/java/org/opendcs/lrgs/dds/LddsCommandHandler.java
```diff
@@ -0,0 +1,126 @@
...
+            ctx.writeAndFlush(res).addListener(ChannelFutureListener.CLOSE);
+        }
+        else if (msg instanceof LddsCommand cmd)
+        {
+            log.atInfo().log("Sending back response. Command was {}", cmd.cmdType());
+            ctx.writeAndFlush(cmd);
+        }
+        else
+        {
+            log.atInfo().log("did not recieve an LddsCommand instance, was " + msg.getClass().getName());
```
ktarbet:
```suggestion
            log.atInfo().log("did not receive an LddsCommand instance, was " + msg.getClass().getName());
```

ktarbet (commented):
When further developed, consider if this this code be run in parallel to existing message code, perhaps run both if the design could support that to increase confidence.

integrationtesting/lrgs/src/test/java/org/opendcs/lrgs/NettyLrgsTest.java
MikeNeilson: Not at this time. Eventually this will be wrapped as a "DdsServer" instead of the raw setup here.

java/opendcs/src/main/java/org/opendcs/lrgs/dds/LddsCommandDecoder.java
MikeNeilson: hmm, probably, but the current test will never be null. can't decide if I want to keep the current factory or make something new either.

java/opendcs/src/main/java/org/opendcs/lrgs/dds/LddsCommandHandler.java
MikeNeilson: Well, maybe, we don't actually have an echo, this is currently a placeholder.

The CommandHandler will be rebuild in the followup to show a cleaner flow. Just felt that changing it all now would make this PR too large.

java/opendcs/src/main/java/org/opendcs/lrgs/dds/LddsMessageDecoder.java
MikeNeilson: No, the contract of the BytesToMessageDecoder state that the ByteBuffer isn't cleared until something is put into the out list.

What happens is that this is repeatedly called until that condition is met as new bytes come in.

MikeNeilson: Good point, though I think the value provided by the header can't go higher than that.

ktarbet: On the next call, the new bytes, in this case (possibly the message), will get put in the header?

ktarbet (approved):

java/opendcs/src/main/java/org/opendcs/lrgs/dds/LddsMessageDecoder.java
MikeNeilson: No the header but the Bytebuffer (the `in` variable). So if only like the few bytes of the header had arrived then we bail, or if the full message hasn't arrived the < length would be return and we'd also just bail.

The bytes that had already arrived stay in `in` (the implementations details are in `ByteToMessageDecoder`) and then was the buffer fully contains a message we can read all the bytes, create the LddsMessage, add it to out, and move on. The ByteToMessageDecoder implementation will see the out has something and remove the number of bytes we've read from the buffer. Which would then be passed on to the next call of this.

Though with DDS in theory the client just waits until we respond so there shouldn't be any waiting data after that point.

PR #2114, opened 2026-08-04
Chore/7.0 aarch64 docker images

### LINKED ISSUE #1050: Multi-Arch container image
There is nothing preventing OpenDCS from running on platforms such as ARM.

Using docker buildx, in some way, create a multi-arch image.

**Additional context**
The operator project attempted this with rust, while it worked, the compiling of Rust was incredibly slow with the qemu vm that was used for arm.

As the main project is java, we should determine if the platforms can share the same build to reduce the number of operations.

dennisgsmith:
## Problem Description

Backports 7.0 branch to support arm64 natively
Fixes #1050 

Relates to https://github.com/opendcs/opendcs/pull/2113

Currently only amd64 images are published.

## Solution

This was a little trickier than https://github.com/opendcs/opendcs/pull/2113 since eclipse-temurin:17-jre-alpine doesn't support arm64 builds ([21-jre-alpine is where they started it seems](https://hub.docker.com/_/eclipse-temurin/tags?name=21-jre-alpine)). For this reason, the base image was migrated to ubuntu-based (jammy) which makes the final image larger but this seemed like a less obtrusive change than updating the image to use jre 21:

| Image  | Compressed | Uncompressed |
|--------|--------------|----------------|
| eclipse-temurin:17-jre-alpine | 64.7 MB | 256 MB |
| eclipse-temurin:17-jre-jammy | 89.1 MB | 374 MB |
| difference | +24.4 MB | +118 MB |

If the added size isn't acceptable, I could try installing jre image on an alpine base image. Let me know if that (or another method) would be preferred.

## how you tested the change

Similar to [`main`](https://github.com/opendcs/opendcs/pull/2113), I created a github action with the relevant bits to build the images:

Temporary test builder CI: https://github.com/dennisgsmith/opendcs/commit/5a9897188df0a77ae060cbfffb60b22ee912f9f3

CI summary/logs: https://github.com/dennisgsmith/opendcs/actions/runs/30939994326

## Where the following done:


Dockerfile
```diff
@@ -12,13 +14,17 @@ RUN --mount=type=cache,target=/root \
     ./gradlew installDist -Dno.docs=true
 # end initial build
 
-FROM eclipse-temurin:17-jre-alpine AS opendcs_base
```
MikeNeilson, line 15: Hmm, I'm a bit loathe to pull in such a large image just because of a jre issue.

If another JRE 17 is available for arm I would argue we use that instead. Those ubuntu/debian images tend to get way more, and often real, vulnerability reports.

Alternatively, since it's likely to just be you and I using these for now. the testing is capped 17 due to some of the required libraries just not working, but the software itself *should* work fine with Java 21. So if we can just make the arm64 image use java 21 that could work.

dennisgsmith: > just make the arm64 image use java 21

Do you mean java 17 would still be used for the amd64 base and 21 used for the arm64 base? Or just switching to java 21 for both? So far I haven't found too many good candidates for a java 17 arm image that's as small as the alpine one.

MikeNeilson: 17 -> amd64
21 -> arm64

*would be better*, but I'm also okay with trying just 21. Though is the issue no java 17 for ARM or no image? it wouldn't be unreasonable just start from the alpine base image and manually install jdk17 as well.

dennisgsmith: > it wouldn't be unreasonable just start from the alpine base image and manually install jdk17 as well.

I agree this is ideal; Seem promising since [openjdk17-jre is available as an apk community package](https://pkgs.alpinelinux.org/packages?name=openjdk17-jre&branch=v3.23&repo=&arch=&origin=&flagged=&maintainer=)

I've updated the image to use this for opendcs_base.

MikeNeilson (dismissed):

.github/workflows/test-aarch64-build.yml
MikeNeilson, line 1: Leave this in. Always good to have a test like this before you get to all the way to the release build.

dennisgsmith:
@MikeNeilson Forgot to add the dockerfile syntax tag at the top causing the build to fail in this CI runner. I updated it so I believe the build should work now.

.github/workflows/test-aarch64-build.yml
MikeNeilson, line 1: Please return this file. Valid tests are *never* useless. Last thing I want is to be debugging this trying to make a release when the time comes.

MikeNeilson (changes requested):

.github/workflows/test-aarch64-build.yml
dennisgsmith: Apologies, I must have committed by mistake before I read your comment, will return the file.

MikeNeilson: No worries. Oddly seems to be peoples default, but at least easy to fix.

We should probably restore that on main as well, but I want to do some clean up there first. Tests good, duplication not so much. Should move it into an reusable action so we aren't duplicating the code just for the test. But I'll take care of that later.

dennisgsmith: Yeah, I had initially just wanted to run it on my fork as a kind of smoke test, but I definitely see the value in keeping it around. Thanks for the feedback.

MikeNeilson (approved):

.github/workflows/test-aarch64-build.yml
```diff
@@ -0,0 +1,39 @@
+---
+name: Test aarch64 Docker Build
+on:
+  push:
+    branches:
+      - chore/7.0-aarch64-docker-images
```
MikeNeilson: That should be 
```
name: Cross Platform Build Smoke Tests
on:
  pull_request:
    branches:
      - "7.0"
    types: [opened, synchronize, reopened]
  push:
    branches:
      - "7.0"
```
Otherwise it's never gonna run.

MikeNeilson (changes requested):
We both completely missed that.

MikeNeilson (approved):
What is this... 5th times the charm?

PR #2118, opened 2026-08-06
RoutingScheduler forced exit.

### LINKED ISSUE #2117: CWMS - RoutingScheduler and Database connections
**Describe the bug**
During database shutdowns the RoutingScheduler is not correctly recovering from database issues.
Instead of continuing to retry, it appears to be simply dead.

**To Reproduce**
1. Start database
2. Start scheduler
3. Stop database
4. See error, then no logs.

**Expected behavior**
Either application exit, or attempts are retry.

**Screenshots**

**Additional context**

MikeNeilson:
## Problem Description

May fix #2117. 

In environment application is hosted in a lost "driver application". While the error detected is getting to the point where the JVM should be existing, the driver application is showing that it is still active.

## Solution

Wrap routsched.execute in a try catch so System.exit(1) and be called on error and System.exit(0) on normal termination.

## how you tested the change

Manually to verify new logging and error codes. Using `./gradlew runApp -Popendcs.app=routesched` behavior seems reasonable. Unfortunately I was not able to totally recreate the conditions. Will test change in USACE cloud environment and update if necessary.

Automated test not applicable, this is baseline JRE behavior.

## Where the following done:

### DIFF
--- a/java/opendcs/src/main/java/decodes/routing/RoutingScheduler.java
+++ b/java/opendcs/src/main/java/decodes/routing/RoutingScheduler.java
@@ -511,7 +511,21 @@ public class RoutingScheduler extends TsdbAppTemplate implements RoutingSchedule
 				});
 		}
 
-		routsched.execute(args);
+		// there may be a few non-daemon threads about that haven't been corrected yet
+		// So whether there is failure or not, forcibly exit the JVM so systems
+		// can detected and restart properly.
+		try
+		{
+			routsched.execute(args);
+			System.exit(0);
+		}
+		catch (Exception ex)
+		{
+			log.atError()
+			   .setCause(ex)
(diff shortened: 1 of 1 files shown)


MikeNeilson:
@opendcs/opendcs-core-devs tagging everyone, but I am just going to force this through.

adamkorynta (approved):

java/opendcs/src/main/java/decodes/routing/RoutingScheduler.java
adamkorynta, line 504: Is this shutdown hook still needed then? or should the log line move into a finally below

MikeNeilson: Technically I don't think it was ever really needed, more informational and for a different usage that what I'm doing right now.

Issue #2110, opened 2026-08-02
LRGS - migrate TCP operations to use Netty


MikeNeilson:
**Is your feature request related to a problem? Please describe.**
The current LRGS network code is difficult to follow and maintain.

**Describe the solution you'd like**
Refactor/modernize code to make use of the Netty style of doing things.

**Describe alternatives you've considered**
Keep custom code

**Additional context**
- https://netty.io/
- Contains built in ProxyProtocol handler that would help solve #2109

PR #2122, opened 2026-08-10
Task/2110 additional dds server implementation

MikeNeilson:
## Problem Description

Partial for #2110. 

## Solution

- Reorg setup code for Netty based DdsServer
- Duplicate and rearrange code from CmdGetMsgBlockExt to allow future reuse.
- Create DdsSession and DdsUserPrincipal object to start isolating such data from implementation details
 - This is containing data that is currently member variables of LddsThread, intention is to have LddsThread *also* use the DdsSession object in the future (after sufficient testing)

Intentionally only implements one Get Message command to demonstrate changes and help show/reveal how we might organize.

Challenges in current design:
The specific command error handling is not as clean. I think the proper solution is still to use this style of mechanism but to have each command have it's own ChannelInboundHandler so that there can be a "happy path" and "specific exception path" as appropriate to each given system.

GetMsgBlockEx should likely implement a new "LddsCommand" interface that otherwise behaves in a similar way. Though I think the return should be some form or `Either` or multi return as several elements depend on things like "x results", though only the GetDcpMessages return anything and it appears to be for status updates; so should more likely be moved to standardized metrics collection.

NOTE: intention is to address the challenges above in future PRs, I'm going very slow on this to make sure various behavior and edge cases can be covered while better understanding how the current LRGS mechanisms work so that implementation and behavior can be improved over time.

## how you tested the change

- Additional integration tests.
- Server starts
- Client connects
- Attempts to retrieve data with default "get all" SearchCriteria.

## Where the following done:

### DIFF
--- a/integrationtesting/lrgs/src/test/java/org/opendcs/lrgs/ArchiveOperationsTestIT.java
+++ b/integrationtesting/lrgs/src/test/java/org/opendcs/lrgs/ArchiveOperationsTestIT.java
@@ -36,7 +36,7 @@ class ArchiveOperationsTestIT
     private static final String MSG_DATA = "Test String.";
 
     @BeforeAll
-    static void setup(LrgsTestInstance lrgs) throws Exception
+    static void setup(LrgsTestInstance lrgs)
     {
         // Store message
         assertNotNull(lrgs);
--- a/integrationtesting/lrgs/src/test/java/org/opendcs/lrgs/ArchiveOperationsTestIT.java
+++ b/integrationtesting/lrgs/src/test/java/org/opendcs/lrgs/ArchiveOperationsTestIT.java
@@ -2,16 +2,11 @@ package org.opendcs.lrgs;
 
 import static org.opendcs.fixtures.assertions.Waiting.assertResultWithinTimeFrame;
 
-import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
-import java.io.File;
-import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
 import java.util.Date;
 import java.util.concurrent.TimeUnit;
 
--- a/integrationtesting/lrgs/src/test/java/org/opendcs/lrgs/ArchiveOperationsTestIT.java
+++ b/integrationtesting/lrgs/src/test/java/org/opendcs/lrgs/ArchiveOperationsTestIT.java
@@ -10,6 +10,7 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 import java.io.File;
 import java.nio.charset.Charset;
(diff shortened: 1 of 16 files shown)


integrationtesting/lrgs/src/test/java/org/opendcs/lrgs/ArchiveOperationsTestIT.java
```diff
@@ -83,7 +84,60 @@ public DcpAddress dcpNameToAddress(String name)
...
+                else
+                {
+                    return false;
+                }
+            }
+            catch (EndOfArchiveException ex)
+            {
+                return false;
+            }
+        }, 15, TimeUnit.SECONDS, 5, TimeUnit.SECONDS,
```
ktarbet, line 98: Would you please help me understand how all the time-windows are working here.

Search Criteria: " now - 1day", to "now"   (local time? )
Retriever:  stop at CurrentTimeMillis() +1 Minute. (system time?)
Assert timeFrame: wait for 15 seconds  (System time?)

And the message was put into the archive during @BeforeAll (likely quite recent)

java/opendcs/src/main/java/lrgs/ldds/CmdAuthHello.java
```diff
@@ -480,4 +480,9 @@ public String toString()
 
 	/** @return the code associated with this command. */
 	public char getCommandCode() { return LddsMessage.IdAuthHello; }
+
```
ktarbet, line 483: consider if getDdsVersion() and the field this.dssVersion could be in the base class.  (the method is repeated in CmdHello )

ktarbet (approved):

java/opendcs/src/main/java/lrgs/ldds/CmdAuthHello.java
ktarbet: note: a test is failing.

integrationtesting/lrgs/src/test/java/org/opendcs/lrgs/ArchiveOperationsTestIT.java
MikeNeilson: ... yeah, that's a lot of time outs.

The ones at the moment are for the test itself.

The retriever... I honestly have not idea, that's been bugging me since I got this far.

The search criteria is the data.


Why does the test need a time out? Because at least the first call to `getNextPassingIndex` doesn't return anything. 

As to why is doesn't immediately return something? I have no idea and have decided for the moment to just roll with it.

I suspect it's something to do with the search logic/timing of indexing being built, within the XML Message Archive implementation. But any attempt to fix that would bloat this PR. This test was only included because I couldn't understand why the NettyLrgsTest failed and wanted to isolate the behavior.

java/opendcs/src/main/java/lrgs/ldds/CmdAuthHello.java
MikeNeilson: Yeah, these are the two that contain a DdsVersion sent by the user, thus need needs to be retrievable.

Alternatively it might make sense for DdsSession to not be a record and allow these to just set it. However, that would depend on a change to LddsCommand that would take DdsSession instead of LddsThread... which is where I think I want to go, but too much work for this PR. hence just exposing the version from the instance that has it.

MikeNeilson:
failing test is an unrelated test of task_list_operations that has been intermittently failing on different test engines.

PR #2125, opened 2026-08-11
Update postgres-driver, ganymed-ssh, usgs-waterdata-api, selenium

ktarbet:


MikeNeilson (dismissed):

ktarbet:
@MikeNeilson  any suggestions to replace the ganymed-ssh?   Considering we depend on netty does that push us in any direction?    I see ganymed used in three places:

* ScpDataSource -- copy files
* LritDcsconnection -- uses  ch.ethz.ssh2.Connection
* SendFileThread -- used by LritDcsMain

MikeNeilson:
> @MikeNeilson any suggestions to replace the ganymed-ssh? Considering we depend on netty does that push us in any direction? I see ganymed used in three places:
> 
>     * ScpDataSource -- copy files
> 
>     * LritDcsconnection -- uses  ch.ethz.ssh2.Connection
> 
>     * SendFileThread -- used by LritDcsMain

Not off the top of my head, though if Netty supports more direct SSH usage that would be a good canditate. Netty does tend to only do the lowest level bits of protocol though so I'm not sure how much extra work it would be.

Alternatively, even windows bundles with an ssh client now, we could always change that system to just using the system available ssh client. Though I think full control of session and other behavior would be better.

MikeNeilson (approved):

MikeNeilson:
> > @MikeNeilson any suggestions to replace the ganymed-ssh? Considering we depend on netty does that push us in any direction? I see ganymed used in three places:
> > ```
> > * ScpDataSource -- copy files
> > 
> > * LritDcsconnection -- uses  ch.ethz.ssh2.Connection
> > 
> > * SendFileThread -- used by LritDcsMain
> > ```
> 
> Not off the top of my head, though if Netty supports more direct SSH usage that would be a good canditate. Netty does tend to only do the lowest level bits of protocol though so I'm not sure how much extra work it would be.
> 
> Alternatively, even windows bundles with an ssh client now, we could always change that system to just using the system available ssh client. Though I think full control of session and other behavior would be better.

These seem to be the top choices:
- https://github.com/apache/mina-sshd (does client and server, we only need the client)
- https://github.com/hierynomus/sshj

Both have reasonably regular releases, Mina-SSHD is an Apache top-level project, does say a lot about it's quality of management.

MikeNeilson (approved):

PR #2127, opened 2026-08-11
Initial work, first table altered.

### LINKED ISSUE #742: Add an "Organization" type and database column

CWMS Supports multiple offices, Having the Reference Postgresql implementation support multiple "Organizations" will likely be a useful feature. Having this supported in only one implementation makes certain testing difficult and interfaces akward.

Create a concept of an Organization (additional field on each datatype). This could map to an actual organization structure like CWMS hierarchy of offices, or simply be a known constant per database/installation/etc.

Carry on the way we are, though I suspect we'll see issues during testing and mapping of DTOs in the rest_api. Right now for CWMS a different connection is required for a given office, even to the same physical database. Having an "Organization" be a fundemental concept makes it possible for the user to set the current organization and queries can retrieve data as appropriate.

**Additional context**

MikeNeilson:
## Problem Description

Fixes #742. 

## Solution

Add org tables and appropriate columns.

After thinking about it for way to long, I've decide to use the postgres row level security and oracle vpd (cwms already does this) mechanisms.

Will be attempting to setup a form of "hierarchy", e.g. the default "org" loads the initial algorithms and the sub orgs can either just use it, or override.

NOTE: this is still *very* draft, opening for comment though.

## how you tested the change

At this time manually, existing integration tests will show if there are any issue with accessing the default org or setting up the "session"

Will then add tests of additional organizations.

## Where the following done:

### DIFF
--- a/schema/opendcs-schemas/src/main/resources/db/OpenDCS-Postgres/schema/V7.6__organizations.sql
+++ b/schema/opendcs-schemas/src/main/resources/db/OpenDCS-Postgres/schema/V7.6__organizations.sql
@@ -13,7 +13,7 @@ create table organization (
 
 
 insert into org_type(id, name, description) OVERRIDING SYSTEM VALUE
-    values(0, '<default>', 'Basic organization with not distinct attributes');
+    values(0, '<default>', 'Basic organization with no distinct attributes');
 insert into organization(id, name, org_type, parent_id) OVERRIDING SYSTEM VALUE
     values (0, 'Default', 0, null);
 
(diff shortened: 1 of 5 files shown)


schema/opendcs-schemas/src/main/resources/db/OpenDCS-Postgres/schema/V7.6__organizations.sql
```diff
@@ -0,0 +1,89 @@
...
+create table organization (
+    id bigint generated always as identity (start with 1) primary key,
+    name varchar(128) not null unique,
+    org_type bigint references org_type(id),
+    parent_id bigint references organization(id)
+);
+
+
+insert into org_type(id, name, description) OVERRIDING SYSTEM VALUE
+    values(0, '<default>', 'Basic organization with not distinct attributes');
```
ktarbet:
```suggestion
    values(0, '<default>', 'Basic organization with no distinct attributes');
```

MikeNeilson:
Closing in favor of pivoting direction slightly.

Will wait for DAO replacement for the general case. will be modifying CWMS Roles to include the cwms office so we can properly assign and constraint, should help inform some of the design in the future.

PR #2135, opened 2026-08-14
Task/1882 complockdao

### LINKED ISSUE #1882: Create Stateless CompLock Dao
The original system lumped Computation App Locks (ProcLock) into the LoadingAppDao.

Future work should remove the need for the ProcLocks, so we'll create a separate AppLockDao for that purpose until such time as those locks are required anymore.

MikeNeilson:
## Problem Description

Fixes #1882. 

## Solution

Implement CompLock operations in stateless DAO.
Return `Result` over throwing an exception. Point of usage can throw the exception if needed.
NOTE: copy the implementation directly from the existing LoadingAppDao, could likely share the logic, but don't see the point as once the new one is used everywhere we'll just delete it.

## how you tested the change

Existing tests of REST API
New integration test.

## Where the following done:

### DIFF
--- a/integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CompLockDaoTestIT.java
+++ b/integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CompLockDaoTestIT.java
@@ -58,9 +58,9 @@ class CompLockDaoTestIT extends AppTestBase
 
             var lockOneUpdate = lockDao.obtainLock(tx, appOne, 0, "bob");
             assertTrue(lockOneUpdate.isSuccess());
-            var firstLock = lockOne.getSuccess();
+            var firstLock = lockOne.success();
 
-            var updatedLock = lockOneUpdate.getSuccess();
+            var updatedLock = lockOneUpdate.success();
             assertNotEquals(firstLock.getHeartbeat(), updatedLock.getHeartbeat());
 
             var lockTwo = lockDao.obtainLock(tx, appTwo, 0, "bob");
--- a/integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CompLockDaoTestIT.java
+++ b/integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CompLockDaoTestIT.java
@@ -54,7 +54,7 @@ class CompLockDaoTestIT extends AppTestBase
             var lockOneFail = lockDao.obtainLock(tx, appOne, 0, "alice");
             assertTrue(lockOneFail.isFailure());
             
-            Thread.sleep(000); // NOSONAR otherwise the dates will match.
+            Thread.sleep(1500); // NOSONAR otherwise the dates will match.
 
             var lockOneUpdate = lockDao.obtainLock(tx, appOne, 0, "bob");
             assertTrue(lockOneUpdate.isSuccess());
--- a/integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CompLockDaoTestIT.java
+++ b/integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CompLockDaoTestIT.java
@@ -53,10 +53,13 @@ class CompLockDaoTestIT extends AppTestBase
             var lockOneFail = lockDao.obtainLock(tx, appOne, 0, "alice");
             assertTrue(lockOneFail.isFailure());
             
+            Thread.sleep(000); // NOSONAR otherwise the dates will match.
+
             var lockOneUpdate = lockDao.obtainLock(tx, appOne, 0, "bob");
             assertTrue(lockOneUpdate.isSuccess());
             var firstLock = lockOne.getSuccess();
-            var updatedLock = lockOne.getSuccess();
+
+            var updatedLock = lockOneUpdate.getSuccess();
             assertNotEquals(firstLock.getHeartbeat(), updatedLock.getHeartbeat());
 
             var lockTwo = lockDao.obtainLock(tx, appTwo, 0, "bob");
--- a/integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CompLockDaoTestIT.java
+++ b/integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CompLockDaoTestIT.java
@@ -1,5 +1,6 @@
 package org.opendcs.dao;
 
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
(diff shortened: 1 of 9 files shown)


integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CompLockDaoTestIT.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }    
+    
+    @Test
+    void test_lock_operations() throws Exception
+    {
+        
+        var lockDao = db.getDao(CompLockDao.class).orElseThrow();
+
+        try (var tx = db.newTransaction())
```
ktarbet, line 50: Doesn't the transaction itself provide some 'locking'?

java/opendcs/src/main/java/org/opendcs/database/dai/CompLockDao.java
```diff
@@ -0,0 +1,86 @@
...
+     * @return
+     * @throws OpenDcsDataException
+     */
+    Optional<TsdbCompLock> getLock(DataTransaction tx, DbKey appId) throws OpenDcsDataException;
+
+    /**
+     * Obtain a specific lock. If the lock already exists, and is not for the same pid and is not stale
+     * LockBusyException will be returned instead of the lock.
+     * 
+     * Otherwise the Lock is returned and know owned by that instance.
```
ktarbet:
```suggestion
     * Otherwise the Lock is returned and known owned by that instance.
```

ktarbet (commented):
checkpoint

integrationtesting/opendcs-tests/src/test/java/org/opendcs/dao/CompLockDaoTestIT.java
MikeNeilson: Not in this context, could use "select for update" to lock the row itself but given how these work I didn't seem worth the hassle.

I expect the Transaction used for the "App Locks", which are more informative than being an actual lock, to be short lived.

java/opendcs-implementation/src/main/java/org/opendcs/database/impl/opendcs/dao/CompLockDaoImpl.java
```diff
@@ -0,0 +1,197 @@
...
+    public Optional<TsdbCompLock> getLock(DataTransaction tx, DbKey appId) throws OpenDcsDataException
+    {
+        var handle = tx.connection(Handle.class)
+                       .orElseThrow(() -> new OpenDcsDataException(SqlErrorMessages.NO_JDBI_HANDLE));
+        try (var select = handle.createQuery(
+                            queries.getInstanceOf(SELECT)
+                                  .add("where", " where lk.loading_application_id = :id ")
+                                  .render()))
+        {
+            select.setSqlLogger(new DetailSqlLogger(log));
```
ktarbet: consider if this should be debug only logging?

java/opendcs-implementation/src/main/java/org/opendcs/database/impl/opendcs/dao/CompLockDaoImpl.java
```diff
@@ -0,0 +1,197 @@
...
+    @Override
+    public Result<TsdbCompLock, LockBusyException> obtainLock(DataTransaction tx, CompAppInfo appInfo, int pid,
+            String host) throws OpenDcsDataException
+    {
+        var lock = getLock(tx, appInfo.getAppId()).orElse(null);
+        if (lock != null)
+        {
+            // check for same application and host. With the introduction of the docker images
+            // every instance generally would have the same PID, just a different hostname.
+            if (lock.getPID() == pid && lock.getHost().equals(host))
```
ktarbet:
```suggestion
            if (lock.getPID() == pid && lock.getHost().equalsIgnoreCase(host))
```

following example above

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppResources.java
```diff
@@ -446,28 +444,30 @@ public Response deleteApp(@Parameter(description = "App ID", required = true, ex
 			},
 			tags = {"OpenDCS Process Monitor and Control (APP)"}
 	)
-	public Response getAppStat() throws DbException
+	public Response getAppStat() throws WebAppException
 	{
-		List<ApiAppStatus> ret = new ArrayList<>();
-		try (LoadingAppDAI dai = getLegacyDatabase().makeLoadingAppDAO())
+		var db = createDb();
+		var dao = db.getDao(CompLockDao.class).orElseThrow();
```
ktarbet:
```suggestion
		var dao = db.getDao(CompLockDao.class).orElseThrow(() -> UNABLE_TO_GET_COMPLOC_DAO); 
```

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppResources.java
ktarbet:
```suggestion
	private static final WebAppException UNABLE_TO_GET_APP_DAO = new WebAppException(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(), "No Loading Application DAO available.");
private static final WebAppException UNABLE_TO_GET_COMPLOC_DAO= new WebAppException(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(), "No Comp Lock DAO available.");
```

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppResources.java
```diff
@@ -476,21 +476,7 @@ static ApiAppStatus map(LoadingAppDAI dai, TsdbCompLock lock) throws DbIoExcepti
...
-					ret.setEventPort(Integer.parseInt(app.getProperties().getProperty("EventPort")));
-				}
-				ret.setAppType(app.getAppType());
-			}
-			catch (DbIoException | NoSuchObjectException | NumberFormatException ex)
-			{
-				throw new DbIoException("Error mapping app status", ex);
-			}
-		}
+		// Event port is not longer supported, don't propgate the data in new systems.
```
ktarbet:
```suggestion
		// Event port is not longer supported, don't propagate the data in new systems.
```

java/opendcs/src/main/java/org/opendcs/database/dai/CompLockDao.java
```diff
@@ -0,0 +1,86 @@
...
+     */
+    void releaseLock(DataTransaction tx, TsdbCompLock lock) throws OpenDcsDataException;
+
+    /**
+     * Check if given Lock (App) is already set. If so, update the heartbeat time and status.
+     *
+     * @param tx
+     * @param lock
+     * @return The updated Lock, or the reason the attempt at the lock failed. Returned lock is always
+     * updated from the database regardless of weather or not the heartbeat or status was updated.
```
ktarbet:
```suggestion
     * updated from the database regardless of whether or not the heartbeat or status was updated.
```

ktarbet (approved):

PR #2138, opened 2026-08-17
Rename FailableResult to Result and add additional behavior, addition…

MikeNeilson:
…ally move Throwing 'functions' into opendcs-api.

## Problem Description

Mostly the super long name was starting to bug me and I wanted to add some additional behavior
to make usage a bit for fluent and reduce large blocks.

## Solution

Rename `FailableResult` to `Result` and moved to opendcs-api sub-project. Also moved the `ThrowingX` interfaces to opendcs-api as at least one was needed to move the `Result` class.

Added additional handlers to make code flow easier in various usages.

## how you tested the change

Existing usages.
New test of just Result.

## Where the following done:


java/opendcs-api/src/main/java/org/opendcs/util/Result.java
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Create Result instance with successful value.
+     *
+     * @param <S>
+     * @param <F>
+     * @param successResult
+     * @return
+     */
+    public static <S, F> Result<S, F> success(S successResult)
```
adamkorynta, line 140: The success/failure factory methods should probably guard against null input.

Confusing evaluation:
`Result.success(null).isSuccess()` - true
`Result.failure(null).isSuccess()` - true

if null is valid, then `getSuccess` and the function methods should wrap with Optional

java/opendcs-api/src/main/java/org/opendcs/util/Result.java
```diff
@@ -0,0 +1,155 @@
...
+            throw new IllegalStateException("Attempt to retrieve 'failure' result of a succesfull result.");
+        }
+        return failResult;
+    }
+
+    /**
+     * Simple callback for any failure results.
+     * @param consumer
+     */
+    public void handleError(Consumer<F> consumer)
```
adamkorynta: would name `onError` to match `onSuccess`

java/opendcs-api/src/main/java/org/opendcs/util/Result.java
```diff
@@ -0,0 +1,155 @@
...
+        {
+            handleSuccess.accept(successResult);
+        }
+    }
+
+    public F getFailure()
+    {
+        if (!isFailure())
+        {
+            throw new IllegalStateException("Attempt to retrieve 'failure' result of a succesfull result.");
```
adamkorynta:
```suggestion
            throw new IllegalStateException("Attempt to retrieve 'failure' result of a successful result.");
```

java/opendcs-api/src/main/java/org/opendcs/util/Result.java
```diff
@@ -0,0 +1,155 @@
...
+    {
+        return failResult == null;
+    }
+
+    public boolean isFailure()
+    {
+        return failResult != null;
+    }
+
+    public S getSuccess()
```
adamkorynta: how about just `success()` and `failure()` to be more idiomatic

java/opendcs-api/src/main/java/org/opendcs/util/functional/ThrowingFunction.java
```diff
@@ -0,0 +1,10 @@
+package org.opendcs.util.functional;
+
+/**
+ * function interface to allow returning result with a checked exception.
+ */
+@FunctionalInterface
+public interface ThrowingFunction<T,R,E extends Exception>
+{
+    R accept(T value) throws E;
```
adamkorynta: `accept` is the method for a `Consumer`. I suggest `apply` to match `Function`

java/opendcs-api/src/main/java/org/opendcs/util/Result.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.function.Function;
+
+import org.opendcs.util.functional.ThrowingFunction;
+
+/**
+ * Pair objects used for returning Successful and failed results for processing in a stream.
+ * @param <S> Desired Object
+ * @param <F> Object containing error details. Most commonly an Exception, but can be anything.
+ */
+public final class Result<S, F>
```
adamkorynta, line 29: I assume you've already evaluated Vavr, but in case not, you can use the Try/Either classes to borrow naming from if this class starts expanding. They've got:

```
success()
failure()
map()
flatMap()
fold()
onSuccess()
onFailure()
```

Given the MIT license, might just be worth pulling in...

java/opendcs-api/src/main/java/org/opendcs/util/functional/ThrowingFunction.java
MikeNeilson: thanks, not sure how I missed that.

java/opendcs-api/src/main/java/org/opendcs/util/Result.java
MikeNeilson: I've heard of it. And that makes sense. We'll see how things go over time.

java/opendcs-api/src/main/java/org/opendcs/util/Result.java
MikeNeilson: Yeah, that makes sense given everything else I'm doing.

java/opendcs-api/src/main/java/org/opendcs/util/Result.java
MikeNeilson: Thanks, thought about it but wasn't sure it if made sense to do. But considering I renamed the entire class it does.

java/opendcs-api/src/main/java/org/opendcs/util/Result.java
MikeNeilson: Good idea, users of the class can choose to use to use optional or not.

MikeNeilson (commented):
Thanks, always nice to get a little push to do that slight bit more.

adamkorynta (dismissed):

adamkorynta (approved):

PR #2141, opened 2026-08-18
Fix/Create new platform

### LINKED ISSUE #2016: Web UI - Platforms - Cannot create a new platform
**Describe the bug**
This might be a user error with trying to figure out the new UI, but when I try to create a new Platform it doesn't appear to be saving.  In addition, when I selected a Config that I know has sensors, none populated in the Sensors section of the Platform.  I was unable to give the Platform a "Name", which maybe could be why it isn't saving?

**To Reproduce**
1. Open Web UI, open Platforms page
2. Click on + icon to add a new platform. A new row at the top with an ID = -1 is added.
3. The user can edit at this point or they can click the Edit icon and edit.  Not sure if that is intentional but both work.
4. See that the Name section is grayed out.  Select a Site and Config, set a designator, agency, and description.  Attempt to save and see that no new row is created.  For NWDM, I was at 1,747 entries and attempted to add two different Platforms unsuccessfully.  The search function doesn't bring up the newly created Platforms and the entries are still at 1,747.  This is true after a refresh of the page as well.

**Expected behavior**
Expected the Platform to save.  Expected Name to be populated once selecting the Site, or to allow the User to edit that field in some way.

**Screenshots**
Example of Platform that didn't save.  Used MCSD as the Config since I knew that had Sensors in the Web UI.
Image

tlelv:
## Problem Description

Fixes #2016 (to ensure we can close it). 

## Problem Description

Fixes #2016.

## Solution

#2063 fixed part of #2016 but two problems remained:

1. **New-platform saves failed silently.** `Platform.tsx`'s `savePlatform` called
   `actions.save?.(...)` without awaiting or catching it. Clicking Save just did nothing. Fixed by awaiting/catching in
   `savePlatform` and showing the real error via an `Alert`

2. **The "Name" field never populated after selecting a Site, and the two
   field labels were swapped.** 

## how you tested the change

## Where the following done:


MikeNeilson (dismissed):
I'm pretty sure I've seen that save error pattern in other, if not all other, pages. 

Don't do it for this PR, but we should probably figure out how to extract that out to some sharable thing.

MikeNeilson (approved):

PR #2143, opened 2026-08-19
Refactor/shared save error

tlelv:
## Problem Description

Fixes #. 

Four pages carried identical copies of the same save-error block: a `saveError` state, a `try/catch` around `actions.save` that logged the raw error and ran it through `apiErrorMessage`, and the same danger `Alert` just above the form's action row.

- `src/pages/platforms/Platform.tsx`
- `src/pages/loading-apps/LoadingApp.tsx`
- `src/pages/computations/algorithms/Algorithm.tsx`
- `src/pages/computations/computations/Computation.tsx`

## Solution

Describe your solution.

Pulled the pattern into one hook and one component. Based on the platform create/save branch, so this should be merged after that one.

## how you tested the change

## Where the following done:


MikeNeilson (dismissed):
I like it. But let's wait for Oskar to review as well. Make sure we aren't missing something important or just useful.

MikeNeilson (dismissed):

oskarhurst (changes requested):
Everything looks good here. I did find one more place with this pattern that you missed

src/components/data-table/AppDataTable.tsx -- saveError state at :462

can you please add it so we don't forget about it.

MikeNeilson (approved):

oskarhurst (approved):

PR #2147, opened 2026-08-21
Refactor/sonar reliability cleanup

tlelv:
## Problem Description

Fixes #. 

Sonarcloud reliability issues.

## Solution

Describe your solution.

## how you tested the change

## Where the following done:


java/cwms-implementation/src/main/java/org/opendcs/database/impl/cwms/dao/CwmsPlatformDaoImpl.java
```diff
@@ -27,7 +27,7 @@ public CwmsPlatformDaoImpl()
                 PlatformMapper.withPrefix("p"),
                 DecodesConfigMapper.withPrefix("config"),
                 TransportMediumMapper.withPrefix("tm"),
-                CwmsSiteMapper.withPrefix("s"),
+                CwmsSiteMapper.forPrefix("s"),
```
MikeNeilson: why is this getting renamed?

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ComputationResources.java
```diff
@@ -370,31 +370,27 @@ public void runComputation(
 			Date startDate = Date.from(startTime);
 			Date endDate = Date.from(endTime);
 
+			final SseChannel channel = new SseChannel(sse, eventSink, compStatus, taskID);
```
MikeNeilson:
```suggestion
			final var channel = new SseChannel(sse, eventSink, compStatus, taskID);
```

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ComputationResources.java
```diff
@@ -562,80 +555,85 @@ else if(parm.getDataTypeId() == null && dataTypeId != null)
...
-		eventSink.send(event);
+				.build());
+	}
+
+	/**
+	 * The Sse factory, the sink and the name/id stamped on every event of a single computation run.
+	 * These four values are always passed together; bundling them keeps the run helpers well under
+	 * the parameter limit and removes the repeated event-builder boilerplate.
+	 */
+	private record SseChannel(Sse sse, SseEventSink eventSink, String eventName, String taskId)
```
MikeNeilson: This may be likely to be used in other components, move to own file.

integrationtesting/fixtures/src/main/java/org/opendcs/fixtures/extensions/lrgs/LrgsTestExtension.java
```diff
@@ -23,19 +27,39 @@ public void beforeAll(ExtensionContext context) throws Exception
...
-            assertDoesNotThrow(() -> //NOSONAR - test fixture code, not production code
+            assertDoesNotThrow(() ->
             {
                 // java.io.tmpdir is redirected to a project-local build directory by the
                 // consuming test tasks (see opendcs-tests/build.gradle), so this isn't
-                // landing in a shared, publicly writable system temp directory.
-                File lrgsHome = Files.createTempDirectory("lrgshome").toFile();
+                // landing in a shared, publicly writable system temp directory. The
+                // directory is also created owner-only where the platform supports it.
+                File lrgsHome = createPrivateTempDirectory().toFile();
```
MikeNeilson: The lrgshome prefix needs to stay, this will be used in the main integration tests soon and it will be useful to distinguish between the lrgs home and other directories more quickly.

java/opendcs/src/main/java/decodes/cwms/CwmsTimeSeriesDAO.java
MikeNeilson, line 1: Bold, I've been putting that one off myself. Thanks for tackling it.

MikeNeilson (changes requested):

java/cwms-implementation/src/main/java/org/opendcs/database/impl/cwms/dao/CwmsPlatformDaoImpl.java
tlelv: Sonar flagged it as: CwmsSiteMapper.withPrefix hides OpenDcsSiteMapper.withPrefix rather than overriding it. i just did it for that but i can revert

MikeNeilson: Yeah, revert, that's rather intentional... though does indicate we might want to revise the overall design of retrieving wrappers. Just not in this PR.

java/opendcs/src/test/java/decodes/cwms/CwmsTimeSeriesDAOTest.java
```diff
@@ -39,4 +67,346 @@ public void testWithinMaxTimeGap()
...
+	 * a test flips or seeds has to be put back or it leaks into whatever runs next in the same JVM.
+	 */
+	@AfterEach
+	void restoreSharedState()
+	{
+		DecodesSettings.instance().retryFailedComputations = false;
+		CwmsTimeSeriesDAO.cache.clear();
+	}
+
+	@Test
```
MikeNeilson: FYI, all of these are going to be difficult to maintain.

They are also tested as part of the integration tests where resulting behavior is tested, thus making it easier to refactor the behavior without excessive work on maintaining test setup.

In addition to that, do back through CwmsTimeSeriesDao, and the others, and put the private back in.

We generally don't want to break encapsulation just for a test.

MikeNeilson (changes requested):

java/opendcs/src/test/java/decodes/cwms/CwmsTimeSeriesDAOTest.java
tlelv: Okay I reverted the tests. One judgment call to bring to light is SseChannel. Since it is a real type and not a test seam, it fell through. The Sse/SseEventSink/name/taskId fell through four
methods, and its test only uses the public API. Nothing outside the package uses
it, so what do you think if you'd rather it were package-private or gone.

MikeNeilson (approved):

java/opendcs/src/test/java/decodes/cwms/CwmsTimeSeriesDAOTest.java
MikeNeilson: SseChannel should stay, that's not a test "seam" or anything, saves more parameters per call so useful as is and can be useful for future work, whatever that may be.

PR #2149, opened 2026-08-24
Task/1217 routing spec rest api

### LINKED ISSUE #1217: RoutingSpec DAO

None

**Additional context**

MikeNeilson:
## Problem Description

Fixes #1217. 

## Solution

- Update API resources to use the new DAO
- Update Routing Dao for discrepancy in expected behavior.
- Correct test data.

## how you tested the change

Existing tests, no external functional change.

## Where the following done:

### DIFF
--- a/java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/RoutingResources.java
+++ b/java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/RoutingResources.java
@@ -90,7 +90,7 @@ public final class RoutingResources extends OpenDcsResource
 {
     private static final Logger log = OpenDcsLoggerFactory.getLogger();
 
-    private static final WebAppException NO_ROUTING_SPEC_DAO = 
+    private static final WebAppException NO_ROUTING_SPEC_DAO =
         new WebAppException(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(),
                             "No Routing Spec DAO is available.");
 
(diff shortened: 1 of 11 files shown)


java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/RoutingResources.java
```diff
@@ -85,1157 +88,1151 @@
...
+    static RoutingSpec map(ApiRouting routing)
+    {
+        RoutingSpec spec = new RoutingSpec();
+        if(routing.getRoutingId() != null)
+        {
+            spec.forceSetId(DbKey.createDbKey(routing.getRoutingId()));
+        }
+        else
+        {
+            spec.forceSetId(DbKey.NullKey);
```
ktarbet, line 406: what is the significance of forceSetId  versus  setId?

ktarbet (approved):

java/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/RoutingResources.java
MikeNeilson: setId throws a DatabaseException if the id was already set, to protect against some issue with the design that apparently creeped up.
forceSetId doesn't care and just does it; which we know we can do here as we created the object and it's clean.

Long term it should go away in favor of builder patterns or at least have something with better immutability control.

PR #2084, opened 2026-07-15
Add standalone DCPMon web UI

krowvin:
## WIP / visibility

This is the original DCPMon UI draft, submitted for early visibility and not ready to merge.

## Summary

- Adds a standalone DCPMon React UI under `javascript/dcpmon-web-ui`.
- Brings over the DCPMon prototype intent, legacy reference material, mock API data, and MSW setup.
- Adapts the UI to the OpenDCS React stack with React 19, Bootstrap, react-bootstrap, Vite, TypeScript, and Vitest.
- Keeps the DCPMon surface disconnected from the main OpenDCS UI routes, per #2078.
- Preserves mock mode for isolated UI development and testing.
- The current child integration in #2157 uses the generated DDS TypeScript client from #2156 and runs against the real DDS-over-HTTP API in Docker Compose.

## Current UI preview

These screenshots show the current child integration from #2157 running against the local real-DDS Compose stack with a temporary 551-location SWT dataset. The exhaustive seed data used for this review is not committed.

### Light mode and status overview

DCPMon light mode overview

### Display settings

Settings include system/light/dark appearance, a searchable timezone, 12/24-hour time, and optional seconds.

DCPMon display settings

### Partial data and expected transmission slots

The expanded station shows received messages alongside the expected missing transmission.

DCPMon partial station expanded

### Missing data expansion

Expected transmission times are supplied by the API and highlighted when no matching message was received.

DCPMon missing station expanded

### Dark mode with expanded data

DCPMon dark mode with expanded data

### Search

Search matches configured identifiers, descriptions, and NESDIS addresses.

DCPMon search results

## Legacy comparison

The current modernization is being compared against the [legacy SWT DCPStatus](https://www.swt-wc.usace.army.mil/dcpstatus/), including its status categories, search workflow, expected-transmission display, low-battery and GPS-sync reporting, and expandable message details.

## PR stack

1. opendcs/dcs_standards#9 publishes the matching DDS-over-HTTP contract.
2. #2156 generates the private DDS TypeScript client and must update to the immutable standards commit.
3. #2084 is this standalone DCPMon UI branch.
4. Child PR #2157 applies the generated-client and real-DDS integration back into this original `dcpmon/init` branch.

Relates to #2078.

## Notes

This PR preserves the original UI work and discussion. The screenshots above reflect the current child integration in #2157; those API and Compose changes remain isolated there for review before being merged back into this branch.

## Validation

Current integration validation includes:

- `npm test -- --run` — 17 tests passing
[...]

Issue #2078, opened 2026-07-09
Add DCPMon Web UI to OpenDCS


krowvin:
**Is your feature request related to a problem? Please describe.**
SWT and NAE have DCP Status pages we are looking to migrate. I am looking to see if DCPMon can support the need our field staff has. 

DDS  over HTTP was proposed in the standards repo and in there I submitted an openapi spec that would support this need:
https://github.com/opendcs/dcs_standards/pull/3/changes

I created a branch in dcpmon that had these changes in it and a new react based webui that used mocking techniques to mimic the UI locally using test data. 
https://github.com/opendcs/dcpmon/tree/web-app

This was before the OpenDCS web ui modernization.

**Describe the solution you'd like**
Proposal in the OpenDCS call on July 09, 2026 to add another directory in the javascript directory to add visibility to this but not connect it to the main UI yet. 

A request was made to make sure to try and match the OpenDCS UI frameworks so that the components will be cross compatible. The idea being these components could be merged into the DCS UI at a later date. 

**Describe alternatives you've considered**

- Use HADDS / another public DCP message viewing site. While useful, it is nice to have multiple options. We also have full control over *how* we display what is effectively the same data. 

**Additional context**

SWT Legacy Site: https://www.swt-wc.usace.army.mil/dcpstatus
NAE current site: https://water.usace.army.mil/office/nae/tools/dcp-summary

PR #2157, opened 2026-08-29
WIP: Use generated DDS client in DCPMon

krowvin:
## WIP / visibility

This is a draft child PR into the original DCPMon UI branch from #2084. Merging this PR updates `dcpmon/init`; it does not merge directly into `main`.

## Summary

- replaces the original UI's hand-written DDS routes and response types with the generated `opendcs-dds-api` TypeScript client
- extends the DDS OpenAPI contract and regenerates the client during the Gradle build; generated output remains private and is not published to npm
- implements real DDS summary, group discovery, message query envelopes, parsed GOES fields, and expected transmission slots
- runs the production-built DCPMon UI against LRGS DDS-over-HTTP in Docker Compose while retaining MSW for isolated development and tests
- expands the dashboard presentation with schedule-aware counts, missing transmission rows, search, display settings, light/dark/system themes, mobile layout, loading skeletons, and lower-level error boundaries

## Generated client usage

DCPMon configures the generated `DefaultApi` with the DDS base path and calls:

- `groupsGet()`
- `dataSummaryGet({ dataGroup })`
- `dataQueryGet({ source, dcpAddress, since })`

The UI no longer constructs the `/groups`, `/data/summary`, or `/data/query` routes itself and consumes generated models directly.

## DDS contract and server alignment

- resolves status groups from LRGS network-list configuration instead of hardcoding SWT
- calculates expected self-timed transmissions from PDT schedules and reports `unknown` when no usable schedule exists
- returns `expectedMessageTotal`, `transmissionSlots`, low-battery state, GPS synchronization state, and summary counts
- uses the NESDIS/DCP address as each summary map key and returns configured identifiers without inventing SHEF identifiers
- returns documented `DcpMessages` envelopes from query/next and data-source objects from `/sources`
- parses GOES transmit/receive time and signal/message fields
- uses a required top-level message discriminator that the generated TypeScript client can narrow
- treats `source=GOES` as a message-type filter rather than requiring an LRGS input literally named GOES

## Compose integration

- enables DDS-over-HTTP in the LRGS container through environment-driven, idempotent configuration
- serves the production Vite bundle and proxies same-origin `/dds` requests to LRGS
- keeps the OpenDCS API on host port 7000 and exposes direct LRGS DDS HTTP on host port 7001
- includes deterministic development-only fixture coverage for complete, partial, parity, missing, and low-battery conditions
- keeps exhaustive SWT review data outside the repository

## PR stack

1. opendcs/dcs_standards#9: publish the matching DDS-over-HTTP contract
2. #2156: generated DDS TypeScript client foundation, targeting `main`; update it to the immutable standards commit
3. #2084: original DCPMon UI branch, targeting `feature/dds-typescript-client`
4. #2157: this generated-client and real-DDS integration, targeting `dcpmon/init`

Relates to #2078 and builds on DDS-over-HTTP work from #789.

## Validation

- forced no-cache build of all four Compose images completed successfully
- generated DDS client validation and TypeScript build completed successfully
- DCPMon lint passed
- DCPMon production build passed
- DCPMon tests passed: 17 tests across 4 files
- DDS HTTP integration coverage passed locally
- live Compose LRGS, DCPMon, OpenDCS API, authentication, and database services reported healthy
- direct DCPMon health and DDS summary probes returned HTTP 200

The PR remains draft while the stacked API/client work is reviewed and merged in order.


java/lrgs-web/src/main/java/org/opendcs/lrgs/http/dds/DdsSummaryService.java
```diff
@@ -0,0 +1,200 @@
...
+import lrgs.common.NetworkList;
+import lrgs.common.NetworkListItem;
+import lrgs.common.SearchCriteria;
+import lrgs.common.SearchSyntaxException;
+import lrgs.lrgsmain.LrgsConfig;
+import lrgs.lrgsmain.LrgsMain;
+
+final class DdsSummaryService
+{
+    static final int DURATION_HOURS = 24;
```
krowvin, line 38: Should this be configurable or always 24?

docker-compose.yaml
```diff
@@ -1,4 +1,46 @@
 services:
+  lrgs:
```
krowvin, line 2: Should the LRGS be in a different docker compose or started with everything else?

docker-compose.yaml
```diff
@@ -1,4 +1,46 @@
...
+    tmpfs:
+      - /lrgs_home:uid=1000,gid=1000,mode=0755
+    healthcheck:
+      test: ["CMD-SHELL", "curl --fail --silent http://localhost:7000/health >/dev/null"]
+      interval: 5s
+      timeout: 3s
+      retries: 30
+      start_period: 20s
+
+  dcpmon:
```
krowvin, line 26: Same goes for the DCPMonitor

Dockerfile
```diff
@@ -45,14 +49,22 @@ LABEL org.opencontainers.image.documentation=https://github.com/opendcs/opendcs/
...
 USER opendcs:opendcs
 WORKDIR /lrgs_home
 ENV LRGSHOME=/lrgs_home
 # DDS Port
 EXPOSE 16003
+# DDS-over-HTTP Port
+EXPOSE 7000
 CMD ["/lrgs.sh"]
 
+FROM nginx:alpine AS dcpmon
```
krowvin, line 63: NGINX hosts the built vite source from dist, and will do the same proxying we have setup for npm run dev for /dds/*. This lets us test the actual production bundle. 

Currently there's mock source if you use npm run dev, having this setup will enable us to fully test dcpmon against the current running OpenDCS setup on localhost. 

Vs if you just want to develop on localhost you could then do the normal `npm run dev` with the mock data. 

Probably could just use the tomcat or embedded Jetty if you prefer? Just under /dcpmon next to /dds? With tomcat having another war. 

Keeping it separate from that tomcat instance could mean that it's independent of that container and we only worry about the web UI and the LRGS.

java/lrgs-web/src/main/java/org/opendcs/lrgs/http/dds/DdsSummaryService.java
MikeNeilson: configurable, but 24 hours is a good default.

If creating a constant use `java.time.TimeUnit` though.... or a `java.time.Duration`,... actually `java.time.Duration` and the input from the user should be the ISO8601 duration or interval (we can "steal" the parsing logic from CDA.)

docker-compose.yaml
MikeNeilson: Just with everything else. eventually it's config will be the the "OpenDCS" database.

docker-compose.yaml
MikeNeilson: It should not be used for this effort. We're adding end points through the DDS-over-HTTP that cover extraction from the archive. So now it's just direct to the XML archive.

the dcp_trans_xxx tables should no longer be used for anything new.

Dockerfile
MikeNeilson: Yeah, don't add thing. Also, FYI, tweak the Lrgs config to start the new DDS-over-HTTP module and use the traefik definition for routing it.

MikeNeilson: Wait, sorry, the LRGS just hosts the API not the UI.
 
For now I say package it up into the OpenDCS UI tomcat instance as another web app there.

Or create it purely as a react component and pull it in as a dependencies to the OpenDCS UI... which ever seems easier for the moment.

MikeNeilson: it *would* be good to let it run standalone rendered from the LRGS's new HttpInput, but let's not focus on that just yet. I think LRGS, at least *our* implementation, is staring to get away from being a standalone thing in that nature.

PR #2162, opened 2026-09-01
Cwms Vertical Datum Conversion Calculation targeting 7.0

### LINKED ISSUE #752: Add Comp that can access CWMS_V_VERT_DATUM_OFFSET and CWMS_V_LOC table for Datum Conversions
Currently Stage to Elevation comps are managed in our database as tab file rating shifts. There are also offsets assigned to the location in the CWMS_V_VERT_DATUM_OFFSET table and CWMS_V_LOC . The offsets in CWMS_V_VERT_DATUM_OFFSET  are returned in CDA calls.

Have one spot in the CWMS database to manage datum conversions. Seems like CWMS_V_VERT_DATUM_OFFSET  can do that, but we need access to edit it and comps to apply the offset to raw data. This comp wouldn't help with access though.

Continue with CWMS rating shifts, or USGS equation, or scalar adder to manage datum conversions.

msweier:
## Problem Description

This AI assisted PR used ChatGPT 5.4 and fixes #752.

CWMS users need a simple way to handle datum conversions in one place instead of managing stage-to-elevation adjustments through rating shifts or other workaround computations. The database supports these conversions so it would be great to use them in DCS.

CWMS stores the vertical datum offsets in the `CWMS_V_VERT_DATUM_OFFSET` which is not accessible from a GUI at the moment so it likely not widley used. In theory STAGE to ELEV offsets could live here too.

The `CWMS_V_LOC` is exposed via a GUI in CWMS-VUE (the Locations tab) and the elevation and datum field there may be a convenient place to store the STAGE to ELEV in the native datum offset. 

## Solution

This PR introduces a new CWMS-only algorithm: `CwmsVerticalDatumConversion`.

 The feature supports two modes:

  - `locationElevationOffset` (the CWMS-VUE offset)
    Starts with a `STAGE` value, uses `CWMS_V_LOC.elevation` and `CWMS_V_LOC.vertical_datum` to convert stage to the location’s native elevation, and then, if desired, uses `CWMS_V_VERT_DATUM_OFFSET` through `cwms_loc.get_vertical_datum_offset` to convert to the requested output datum.

  - `cwmsDatumOffset` (the "hidden" table offsets)
    Converts directly from `datum1` to `datum2` using `CWMS_V_VERT_DATUM_OFFSET` through `cwms_loc.get_vertical_datum_offset`. This could support a STAGE to ELEV conversion if the offsets are stored in that table.

 This PR also includes the standard algorithm metadata, tests, and user-facing documentation needed to introduce the feature.

## how you tested the change
 - Added unit tests for `CwmsVerticalDatumConversion`
  - Verified direct datum-to-datum conversion behavior testing locally on windows
  - Verified stage-to-native-elevation behavior using `CWMS_V_LOC` testing locally on windows
  - Verified native-datum to target-datum conversion using `CWMS_V_VERT_DATUM_OFFSET` vertical datum offsets testing locally on windows
  - Verified blank `datum2` resolution from `CWMS_V_LOC.vertical_datum` testing locally on windows
  - Built the docs

## Where the following done:


install/src/main/dist/imports/comp-standard/CwmsVerticalDatumConversion.xml
MikeNeilson, line 1: You can delete this now, it's created from the annotations you've applied to the algorithm.

java/opendcs/src/main/java/decodes/cwms/algo/CwmsVerticalDatumConversion.java
```diff
@@ -0,0 +1,548 @@
...
+		name = "conversionMode",
+		propertySpecType = PropertySpec.STRING,
+		description = "Conversion source: 'cwmsDatumOffset' (default) uses "
+			+ "cwms_loc.get_vertical_datum_offset; 'locationElevationOffset' "
+			+ "uses CWMS_V_LOC.elevation and CWMS_V_LOC.vertical_datum from CWMS-VUE Locations tab for "
+			+ "STAGE to native-datum elevation before any native-to-datum2 "
+			+ "CWMS conversion.")
+	public String conversionMode = MODE_CWMS_DATUM_OFFSET;
+
+	String _propertyNames[] =
```
MikeNeilson: you can remove this block.

java/opendcs/src/main/java/decodes/cwms/algo/CwmsVerticalDatumConversion.java
```diff
@@ -0,0 +1,548 @@
...
+			String verticalDatum)
+		{
+			this.elevationOffsetInInputUnit = elevationOffsetInInputUnit;
+			this.elevationInMeters = elevationInMeters;
+			this.verticalDatum = verticalDatum;
+		}
+	}
+
+	@FunctionalInterface
+	interface LocationSiteLoader
```
MikeNeilson, line 162: interesting, needs some commentary on why it exists.

I'd probably also just call it SiteLoader and have it take SiteName objects over a string... and be in it's own file. (I can see some future use, only the comment is really needed right now.)

java/opendcs/src/main/java/decodes/cwms/CwmsSiteDAO.java
```diff
@@ -517,4 +517,33 @@ public void readSite(Site site) throws DbIoException, NoSuchObjectException
 			throw new DbIoException("Unable to retrieve site.", ex);
 		}
 	}
+
+	public Site readSiteFromCwmsLoc(DbKey id) throws DbIoException, NoSuchObjectException
+	{
+		Site site = new Site();
```
MikeNeilson, line 508: the SiteDao already retrieves the Site information from the cwms location table. Just use the existing readById or whatever it's called.

java/opendcs/src/main/java/decodes/cwms/algo/CwmsVerticalDatumConversion.java
```diff
@@ -0,0 +1,548 @@
...
+			elevationOffsetInInputUnit,
+			elevationInMeters,
+			normalizedVerticalDatum);
+	}
+
+	private Site readLocationSiteFromCwmsLoc(String locationId, String office)
+		throws DbCompException
+	{
+		CwmsSiteDAO siteDao = new CwmsSiteDAO((CwmsTimeSeriesDb) tsdb, office);
+		try
```
MikeNeilson:
```suggestion
		try (CwmsSiteDAO siteDao = new CwmsSiteDAO((CwmsTimeSeriesDb) tsdb, office))
```

java/opendcs/src/main/java/decodes/cwms/algo/CwmsVerticalDatumConversion.java
```diff
@@ -0,0 +1,548 @@
...
+		{
+			throw new DbCompException(
+				"Failed to load CWMS_V_LOC.elevation and CWMS_V_LOC.vertical_datum "
+			  + "for location=" + locationId + ".",
+				ex);
+		}
+		finally
+		{
+			siteDao.close();
+		}
```
MikeNeilson:
```suggestion
```

java/opendcs/src/main/java/decodes/cwms/algo/CwmsVerticalDatumConversion.java
```diff
@@ -0,0 +1,548 @@
...
+		Database db = Database.getDb();
+		if (db == null || db.unitConverterSet == null)
+		{
+			throw new DbCompException(
+				"CwmsVerticalDatumConversion conversionMode=locationElevationOffset "
+			  + "cannot convert CWMS_V_LOC.elevation from '" + fromUnit + "' to '"
+			  + toUnit + "' because engineering unit converters are not initialized.");
+		}
+
+		EngineeringUnit fromEu = EngineeringUnit.getEngineeringUnit(fromUnit);
```
MikeNeilson: Get the converter once and store it.

java/opendcs/src/test/java/decodes/cwms/algo/CwmsVerticalDatumConversionTest.java
```diff
@@ -0,0 +1,365 @@
...
+		assertEquals("LOCAL", stubDao.lastDatum1);
+		assertEquals("NAVD88", stubDao.lastDatum2);
+		assertEquals(baseTime, stubDao.lastDatetime);
+		assertEquals(191.88, algo.valueInDatum2.getDoubleValue(), 1e-9);
+	}
+
+	private static void setPrivateField(Object target, String fieldName, Object value)
+		throws NoSuchFieldException, IllegalAccessException
+	{
+		Class<?> cls = target.getClass();
```
MikeNeilson: This a strong indication that something is fundementally wrong with the design (there is, it's not anyone's fault.

at the same time, there are existing utilities that perform this function, see DbAlgorithmExecutive and AW_AlgorithmBase to verify if they can be used directly, or at least extracted out to a shared utility.

MikeNeilson (changes requested):
The new Test is good for the concepts.

This does need an integration test.

@oskarhurst can you work with Mitch on some of the cleanup and getting a proper integration test working?

install/src/main/dist/imports/comp-standard/CwmsVerticalDatumConversion.xml
msweier: Yep forgot about that. Thanks

java/opendcs/src/main/java/decodes/cwms/CwmsSiteDAO.java
msweier: Right, this was cleaned up a little. I originally the SiteDao but the opendcs site properties were clobbering those in the cwms table. In my case, I had a native datum of 88 in opendcs and LOCAL (MSL1912) in cwms. Assuming cwms is the source of truth, I left in the site read. Unless there is a good way to make opendcs congruent with cwms? Seems like we should just have one spot for that metadata stuff eventually....I assume that is on the list somewhere.

java/opendcs/src/main/java/decodes/cwms/algo/CwmsVerticalDatumConversion.java
msweier: Good idea!

java/opendcs/src/main/java/decodes/cwms/CwmsSiteDAO.java
MikeNeilson: Hmm. Let me dig around a little. over in the main branch were completely rebuilding the DAO system. This doesn't seem unreasonable but it does seem like the datum should just be removed from the OpenDCS site properties and the CWMS value used (that is what the modernized DAO does.)

java/opendcs/src/main/java/decodes/cwms/algo/CwmsVerticalDatumConversion.java
MikeNeilson: okay, that makes some sense. Not ideal, but given the age of the design it's an acceptable workaround.

java/opendcs/src/main/java/decodes/cwms/algo/CwmsVerticalDatumConversion.java
```diff
@@ -396,8 +404,7 @@ private LocationElevationInfo loadLocationElevationInfo() throws DbCompException
 	private Site readLocationSiteFromCwmsLoc(String locationId, String office)
 		throws DbCompException
 	{
-		CwmsSiteDAO siteDao = new CwmsSiteDAO((CwmsTimeSeriesDb) tsdb, office);
-		try
+		try (CwmsSiteDAO siteDao = new CwmsSiteDAO((CwmsTimeSeriesDb) tsdb, office))
```
MikeNeilson, line 407: This could also be `...siteDao = (CwmsSiteDao)tsdb.makeSiteDao()`

but let's just leave it for now, it's clearly a CWMS Only algorithm and certainly will be in the 7.0 branch.

That said, when we port this to main (that's what CWMS is using in the cloud environments) we'll want to use more generic interfaces.

MikeNeilson (changes requested):
That looks better, still needs the integration tests.

If @oskarhurst can't help this week, @ktarbet or I should be able to next week.

msweier:
> That looks better, still needs the integration tests.

Sorry i just pushed during your review.

java/opendcs/src/main/java/decodes/cwms/algo/CwmsVerticalDatumConversion.java
```diff
@@ -0,0 +1,592 @@
...
+		}
+
+		Database db = Database.getDb();
+		if (db == null || db.unitConverterSet == null)
+		{
+			throw new DbCompException(
+				"CwmsVerticalDatumConversion conversionMode=locationElevationOffset "
+			  + "cannot convert CWMS_V_LOC.elevation from '" + fromUnit + "' to '"
+			  + toUnit + "' because engineering unit converters are not initialized.");
+		}
```
MikeNeilson, line 536:
```suggestion
```

You've already attempted to build the unit converter above, additional checks are not required.

java/opendcs/src/main/java/decodes/cwms/algo/CwmsVerticalDatumConversion.java
```diff
@@ -0,0 +1,592 @@
...
+
+		throw new DbCompException(
+			"Unknown effectiveDateMode '" + effectiveDateMode
+		  + "'. Expected 'latestOnOrBefore' or 'latestOverall'.");
+	}
+
+	private double convertUnits(double value, String fromUnit, String toUnit)
+		throws DbCompException
+	{
+		if (fromUnit == null || toUnit == null || fromUnit.equalsIgnoreCase(toUnit))
```
MikeNeilson, line 524:
```suggestion
		if (fromUnit.equalsIgnoreCase(toUnit))
```

You've already verify neither of these values are null.

java/opendcs/src/main/java/decodes/cwms/algo/CwmsVerticalDatumConversion.java
MikeNeilson: actually just called the unit converter. see comment below in the initialize method.

java/opendcs/src/main/java/decodes/cwms/algo/CwmsVerticalDatumConversion.java
```diff
@@ -0,0 +1,592 @@
...
+				ex);
+		}
+	}
+
+	private void initializeLocationElevationUnitConverter(String fromUnit, String toUnit)
+		throws DbCompException
+	{
+		if (fromUnit == null || toUnit == null || fromUnit.equalsIgnoreCase(toUnit))
+		{
+			locationElevationUnitConverter = null;
```
MikeNeilson, line 565:
```suggestion
			locationElevationUnitConverter = new NullConverter();
```

`NullConverter` is a passthrough. It saves the trouble of having to do the if check as that choice has already been determined here.

MikeNeilson (commented):
That looks reasonable, I still want to dig into the SiteDao a bit, will likely not be able to do so today though.

java/opendcs/src/main/java/decodes/cwms/algo/CwmsVerticalDatumConversion.java
msweier: I tried this, but in 7.0 the NullConverter requires two EngineeringUnit args so it failed.

msweier: I updated this to use NullConverter with the EngineeringUnit args

PR #2164, opened 2026-09-03
Configure Vite allowed hosts through the environment

krowvin:
## Summary
Configure Vite development-server allowed hosts through the `DEV_ALLOWED_HOSTS` process environment variable instead of a hardcoded organizational domain. Docker images or workspace environments can supply a comma-separated list of hostnames; whitespace and empty entries are ignored, and an unset or empty value preserves Vite defaults.

Document the Dockerfile setting, domain suffix matching, and restart behavior in the UI README. The variable is not exposed to browser code and does not configure the production web server.

## Validation
- Passed six environment-value cases using the allowedHosts expression extracted from the changed config with Vite 7.3.6 in an isolated server fixture: unset, empty, whitespace-only, exact host, multiple hosts, and domain suffix.
- Live HTTP checks confirmed configured hosts and localhost are accepted and unrelated hosts are rejected.
- Prettier checks passed for both changed files; `git diff --check` passed.
- Full application build and Docker-image integration were not run.

### DIFF
--- a/javascript/opendcs-web-ui-react/vite.config.ts
+++ b/javascript/opendcs-web-ui-react/vite.config.ts
@@ -31,6 +31,7 @@ export default defineConfig({
     ],
   },
   server: {
+    allowedHosts: [".apps.hecdev.net"],
     fs: {
       // Allow serving files from one level up the project root
       allow: [
--- a/javascript/opendcs-web-ui-react/vite.config.ts
+++ b/javascript/opendcs-web-ui-react/vite.config.ts
@@ -31,7 +31,10 @@ export default defineConfig({
     ],
   },
   server: {
-    allowedHosts: [".apps.hecdev.net"],
+    allowedHosts: (process.env.DEV_ALLOWED_HOSTS ?? "")
+      .split(",")
+      .map((host) => host.trim())
+      .filter(Boolean),
     fs: {
       // Allow serving files from one level up the project root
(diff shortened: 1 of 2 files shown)


javascript/opendcs-web-ui-react/vite.config.ts
ktarbet, line 1: Could the allowedHosts be moved to .env.local (or something like that). to keep this more private?

krowvin: Good point, and it doesn't make it configurable for others that might want to add their own.

ktarbet (approved):

ktarbet:
I see this error:

```
Node 20 is being deprecated. This workflow is running with Node 24 by default. If you need to temporarily use Node 20, you can set the ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true environment variable.

MikeNeilson:
> I see this error:
> 
> ```
> Node 20 is being deprecated. This workflow is running with Node 24 by default. If you need to temporarily use Node 20, you can set the ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true environment variable.
> ```

That's from various actions that we use. I know I had at least one updated upstream (though for the life of me I can't remember which one) but it's been taking everyone a while to make that change.

PR #2166, opened 2026-09-04
fix/Manual ts computation

tlelv:
## Problem Description

Fixes #. 

Running a computation manually from the web UI returned no data and no useful trace,
while the same computation run automatically by the computation processor worked fine.

## Solution

Describe your solution.

So I found out manual runs never saved their output. It computed everything and discarded it with some more inefficiencies. It now collect outputs and saves them once the batch has joined.

## how you tested the change

## Where the following done:


integrationtesting/opendcs-tests/src/test/java/org/opendcs/odcsapi/res/it/ComputationResourcesIT.java
```diff
@@ -856,6 +863,24 @@ void testExecuteComputationCWMS() throws Exception
 						foundCount++;
 						JsonPath jsonPath = JsonPath.from(sseEvent.readData(String.class, MediaType.TEXT_PLAIN_TYPE));
 						assertEquals(expectedTsList.size(), jsonPath.getList("tsIds").size(), "Expected " + expectedTsList.size() + " results");
+						// A run that produced output must report it with the real database key --
```
MikeNeilson: Slight snag here... on a manual run we *don't* want to save anything until the user has reviewed the data.

... how we deal with that, I have no idea (other than perhaps storing it in the user session and adding an extra hint variable (like the run id) to the timeseries retrieval endpoint to divert there. We can discuss next week.

tlelv: I get what you mean. I think it was better to rework it so the run persists nothing and the computed values now come back inline, but it still fixes the original "manual run shows nothing" bug.

tlelv:
For this PR, rather than it "Persists the computed output", it returns computed values inline, and persist nothing. A manual run must not write until the operator has reviewed the numbers, so `executeAndPublishResult` collects the computed series and the `Results` carries them in a new `ApiCompResults.data` field. Manual run does not write to the database.

MikeNeilson (commented):
This seems reasonable. I certainly don't see any issues with smaller data sets. I suspect we'll have issues with larger data sets, but I think this puts us on good footing to deal with that later.

MikeNeilson (commented):
Oh before I forget, the test should include some data with quality flags. Keep the existing test without, but for, say screening algorithms, seeing those is rather critical.

MikeNeilson (approved):

PR #2169, opened 2026-09-08
Refactor/55 decode lang subproject

### LINKED ISSUE #55: DECODES language components as a subproject/module

Instead of people just rewriting everything allow the DECODES ecosystem to be used in downsteam project by creating a seperate jar.

I think we'd still bundle the class filesninto opendcs.jar but this would involve trying to isolate what could be generic vs what interface are clearly specific to opendcs as a whole so there could be an opendcs-core jar as well.

I knoe the CWBI folks dont use the bulk of OpenDCS for GOES data just enough to have more processable files. This would let them reduce container siE by not bring in components that arent ever used.

**Additional context**
My hope is that this is the first in a series of reorganizing for future development and operational needs.

MikeNeilson:
## Problem Description

Fixes #55. 

## Solution

I considered just trying to migrate the existing DECODES processing to a subproject, however it is too tied to the existing structures. Instead I've created an Antlr4 grammar (see https://github.com/opendcs/opendcs-shef for some existing work) that will be able to output appropriate parsers for any language (like JavaScript/TypeScript for the web UI).

At this time, I'm just outputting and verifying the parser. I believe the end goal should be to output a "Compiled Script" that can then be easily passed an input message and a callback for data output. However, I'm not sure if that can be easily done for two languages and it may be reasonably fast and easy to just treat the parsed script as "compiled", will depend on if the parser can be reused, I think so.

## how you tested the change

TODO, will create additional tests of the parser itself and some sort of sample usage in Java and JavaScript/TypeScript.

## Where the following done:


MikeNeilson:
The current parse tree based on the grammar.

I'm treating what's created for the tests (the line based .script files) as the "format". Basic idea is that anything in the database/xml right now would get combined into one string and that parsed out.

NOTE: the parsing itself only verifies basic structure. While there is a fair amount of validation that can be done in the grammar, further code is required for full validation and operations. Such as if linking same labeled format statements. Or in other words, the grammar parses, then we'll still need to "compile" to something that can be executed.

decodes_antlr4_parse_tree


To perform this operation locally:
`./gradlew :decodes-lang:grun --info` will pop up the GUI showing the parse tree, the terminal will contain information about how it was parsed.

MikeNeilson:
I think I'm going to split this into it's own repo. So far I've been able to stage everything to avoid depending on any internals while allowing for the current and future implemenation details.

Certainly still more work to do, but will be faster builds and tests without having to run everything else which isn't affected by this.

MikeNeilson:
Moved work to https://github.com/opendcs/decodes-lang

PR #2177, opened 2026-09-10
Fix/inconsistent page loads

### LINKED ISSUE #2052: Web UI - Visual differences when loading pages within the same session
> Looks great, thank you!
> 
> I'll note here that the first time I tried to test this, I didn't get the dropdown for my Units and only had one Data Type Standard available in the dropdown.  But after hitting Cancel and moving around a bit, then coming back, the dropdowns were both there and correct.  Not sure why that occurred or if it was just a fluke. 

 _Originally posted by @jbatterman in [#2030](https://github.com/opendcs/opendcs/issues/2030#issuecomment-4803002873)_

Throughout my testing of the Web UI, I have noticed that if I leave a page and come back to it (in the same session), sometimes I see different results when the page loads.  Two examples below.

1. On the Platforms page, sometimes when it loads it shows the usgs ID as the Site Name, and other times it loads showing the nwshb5 ID and cwms ID.
2. In multiple areas when testing edits, the dropdown for Units may appear or may not appear as a dropdown box.  I also have seen dropdowns that don't have all the options available, when on returning to that page and editing again, the dropdown has all the options available.

It's tough to replicate these things and I don't necessarily take screenshots when something looks different, but if I catch one again I'll try to screenshot and add a comment here.

tlelv:
## Problem Description

Fixes #2052. 

## Solution

Describe your solution.

* `PlatformDaoImpl.sql.stg` will order the site-name rows by preferred-type-first. This makes both the `sitenames` map order and the server-computed `name` field stable.
* `REFERENCE_DATA_CACHE` sets `gcTime >= staleTime` and is applied to the reference-data queries.
* `AppDataTable` has a column render fix so "edit" cells pick up the reference data that will arrive after the row is opened.

## how you tested the change

* Added integration test 
* On the platforms page, reloaded repeatedly on a site with a lot of name types and confirmed the Site column no longer changes name type between loads. 
* Presentation Groups -> edit a group -> edit an element: Units renders as a
   `<select>` with the full unit list 

## Where the following done:

### DIFF
--- a/java/cwms-implementation/src/main/resources/org/opendcs/database/impl/cwms/dao/CwmsPlatformDaoImpl.sql.stg
+++ b/java/cwms-implementation/src/main/resources/org/opendcs/database/impl/cwms/dao/CwmsPlatformDaoImpl.sql.stg
@@ -51,7 +51,11 @@ with platforms(id, agency, isproduction, description, lastmodifytime, configid, 
     <if(platform_sensor_join)><platform_sensor_join><endif>
     <if(platform_sensor_props_join)><platform_sensor_props_join><endif>
 
-    order by p.mediumtype <collate> asc, p.mediumid <collate> asc, p.platformdesignator <collate> asc
+    order by p.mediumtype <collate> asc, p.mediumid <collate> asc, p.platformdesignator <collate> asc<if(site_name_columns)>,
+        case
+            when lower(sn.nametype) = lower(:preferredType) then 0
+            else 1
+        end, sn.nametype <collate> asc, sn.sitename <collate> asc<endif>
 >>
 
 platformMerge(dual) ::= <<
--- a/java/opendcs-implementation/src/main/resources/org/opendcs/database/impl/opendcs/dao/PlatformDaoImpl.sql.stg
+++ b/java/opendcs-implementation/src/main/resources/org/opendcs/database/impl/opendcs/dao/PlatformDaoImpl.sql.stg
@@ -33,7 +33,11 @@ with platforms(id, agency, isproduction, description, lastmodifytime, configid, 
     <if(platform_sensor_join)><platform_sensor_join><endif>
     <if(platform_sensor_props_join)><platform_sensor_props_join><endif>
 
-    order by p.mediumtype <collate> asc, p.mediumid <collate> asc, p.platformdesignator <collate> asc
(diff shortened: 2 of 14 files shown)


java/cwms-implementation/src/main/resources/org/opendcs/database/impl/cwms/dao/CwmsPlatformDaoImpl.sql.stg
```diff
@@ -51,7 +51,11 @@ with platforms(id, agency, isproduction, description, lastmodifytime, configid,
     <if(platform_sensor_join)><platform_sensor_join><endif>
     <if(platform_sensor_props_join)><platform_sensor_props_join><endif>
 
-    order by p.mediumtype <collate> asc, p.mediumid <collate> asc, p.platformdesignator <collate> asc
+    order by p.mediumtype <collate> asc, p.mediumid <collate> asc, p.platformdesignator <collate> asc<if(site_name_columns)>,
```
MikeNeilson, line 54:
```suggestion
    order by p.mediumtype <collate> asc, p.mediumid <collate> asc, p.platformdesignator <collate> asc
    <if(site_name_columns)>,
```

Bit hard to spot if it's on the end there. (the comma also just needs to exists, doesn't have to be on the same line.)

java/opendcs-implementation/src/main/resources/org/opendcs/database/impl/opendcs/dao/PlatformDaoImpl.sql.stg
```diff
@@ -33,7 +33,11 @@ with platforms(id, agency, isproduction, description, lastmodifytime, configid,
     <if(platform_sensor_join)><platform_sensor_join><endif>
     <if(platform_sensor_props_join)><platform_sensor_props_join><endif>
 
-    order by p.mediumtype <collate> asc, p.mediumid <collate> asc, p.platformdesignator <collate> asc
+    order by p.mediumtype <collate> asc, p.mediumid <collate> asc, p.platformdesignator <collate> asc<if(site_name_columns)>,
```
MikeNeilson, line 36:
```suggestion
    order by p.mediumtype <collate> asc, p.mediumid <collate> asc, p.platformdesignator <collate> asc
    <if(site_name_columns)>,
```

MikeNeilson (changes requested):

MikeNeilson (approved):

PR #2183, opened 2026-09-14
Prevent browsers from caching index.html

ktarbet:
## Problem Description

When developing on ubuntu/firefox I often get a blank page at http://localhost:7000/ after a new build

## Solution

Disable caching in tomcat

## how you tested the change

manually


tlelv:
When I was looking through for tomcat, it got to / and /index.html but the deeper links like /platforms get index.html from the 404 <error-page>.

What do you think about something like a dispatcher?

<'filter-mapping'>
&nbsp;&nbsp;<'filter-name>ExpiresFilter</filter-name'>
&nbsp;&nbsp;<'url-pattern>/*</url-pattern'>
&nbsp;&nbsp;<'dispatcher>REQUEST</dispatcher'>
&nbsp;&nbsp;<'dispatcher>ERROR</dispatcher'>
<'/filter-mapping'>

tlelv (changes requested):
When I was looking through for tomcat, it got to / and /index.html but the deeper links like /platforms get index.html from the 404 .

What do you think about something like a dispatcher?

<'filter-mapping'>
  <'filter-name>ExpiresFilter</filter-name'>
  <'url-pattern>/*</url-pattern'>
  <'dispatcher>REQUEST</dispatcher'>
  <'dispatcher>ERROR</dispatcher'>
<'/filter-mapping'>

ktarbet:
If this is breaking other pages, I'll close the PR.

PR #2189, opened 2026-09-14
Slow down macos - try to increase macos build sucess rate.

ktarbet:
MacOs Builds are often failing on github.  Trying without parallel, and ensure storybook timeout is specifically set.


javascript/opendcs-web-ui-react/vite.config.ts
```diff
@@ -105,6 +105,8 @@ export default defineConfig({
           // The cache (see .github/workflows/build.yml) covers the broad
           // case; this retry handles the residual single-file flake.
           retry: 1,
+          // take it easy on macOS. github provides 3 vCPUs (and often fails)
```
tlelv, line 108: This part might make the macOS part longer by turning off its parallelism. Im thinking:
* Vitest's browser mode runs min(12, cpus − 1) story files in parallel, so 2 on the 3-vCPU macOS runner. 
* fileParallelism: false makes that 1, which can make the Storybook step up to around 2× slower on macOS. 

So I would think to try without this first and add later? What do you think

tlelv (approved):

javascript/opendcs-web-ui-react/vite.config.ts
ktarbet: The macos build has been failing regularly; this is an attempt to actually slow macos down by turning off parallel, hopting it will be more stable with tradeoff of slower.

